PR #41, opened 2023-01-25
Automated generation of two test central repositories

marckhouzam:

This PR allows to have two test Central Repositories running locally in docker.
This will allow to test the upcoming changes to support the Central Repository feature.

Because the Tanzu CLl does not yet support the Central Repository format, it cannot be used to test this PR.
But we can test it manually.  You only need docker to be able to use this test repos.

The below starts the test central repo, pulls the database from it and then queries the DB.
```
make start-test-central-repo
imgpkg pull -i localhost:9876/tanzu-cli/plugins/central:large -o tmp
echo 'SELECT * FROM PluginBinaries WHERE pluginname="kubeplugin" AND version="v9.9.9";' | sqlite3 -batch tmp/central.db
```
The below pulls a (fake) plugin binary from the test central repo and runs it.
```
imgpkg pull -i localhost:9876/tanzu-cli/plugins/vmware/tkg/darwin/arm64/tmc/twotargets:v9.9.9 -o tmp2
chmod u+x tmp2/fakeplugin2.sh
./tmp2/fakeplugin2.sh info
```

### Additional information

You need `docker` installed to use the test Central Repo.

The `README.md` mentions how to use the CLI with this test central repo, those commands don't work yet; they will be made available in follow-up PRs.


hack/central-repo/README.md
```diff
@@ -0,0 +1,66 @@
...
+A tarball is used to dramatically reduce the size saved in git.
+
+If it is necessary to re-generate a new test central repo, it took around 1 minute on a Mac M1.
+The procedure follows:
+
+```bash
+cd tanzu-cli
+make stop-test-central-repo
+cd hack/central-repo
+\rm -rf registry-content registry-content.bz2
```
anujc25, line 61: Is this line needed? Looks like it is getting ignored.

hack/central-repo/upload-plugins.sh
```diff
@@ -0,0 +1,140 @@
...
+                    echo $sql_cmd | sqlite3 -batch $database
+                fi
+
+                # For efficiency, only push the plugin binaries of the two latest versions for 4 plugins
+                # The plugins are: twotargets (for both targets), kubeplugin, tmcplugin 
+                if [ $name = "twotargets" ] || [ $name = "kubeplugin" ] || [ $name = "tmcplugin" ]; then
+                    if [ $version = "v8.8.8" ] || [ $version = "v9.9.9" ]; then
+                        ${dry_run} imgpkg push -i $repoBasePath/$image_path:$version -f $tmpPluginPhase2 --registry-insecure
+                    fi
+                fi
```
anujc25, line 106: We should also document this in the README specifying only these plugin binaries will be available in the central repository. Other plugins are added to the database for testing but plugin binaries are not available within this local test central repository.

Also, the PR description mentions running `imgpkg pull -i localhost:9876/tanzu-cli/plugins/vmware/tkg/darwin/arm64/tmc/plugin0:v9.9.9 -o tmp2` command to fetch plugin binary but binary for `plugin0` doesn't exist in the central repository and command fails with an error. PR description needs to be updated to specify the plugin binary which exists in the local test central repository.

hack/central-repo/upload-plugins.sh
```diff
@@ -0,0 +1,140 @@
...
+if [ $fast = "off" ]; then
+    for idx in {0..94}; do
+        target_rand=$(($RANDOM % 3))
+        case $target_rand in
+        0) target=global
+           ;;
+        1) target=k8s
+           ;;
+        2) target=tmc
+           ;;
```
anujc25, line 130: Just to clarify, we are storing all non-targeted plugins with Target value as `global` in the database. Correct? 
Basically `TARGET` field within the database will have either a `global`, `k8s`, or `tmc` value and it should not be empty.

anujc25 (commented):
Overall this looks good. Just have few minor questions.

hack/central-repo/README.md
marckhouzam: I think you are right that removing `registry-content.bz2` is not necessary as the below `tar` command will re-generate it.  I added it because if someone forgets to run the tar command, they will more easily notice because git will mention that the `registry-content.bz2` has been removed.

As for `registry-content` I haven't tested without removing it, but I assumed it was important because if we don't, the modified generation we are running now would start with a registry that was not empty.

hack/central-repo/upload-plugins.sh
marckhouzam: > We should also document this in the README specifying only these plugin binaries will be available in the central repository. Other plugins are added to the database for testing but plugin binaries are not available within this local test central repository.

😉  https://github.com/vmware-tanzu/tanzu-cli/pull/41/files#diff-3f8fe7016a74b43b2511fe3745969ec27a165a0e396225e8da9fd98365f4fd48R41-R44

BTW, this approach is a little awkward because not all plugins are available to install.
But it does make generation much faster and the `registry-content.bz2` git file much smaller.
Do you think it makes using the test central repo too difficult?

> Also, the PR description mentions running `imgpkg pull -i localhost:9876/tanzu-cli/plugins/vmware/tkg/darwin/arm64/tmc/plugin0:v9.9.9 -o tmp2` command to fetch plugin binary but binary for `plugin0` doesn't exist in the central repository and command fails with an error. PR description needs to be updated to specify the plugin binary which exists in the local test central repository.

Woops!  Updated.  Thanks.

hack/central-repo/upload-plugins.sh
marckhouzam: I'm glad you caught that 😄 .  This is something I needed to discuss with you to align with the publishing tooling.
I used the "global" keyword to make it easier to read the database manually, but in the code I convert it back to the empty string to fit with the code base.  I am fine with leaving the field empty here.  What do you prefer?

I wanted to confirm with you that we don't support a plugin in the "global" target AND the same plugin name in the "tmc" target?  The only possible double target plugin name is "k8s" and "tmc", right?

hack/central-repo/README.md
anujc25: Got it. Curious, is `\` at the start of the command specified for any specific reason? What does that mean?

prkalle (approved):
LGTM. These changes are great @marckhouzam !
The changes are good to go once the comment on "global" as `TARGET` value is sorted out.

hack/central-repo/upload-plugins.sh
anujc25: > I wanted to confirm with you that we don't support a plugin in the "global" target AND the same plugin name in the "tmc" target? 

We don't support a plugin in the "global" target AND the same plugin name in the **"k8s"** target.

The below combinations are supported:

```
name: plugin0 , target: global
name: plugin0 , target: tmc
```
```
name: plugin1 , target: k8s
name: plugin1 , target: tmc
```

The below combination is NOT supported:
```
name: plugin2 , target: k8s
name: plugin2 , target: global
```

anujc25: > I used the "global" keyword to make it easier to read the database manually, but in the code I convert it back to the empty string to fit with the code base. I am fine with leaving the field empty here. What do you prefer?

I am fine with using "global" keyword in the database.

hack/central-repo/README.md
anujc25: Thanks for the context. I learned this today :)

hack/central-repo/upload-plugins.sh
marckhouzam: > The below combinations are supported:
>
> name: plugin0 , target: global
> name: plugin0 , target: tmc

@anujc25 That's the one I've been thinking about.  How would a user install plugin0 for the global target?
`tanzu plugin install plugin0` will find two plugins but we don't have a target name for the global target to use with the `--target` flag

marckhouzam:
> The below combinations are supported:
>
> name: plugin0 , target: global
> name: plugin0 , target: tmc

@anujc25 That's the one I've been thinking about.  How would a user install plugin0 for the global target?
`tanzu plugin install plugin0` will find two plugins but we don't have a target name for the global target to use with the `--target` flag.  So I think only the tmc one can actually be installed.

I'm bringing this up because I was thinking we may need to introduce an explicit `global` target.   The alternative is to not allow the combo global and tmc.

What do you think?

anujc25:
> How would a user install plugin0 for the global target?
tanzu plugin install plugin0 will find two plugins but we don't have a target name for the global target to use with the --target flag. So I think only the tmc one can actually be installed.
>
> I'm bringing this up because I was thinking we may need to introduce an explicit global target. The alternative is to not allow the combo global and tmc.

Good point. We need to think about this UX more but introducing `global` in the database makes sense. 

@marckhouzam can you please help file an issue regarding installing plugins when there is a name conflict between the `global` and `tmc` targeted plugins?

PR #45, opened 2023-01-26
Central Repository: plugin install and sync

marckhouzam:

This commit creates a new OCI discovery implementation which handles the Central Repository format.

When enabling the "features.global.central-repository" feature flag, it is now possible install plugins from a Central Repository.

`tanzu plugin install`: working
`tanzu plugin clean`: working
`tanzu plugin upgrade`: working
`tanzu plugin describe`: working
`tanzu plugin list`: lists *all* plugins from the Central Repo.  #53 fixed this.
`tanzu plugin sync`: only syncs from current contexts, but not yet installing from Central repo.  This will be part of another PR, since it is a substantial change.

This first step fetches all plugins from the SQLite DB all the time. Then the plugin manager filters the list of plugins to find the correct one.  I took this approach to reduce the code changes necessary, but this will optimized in a follow-up effort.

I've tested this with a test Central Repository from #41.  You can follow these steps:
```
: Checkout PR 41 to start a test central repo locally in docker
make start-test-central-repo

: Checkout this PR
make build

: remove you configuration files to start fresh
tz config set features.global.central-repository true
tz plugin source add -n default -t oci -u localhost:9876/tanzu-cli/plugins/central:large
export ALLOWED_REGISTRY=localhost:9876

: Now you can start using the test Central Repo
tz plugin list # This will later not list everything
tz plugin install cluster
tz plugin install cluster --target tmc
tz tmc cluster

tz plugin install feature
tz  feature
tz k8s  feature

tz plugin list
tz plugin upgrade  feature
tz plugin delete  cluster --target tmc

tz plugin clean
```
To test plugin sync:

```
tz context create --name tmc-unstable --endpoint unstable.tmc-dev.cloud.vmware.com --staging
# See that only plugins from the context are synched, not all the ones in the central repo

tz context create --name defaultK3d --kubecontext k3d-k3s-default --kubeconfig /Users/kmarc/.config/k3d/k3s-default/kubeconfig.yaml
```

### Release note

```release-note
The Tanzu CLI now supports discovering plugins from a Centralized Repository of plugins.
All plugins can be found in that repository and are installed from that repository.
The synchronization of plugin (through "plugin sync" or setting a context) no longer
installs "standalone" plugins; instead a user must explicitly install such plugins
with the "plugin install" command.
```

### Additional information

1. There are no new unit test in this PR.  They will be added in a follow-up.
1. I'm not handling the test plugin that might be delivered with a plugin


pkg/discovery/oci_central.go
```diff
@@ -0,0 +1,246 @@
...
+func (od *OCIDiscoveryForCentralRepo) getPluginsFromDB(dbFilePath string) ([]Discovered, error) {
+	db, err := sql.Open("sqlite3", dbFilePath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to open the DB for discovery '%s'", od.Name())
+	}
+	defer db.Close()
+
+	// We need to order the results properly because the logic of processRows()
+	// expects an ordering of PluginName, then Target, then Version.
+	rows, err := db.Query("SELECT * FROM PluginBinaries ORDER BY PluginName,Target,Version")
```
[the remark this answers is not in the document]
marckhouzam: Excellent point!  Will update.

marckhouzam:
BTW, with this PR, you can easily manually access the DB of the Central Repo being used, if needed:
`sqlite3 ~/.cache/_tanzu/plugin_db/central.db`.

pkg/discovery/oci_central.go
```diff
@@ -0,0 +1,250 @@
...
+
+	// Convert from plugin pointers to plugins
+	// TODO(khouzam): continue optimizing by converting every call
+	// to using pointers
+	var allPlugins []Discovered
+	for _, pluginPtr := range allPluginPtrs {
+		allPlugins = append(allPlugins, *pluginPtr)
+	}
+	return allPlugins, nil
+}
```
anujc25, line 120: Should we put this into a separate golang package that handles all the SQL queries together?
We need to add more SQL queries that can be used at multiple places. Combining all SQL queries related stuff into separate packages might be better. WDYT? @marckhouzam

marckhouzam: That is a good idea.  I will work something out and update.

marckhouzam:
I've added a commit to this PR to so that `tz plugin sync` now works as expected by only synching plugins provided by the context.

marckhouzam:
> I've added a commit to this PR to so that `tz plugin sync` now works as expected by only synching plugins provided by the context.

Actually, this is incomplete for the Central Repo feature.  Context plugins need to be installed from the Central Repo, which this commit does not.  I will update.

pkg/discovery/oci_central.go
vuil: +1 to factoring out the db access from the logic of obtaining the 'db' via the OCI registry to minimize churn should we elect to use a different db backend in the future.

marckhouzam: I've pushed a commit to address this.

The approach I took is to create a new `DiscoveryBackend` interface. A `DiscoveryBackend` represents the content management logic behind a plugin `Discovery`.  The commit also creates an implementation of this new `DiscoveryBackend` for the SQLite backend of the Central Repository.

Note: The `DiscoveryBackend` support is included in the "discovery" package itself at the moment.  This is so that it can use the `Discovered` type.  Making the new `DiscoveryBackend` part of its own package would prevent it from using the `Discovered` type due to a cyclical dependency.  This would then force us to create yet another type to characterize plugins for the `DiscoveryBackend` to use.  This new type would then need to be converted to the `Discovered` type by the `Discovery`.

An alternative would be to move the `Discovered` type to its own package that could be imported by both the "discovery" package, and a new "discoverybackend" package.

We should discuss this.

pkg/fakes/registy.go
```diff
@@ -64,6 +66,11 @@ type Registry struct {
 	invocationsMutex sync.RWMutex
 }
 
+// TODO(khouzam)
+func (fake *Registry) DownloadImage(arg1 string, arg2 string) error {
```
prkalle, line 70: nit: you should be able to generate the fakes using `make generate-fakes`

prkalle (commented):
The overall changes look good. Thanks!

pkg/discovery/sqlite_backend.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/vmware-tanzu/tanzu-cli/pkg/catalog"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/distribution"
+	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
+)
+
+const (
+	// sqliteDBFileName is the name of the DB file that is stored in
+	// the OCI image describing the content of the Central Repository
+	sqliteDBFileName = "central.db"
```
vuil, line 27: Thanks for updating the Discovery name and factoring out the db access bit.

nit: (If we want we can address in a followup), I see there is still a fair bit of references to central repository, 
in comments, types and variables such as this. 
I think presenting the implementation more generally does not preclude using it specifically for the Central Repo, but painting this as meant for central repo then finding we might have to use it in other scenarios will lead to more awkwardness. 

e.g. the db file here could be named something like plugins.db or plugin_inventory.db and still make sense.

pkg/pluginmanager/manager.go
```diff
@@ -681,10 +691,19 @@ func DeletePlugin(options DeletePluginOptions) error {
 	return nil
 }
 
-// SyncPlugins automatically downloads all available plugins to users machine
+// SyncPlugins automatically installs all plugins of the current contexts
```
vuil, line 694: nit: comment is only correct if the CR feature is enabled.

pkg/pluginmanager/manager.go
```diff
@@ -681,10 +691,19 @@ func DeletePlugin(options DeletePluginOptions) error {
...
 func SyncPlugins() error {
 	log.Info("Checking for required plugins...")
-	plugins, err := AvailablePlugins()
+	var plugins []discovery.Discovered
+	var err error
+	if configlib.IsFeatureActivated(constants.FeatureCentralRepository) {
+		plugins, err = DiscoverServerPlugins()
```
vuil, line 700: I understand the rationale behind this, but it might be good to put a comment to explain why.
Also, this being a rather significant user-facing change. Please make sure to summarize it in a sentence or two under "Release Notes" part of the PR since at some point we will turn on the tooling to aggregate release notes in our releases.

vuil (approved):
lgtm, thanks for the refactoring and renaming. 
I have some nits that are not blocking

pkg/fakes/registy.go
marckhouzam: I didn't know that.  Thanks @prkalle!  I will update.

pkg/discovery/sqlite_backend.go
marckhouzam: Agreed.  I will clean this up.  BTW I much prefer your use of the term "inventory" instead of "content" that I've been using everywhere.  Will update that too 👍

pkg/pluginmanager/manager.go
marckhouzam: What do you mean by this?  Do you mean, "if a CR exists for a plugin", or is there an actual feature-flag for this feature?  Maybe you are referring to the context-aware feature-flag?

pkg/discovery/sqlite_backend.go
marckhouzam: I've cleaned all that up and removed all references to a "central repo", except in the commit text and the feature-flag name.

pkg/pluginmanager/manager.go
vuil: Sorry I should have been clearer. It looks like the comment only applies to the case where
`configlib.IsFeatureActivated(constants.FeatureCentralRepository)` is true. When false we install more than just plugins of the current contexts.

marckhouzam: Of course. Will update. Thanks!

pkg/discovery/oci_central.go
marckhouzam: We have discussed this and we have agreed that we will create a new type that the `DiscoveryBackend` can return, so as to decouple the `DiscoveryBackend` from the `discovery` package.
This will be done in a follow-up PR to allow this PR to get in now.

pkg/pluginmanager/manager.go
marckhouzam: Done

pkg/pluginmanager/manager.go
marckhouzam: Oh, I understand why I was confused.  I interpreted "CR" to mean "custom resource" when you meant "Central Repository" 😄 .  We're going to have to be careful about that one 🤯

pkg/carvelhelpers/fetcher.go
```diff
@@ -47,6 +47,22 @@ func DownloadImageBundleAndSaveFilesToTempDir(imageWithTag string) (string, erro
 	return tmpDir, nil
 }
 
+// DownloadDBImage reads a plain OCI image and saves its
+// files to the specified location.
+func DownloadDBImage(imageWithTag, destinationDir string) error {
```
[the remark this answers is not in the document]
marckhouzam: Good catch.  To be consistent with the other function `DownloadImageBundleAndSaveFilesToTempDir()` I have renamed this one `DownloadImageAndSaveFilesToDir()`

PR #47, opened 2023-01-27
Remove "Describe" from the Discovery interface

### LINKED ISSUE #43: Discovery interface does not consider target concept
**Bug description**

The plugin discovery interface provides a `Describe(name)` function to obtain a single plugin from the discovery as can be seen here:
https://github.com/vmware-tanzu/tanzu-cli/blob/45b251cecc87cabe3507b0817ef70c3056662dc9/pkg/discovery/interface.go#L26

With the introduction of "targets" in the CLI, a plugin is now uniquely identified by a name-target pair.  The `Describe` interface must therefore be updated to match.

This is not technically a problem at the moment  because a plugin discovery does not normally contain plugins for different targets.  For example a TKG OCI discovery will only contain plugins for the the `k8s` target, while a TMC REST discovery will only contain plugins for the `tmc` target. 

But this is important to fix for the Central Repository which will hold plugins from different targets.

**Expected behavior**

When calling the `Describe` function on a discovery, it needs to be possible to specify the target.

**Output of `tanzu version`**

Today's main: 6ab6177d

marckhouzam:

A plugin is now associated with a target, so to uniquely identify a plugin, we normally need a name-target combination.

The Describe function of the Discovery interface only took a plugin name so if the discovery contained plugins using the same name for two different targets, the function would not know which plugin to return.  Since the return value is a single plugin, we didn't have the option to return all plugins of a certain name.

We considered adding a "target" parameter to the Describe function signature.  However, this was a little complex for a REST discovery as its current endpoint support did not explicitly support the specification of a target.

Since the Describe function is actually never used, we decided that instead of spending time fixing these issues, we would simply remove it.

Fixes #43

make all

### DIFF
--- a/pkg/discovery/rest_test.go
+++ b/pkg/discovery/rest_test.go
@@ -5,7 +5,6 @@ package discovery
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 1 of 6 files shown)


anujc25 (approved):
LGTM. Thanks

pkg/discovery/rest_test.go
```diff
@@ -114,35 +113,14 @@ func createTestServer() *httptest.Server {
...
-	})
 	return httptest.NewServer(m)
 }
 
 func TestRESTDiscovery(t *testing.T) {
 	s := createTestServer()
 	defer s.Close()
 
-	d := NewRESTDiscovery("test", s.URL, basePath)
+	ddd := NewRESTDiscovery("test", s.URL, basePath)
```
marckhouzam, line 123: Woops, I missed reverting this change.  Will do now.

PR #54, opened 2023-02-01
Use v0.0.1 for test plugins to match TMC

marckhouzam:

When creating a staging context for TMC, the version of the plugins recommended by the endpoint is `v0.0.1`.  We make that version available in the test Central Repos to be able to test installing those plugins.

This allows to test the `tanzu plugin sync` scenario, or `tanzu context create` which will trigger a sync.


vuil (approved):
Thanks, lgtm. because all other things being equal, this improves the test/dev experience when interacting with TMC endpoints. That said,
wouldn't we have a similar situation with respect the TKG discovered plugins?

marckhouzam:
> That said, wouldn't we have a similar situation with respect the TKG discovered plugins?

Agreed.  I've been improving the test central repo incrementally as testing needs arise, but yes, we'll probably need a similar change for TKG soon.

PR #64, opened 2023-02-10
Provide smarter "plugin install/upgrade" for Central Repository Discovery

marckhouzam:

When the Central Repo feature is enabled, we use an OCI discovery with an SQLite PluginInventory.
Before this PR, installing/upgrading a plugin was done by getting **all** plugins, and then looping through the list in Go to find the desired one.

This PR uses the SQLite DB to do a targeted search, and only fetch the information for the desired plugin by filtering by name, target, version, os and arch.

There are actually no user-visible changes due to this change: it is just an improvement on the efficiency of the implementation.

Actually, there is one user-visible change: `tz plugin install all` is no longer supported when the Central Repo feature is enabled.  It does not make sense to intall **all** plugins present in the Central Repo.  An error message is shown to the user if the `all` argument is used (see testing section).

To introduce this filtering functionality in a generic way, two generic filtering concepts are introduced, one for the `Discovery` (a `DiscoveryCriteria`) and one for the `PluginInventory` (a `PluginInventoryFilter`).  To clarify, this is what we use to install/upgrade plugin:
```
 _______________     _______________     _________________
|               |   |               |   |                 |
| PluginManager |-->| OCI Discovery |-->| PluginInventory |                
|_______________|   |_______________|   |_________________|
```
So, with this PR:
1. the `PluginManager` specifies a new `DiscoveryCriteria` when creating the OCI Discovery for the Central Repo.  The criteria tells the Discovery what plugins should be discovered (by name, target, version, os, arch).
1. the OCI Discovery converts the `DiscoveryCriteria` into a similar `PluginInventoryFilter`, which it passes to the `PluginInventory`
1. the `PluginInventory` (which is an SQLite engine at the moment) uses the `PluginInventoryFilter` to write specific SQL queries that respect the `PluginInventoryFilter`.

**Will be filled soon**

### Release note

```release-note
The command `tanzu plugin install` no longer accepts the `all` argument.
```

### Additional information

1. Tests are added for the `plugininventory` package (91% coverage)
1. I would like to discuss the best way to add tests for the changes to the `PluginManager`, which will need to stub the Central Repo somehow.
2. The PR also assumes that the Central Repo has the `RecommendedVersion` field empty for the moment.
3. The `verifyPluginPreDownload(..)` function now verifies the repo using the requested `version` of the plugin, instead of always using the `recommendedVersion`.  This should cause not change in functionality IIUC.  It is required because with a more targeted query of plugins, the recommendedVersion artifact is not always fetched.  For example, if installing an older version of the `cluster` plugin, say `v0.26.0`, the "smart" install will only fetch information about that version; so, if we were to attempt to `verifyPluginPreDownload()` using the RecommendedVersion of `v0.28.0`, that information would not be available, and would need to be fetched separately from the SQLite DB; so, instead, we just use the requested version.


marckhouzam:
I got some good feedback from @vuil and will rework some aspects of the PR to be more generic and facilitate testing at the same time.
I converted the PR to a draft until that work is done.

marckhouzam:
This PR is now ready for review 😌 .  It kind of exploded in size with the unit tests, but the non-test code changes only account for 32% of the PR.

pkg/command/plugin.go
```diff
@@ -259,16 +266,23 @@ func newUpgradePluginCmd() *cobra.Command {
...
+				if err != nil {
+					return err
+				}
 			}
 
 			err = pluginmanager.UpgradePlugin(pluginName, pluginVersion, getTarget())
 			if err != nil {
 				return err
 			}
-			log.Successf("successfully upgraded plugin '%s' to version '%s'", pluginName, pluginVersion)
```
marckhouzam, line 271: The "version" info is removed from this printout because with the more efficient SQL DB, we don't know which version we are upgrading too here, we just as to install the `recommendedVersion`, whatever that version is.

This change in printout is not a big deal because it is preceded by another printout mentioning the version:
```
ℹ  Installing plugin 'management-cluster:v0.0.1' with target 'mission-control'
✔  successfully upgraded plugin 'management-cluster'
```

pkg/command/plugin_test.go
```diff
@@ -198,3 +200,126 @@ func TestDeletePlugin(t *testing.T) {
...
+			expectedFailure:    true,
+			expectedErrorMsg:   "the 'all' argument is not longer supported. You must install each plugin individually",
+		},
+		{
+			test:               "invalid target",
+			centralRepoFeature: "true",
+			args:               []string{"plugin", "install", "--target", "invalid", "myplugin"},
+			expectedFailure:    true,
+			expectedErrorMsg:   "invalid target specified. Please specify correct value of `--target` or `-t` flag from 'kubernetes/k8s/mission-control/tmc'",
+		},
```
marckhouzam, line 225: The above tests are only negative tests.
It is complicated to actually test installing a plugin from this part of the code, which I assume is why there was never a test for in this file.

pkg/discovery/interface.go
```diff
@@ -26,11 +26,27 @@ type Discovery interface {
...
+	Version string
+	// OS of the plugin binary in `GOOS` format.
+	OS string
+	// Arch of the plugin binary in `GOARCH` format.
+	Arch string
+}
+
 // CreateDiscoveryFromV1alpha1 creates discovery interface from v1alpha1 API
-func CreateDiscoveryFromV1alpha1(pd configtypes.PluginDiscovery) (Discovery, error) {
+func CreateDiscoveryFromV1alpha1(pd configtypes.PluginDiscovery, criteria *PluginDiscoveryCriteria) (Discovery, error) {
```
marckhouzam, line 45: I don't love adding the `criteria` here and not respecting it for any discovery other than OCI.
However, the go code using the discoveries does filter the plugin returned, so this is safe.

pkg/discovery/oci_dbbacked.go
```diff
@@ -30,6 +31,18 @@ type DBBackedOCIDiscovery struct {
...
 	// This image contains a single SQLite database file.
 	image string
+	// criteria specified different conditions that a plugin must respect to be discovered.
+	// This allows to filter the list of plugins that will be returned.
+	criteria *PluginDiscoveryCriteria
+	// pluginDataDir is the location where the plugin data will be stored once
+	// extracted from the OCI image
+	pluginDataDir string
+	// inventory is the pluginInventory to be used by this discovery.
+	inventory plugininventory.PluginInventory
```
marckhouzam, line 41: The use of this field allows the unit tests to change the `inventory` to use a stub.

pkg/discovery/oci_dbbacked_test.go
```diff
@@ -0,0 +1,411 @@
...
+	},
+}
+
+type stubInventory struct{}
+
+func (stub *stubInventory) GetAllPlugins() ([]*plugininventory.PluginInventoryEntry, error) {
+	return pluginEntries, nil
+}
+
+// nolint: gocyclo
```
marckhouzam, line 213: This stub filtering function is too complex for the linter 😄 
But I had spent enough time getting those tests to run, so I just disabled the linter for it.

This is why it is better to use SQL than our own Go logic for this smart filtering 😉

pkg/pluginmanager/manager.go
```diff
@@ -1015,8 +1019,8 @@ func getPluginInfoResource(pluginFilePath string) (*cli.PluginInfo, error) {
 
 // verifyPluginPreDownload verifies that the plugin distribution repo is trusted
 // and returns error if the verification fails.
-func verifyPluginPreDownload(p *discovery.Discovered) error {
-	artifactInfo, err := p.Distribution.DescribeArtifact(p.RecommendedVersion, runtime.GOOS, runtime.GOARCH)
+func verifyPluginPreDownload(p *discovery.Discovered, version string) error {
+	artifactInfo, err := p.Distribution.DescribeArtifact(version, runtime.GOOS, runtime.GOARCH)
```
marckhouzam, line 1023: @anujc25 Please pay attention to this change as it is not limited to the new Central Repo feature.
I believe it is safe, but I would appreciate your confirmation.

The reason is:
The verifyPluginPreDownload(..) function now verifies the repo using the requested version of the plugin, instead of always using the recommendedVersion. This should cause not change in functionality IIUC. It is required because with a more targeted query of plugins, the recommendedVersion artifact is not always fetched. For example, if installing an older version of the cluster plugin, say v0.26.0, the "smart" install will only fetch information about that version; so, if we were to attempt to verifyPluginPreDownload() using the RecommendedVersion of v0.28.0, that information would not be available, and would need to be fetched separately from the SQLite DB; so, instead, we just use the requested version.

pkg/pluginmanager/manager.go
```diff
@@ -1091,3 +1095,24 @@ func FindVersion(recommendedPluginVersion, requestedVersion string) string {
...
+
+// getPluginDiscoveries returns the plugin discoveries found in the configuration file.
+func getPluginDiscoveries() ([]configtypes.PluginDiscovery, error) {
+	if configlib.IsFeatureActivated(constants.FeatureCentralRepository) {
+		pd, err := getPreReleasePluginDiscovery()
+		if err != nil {
+			return nil, err
+		}
+		return []configtypes.PluginDiscovery{pd}, nil
+	}
```
marckhouzam, line 1107: It is the above block that uses the `TANZU_CLI_PRE_RELEASE_REPO_IMAGE` variable.
If we comment this, we can successfully run the new `Test_InstallPlugin_InstalledPlugins_Central_Repo` unit test.

marckhouzam (commented):
Some points to consider in-lined in the code.

pkg/discovery/oci.go
```diff
@@ -29,11 +32,23 @@ type OCIDiscovery struct {
...
-func NewOCIDiscovery(name, image string) Discovery {
+func NewOCIDiscovery(name, image string, criteria *PluginDiscoveryCriteria) Discovery {
 	if config.IsFeatureActivated(constants.FeatureCentralRepository) {
+		// The plugin inventory uses relative image URIs to be future-proof.
+		// Determine the image prefix from the main image.
+		// E.g., if the main image is at project.registry.vmware.com/tanzu-cli/plugins/plugin-inventory:latest
+		// then the image prefix should be project.registry.vmware.com/tanzu-cli/plugins/
+		imagePrefix := path.Dir(image)
+		// The data for the inventory is stored in the cache
+		pluginDataDir := filepath.Join(common.DefaultCacheDir, inventoryDirName, name)
```
marckhouzam, line 43: Notice this change where the inventory SQLite DB is moved from:
`~/.cache/_tanzu/plugin_inventory/plugin_inventory.db` to
`~/.cache/_tanzu/plugin_inventory/<discoveryName>/plugin_inventory.db`

This is to allow having multiple discovery sources using SQLite at the same time.

@anujc25 Does that fits with your publication approach?

anujc25: Yes. This seems correct.

anujc25: Yes. This seems reasonable.

pkg/plugininventory/sqlite_inventory.go
```diff
@@ -56,40 +63,127 @@ type inventoryDBRow struct {
...
+			default:
+				return whereClause, fmt.Errorf("invalid target for plugin: %s", string(filter.Target))
+			}
+
+			whereClause = fmt.Sprintf("%s Target='%s' AND", whereClause, target)
+		}
+		if filter.Version != "" {
+			if filter.Version == cli.VersionLatest {
+				// We want the recommended version of the plugin
+				whereClause = fmt.Sprintf("%s Version=RecommendedVersion AND", whereClause)
```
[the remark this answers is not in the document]
anujc25: We can keep it as it but please add a commit specifying the details and mention that this is a dead code that might be useful in the future.

anujc25 (approved):
LGTM. Thanks for the detailed PR description and testing section.

PR #73, opened 2023-02-20
Use latest `tanzu-plugin-runtime` and add `Target` field to the PluginDescriptor

anujc25:

This change updates the PluginDescriptor to provide `target` information as part of the contract as required by tanzu-plugin-runtime ([tanzu-plugin-runtime/pull/17](https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/17))

Also renames `TargetNone` to `TargetUnknown`.

Fixes #

- Existing Unit test should cover the changes made

### Release note

```release-note
N/A
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/tj/assert v0.0.3
 	github.com/vmware-tanzu/carvel-ytt v0.40.0
 	github.com/vmware-tanzu/tanzu-framework/capabilities/client v0.0.0-20230130173350-eeda69d80a24
-	github.com/vmware-tanzu/tanzu-plugin-runtime v0.0.0-20230221194443-4d4c14fcdb5c
+	github.com/vmware-tanzu/tanzu-plugin-runtime v0.0.0-20230223175026-9c303222c664
 	go.pinniped.dev v0.20.0
 	go.uber.org/multierr v1.8.0
 	golang.org/x/mod v0.7.0
--- a/pkg/command/plugin.go
+++ b/pkg/command/plugin.go
@@ -178,7 +178,7 @@ func newDescribePluginCmd() *cobra.Command {
 			}
 			pluginName := args[0]
 
-			if !configtypes.IsValidTarget(targetStr) {
+			if !configtypes.IsValidTarget(targetStr, true, true) {
 				return errors.New("invalid target specified. Please specify correct value of `--target` or `-t` flag from 'kubernetes/k8s/mission-control/tmc'")
 			}
 
@@ -209,7 +209,7 @@ func newInstallPluginCmd() *cobra.Command {
 
 			pluginName := args[0]
 
-			if !configtypes.IsValidTarget(targetStr) {
+			if !configtypes.IsValidTarget(targetStr, true, true) {
 				return errors.New("invalid target specified. Please specify correct value of `--target` or `-t` flag from 'kubernetes/k8s/mission-control/tmc'")
 			}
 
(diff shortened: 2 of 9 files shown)


go.mod
```diff
@@ -2,7 +2,10 @@ module github.com/vmware-tanzu/tanzu-cli
 
 go 1.19
 
-replace cloud.google.com/go => cloud.google.com/go v0.102.1
+replace (
+	cloud.google.com/go => cloud.google.com/go v0.102.1
+	github.com/vmware-tanzu/tanzu-plugin-runtime => github.com/anujc25/tanzu-plugin-runtime v0.0.0-20230220221618-695e5b2b7d7b
```
anujc25, line 7: Update this once https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/17 is merged.

pkg/command/plugin.go
```diff
@@ -161,7 +161,7 @@ func newDescribePluginCmd() *cobra.Command {
 			}
 			pluginName := args[0]
 
-			if !configtypes.IsValidTarget(targetStr) {
+			if !configtypes.IsValidTarget(targetStr, true) {
 				return errors.New("invalid target specified. Please specify correct value of `--target` or `-t` flag from 'kubernetes/k8s/mission-control/tmc'")
```
marckhouzam, line 165: I see you haven't updated these error message.  I assume it is because a user should never have to specify the "global" target?  The only time we could have a target conflict is between "tmc" and "k8s", since we agree that if a plugin is for the "global" target, it can't be for any other target.

However, it is possible for someone to write a plugin for "k8s" and another for "global" and use the same name, and then install them using `--local` (or with different plugin sources).  I suggest we open a separate issue as there are other protections we should probably put in place for that case: right now I believe we'll try to install both.

marckhouzam (changes requested):
1. Would you mind changing the main line of the commit message to be more descriptive and instead put the link in the body of the message?
2. Minor: I just noticed `context list` can take the `--target` flag.  Does it make sense to allow `tanzu context list --target global`? 

I will "request changes" so that we don't forget to update the go.mod file before merging.

pkg/command/plugin.go
anujc25: I was thinking we can fix this as a separate fix.

I was going to ping you for this one as well. I remember we discussed a possible issue where a plugin is a `global` plugin and the same-named plugin is available under the `tmc` target but was not possible to install.

Have we filled an issue to track that? I was planning to fix this as part of the separate bug fix that we discovered earlier.

anujc25:
> Minor: I just noticed context list can take the `--target` flag. Does it make sense to allow `tanzu context list --target global`

No. The context list cannot specify the target as `global` as `global` is a special target. It is not associated with any context.

marckhouzam:
> > Minor: I just noticed context list can take the `--target` flag. Does it make sense to allow `tanzu context list --target global`
> 
> No. The `tanzu context list` cannot specify the target as `global` as `global` is a special target. It is not associated with any context.

Sorry, I didn't formulate my question clearly.
With this PR, I am able to do `tanzu context list --target global`.  I just get an empty output, no error. I was wondering if we should throw an error if the user does this, or if the empty output is sufficient?

anujc25:
> Sorry, I didn't formulate my question clearly.
With this PR, I am able to do tanzu context list --target global. I just get an empty output, no error. I was wondering if we should throw an error if the user does this, or if the empty output is sufficient?

Thanks for the clarification. It makes sense to me now. I have updated the code to do proper validation for the `tanzu context list`.

pkg/command/plugin.go
anujc25: Filled: https://github.com/vmware-tanzu/tanzu-cli/issues/77

marckhouzam (approved):
LGTM.  Thanks!

PR #78, opened 2023-02-23
Add support for Plugin Groups

marckhouzam:
### What this PR does / why we need it

This PR adds support for Plugin Groups to the CLI.

The first commit adds some test Plugin Groups to the local test central repository.  By running `make start-test-central-repo`, the image at `localhost:9876/tanzu-cli/plugins/central:small` or `localhost:9876/tanzu-cli/plugins/central:large` will provide some Plugin Groups.

The rest of the PR adds support for the CLI to:
1. list the plugin groups using `tanzu plugin group search`
1. install the plugins in a group using `tanzu plugin install <plugin | all> --group <groupID>`
2. 'all' is the default behaviour when using groups, so we can do `tanzu plugin install --group <groupID>` (without 'all')

### Describe testing done for PR

```
$ rm ~/.config/tanzu/config*
$ tz plugin clean
✔  successfully cleaned up all plugins
$ tz config set features.global.central-repository true
$ make start-test-central-repo
$ export TANZU_CLI_PRE_RELEASE_REPO_IMAGE=localhost:9876/tanzu-cli/plugins/central:small

$ tz plugin list
Installed Plugins
  NAME  DESCRIPTION  TARGET  VERSION  STATUS

# List plugin groups found in the local test central repo
$ tz plugin group search
  GROUP
  vmware-tkg/v1.0.0
  vmware-tkg/v2.1.0
  vmware-tmc/v1.2.3
  vmware-tmc/v9.0.0
# List groups in yaml format
$ tz plugin group search -o yaml
- group: vmware-tkg/v1.0.0
- group: vmware-tkg/v2.1.0
- group: vmware-tmc/v1.2.3
- group: vmware-tmc/v9.0.0

# Install all plugins in the group
$ tz plugin install all --group vmware-tkg/v2.1.0
ℹ  Installing plugin 'cluster:v9.9.9' with target 'kubernetes'
ℹ  Installing plugin 'feature:v9.9.9' with target 'kubernetes'
ℹ  Installing plugin 'kubernetes-release:v9.9.9' with target 'kubernetes'
ℹ  Installing plugin 'management-cluster:v9.9.9' with target 'kubernetes'
ℹ  Installing plugin 'package:v9.9.9' with target 'kubernetes'
ℹ  Installing plugin 'secret:v9.9.9' with target 'kubernetes'
ℹ  Installing plugin 'telemetry:v9.9.9' with target 'kubernetes'
✔  successfully installed all plugins from group 'vmware-tkg/v2.1.0'
[...]

PR #85, opened 2023-03-01
Use `logger` from `tanzu-plugin-runtime` module and Update references

anujc25:

- As part of https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/24 a new logger was implemented as part of `tanzu-plugin-runtime` module
- This PR updates the tanzu-cli references to use this new logger from `tanzu-plugin-runtime`

Fixes #

### Release note

```release-note
Use `logger` from `tanzu-plugin-runtime` module
```


pkg/auth/csp/selfmanaged.go
```diff
@@ -85,7 +85,7 @@ func interruptHandler(d context.CancelFunc) {
 	signal.Notify(c, os.Interrupt)
 	for range c {
 		d()
-		log.Fatal("Login flow interrupted\n")
+		log.Fatal(errors.New("login flow interrupted"), "")
```
marckhouzam, line 88: Should we use the message parameter instead?
```suggestion
		log.Fatal(nil, "login flow interrupted")
```

pkg/auth/wcp/discovery.go
```diff
@@ -26,16 +26,16 @@ func IsVSphereSupervisor(endpoint string, httpClient *http.Client) (bool, error)
...
 	if err != nil {
-		log.Debug("Failed to test for vSphere supervisor: %+v", err)
+		log.Infof("Failed to test for vSphere supervisor: %+v", err)
 		return false, err
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode == http.StatusOK {
-		log.Debug("Got login banner from server")
+		log.Info("Got login banner from server")
```
marckhouzam, line 35: This does not look like an error path.  The old debug log will now always be printed IIUC.
Is there a way to reduce the level for this log?

pkg/command/context.go
```diff
@@ -359,7 +359,7 @@ func globalLogin(c *configtypes.Context) (err error) {
 		issuer = csp.StgIssuer
 	}
 	if apiTokenExists {
-		log.Debug("API token env var is set")
+		log.Info("API token env var is set")
```
marckhouzam, line 362: This one is not on an error path and will print when it didn't before.
But maybe in this case it is actually good, as it might be interesting to the user.  I'm personally ok with keeping this one.  Maybe @prkalle has an opinon.
```
$ export TANZU_API_TOKEN=allo
$ tz context create --name tmc-unstable --endpoint unstable.tmc-dev.cloud.vmware.com --staging
API token env var is set
[...]
```

pkg/command/login.go
```diff
@@ -374,7 +374,7 @@ func globalLoginUsingServer(s *configtypes.Server) (err error) {
 		issuer = csp.StgIssuer
 	}
 	if apiTokenExists {
-		log.Debug("API token env var is set")
+		log.Info("API token env var is set")
```
marckhouzam, line 377: Same case as for `context`, in case we decide to change anything in `context.go`

pkg/discovery/kubernetes.go
```diff
@@ -65,7 +65,7 @@ func (k *KubernetesDiscovery) GetDiscoveredPlugins(clusterClient cluster.Client)
 		if err != nil {
 			logMsg += err.Error()
 		}
-		log.Debug(logMsg)
+		log.Info(logMsg)
```
marckhouzam, line 68: This is potentially not on an error path (for clusters that don't have the CLIPLugin CRD) and will show all the time now

cmd/plugin/test/main.go
```diff
@@ -4,9 +4,9 @@
 package main
 
 import (
+	"log"
```
marckhouzam, line 7: Is this the right "log" import?

cmd/plugin/builder/command/cli_compile.go
```diff
@@ -26,11 +25,12 @@ import (
 
 	"github.com/vmware-tanzu/tanzu-cli/cmd/plugin/builder/types"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/cli"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
 )
 
 func init() {
 	// TODO, aunum is an old logging library, need to replace w/ klog, but we want timestamps immediately for build optimizations, so one time hack.
-	log.Timestamps = true
+	// log.SetTimestamp(true)
```
marckhouzam, line 33: So timestamps are gone now.  What did @vuil mean with his comment above?

marckhouzam (changes requested):
I'm glad we are trying to stop using the old logger before our 1.0 release.
However the changes have impacts on the friendliness of the output:
1. All the old `Success` printouts used to have a prefix ✔.  This is now gone.
2. Info printouts used to be prefixed with ℹ.  We may want to replace this with `[i]`
3. Warning logs used to be prefixed with !.  We may wan to replace it with `[!!]` (I put two `!` as it looked to close to `[i]` with a single one)
4. Errors used to be prefixed with ✖.  We may want to replace it with `[X]`

Do you also want to remove the use of the old logger for new plugins generated by the `builder` plugin?
That is just a couple of lines in:
- main.go.tmpl
- gomod.tmpl
- dotgolangci.txt.tmpl

cmd/plugin/builder/command/publish/oci.go
```diff
@@ -58,7 +58,7 @@ func (o *OCIPublisher) PublishPlugin(sourcePath, version, os, arch, plugin strin
 	if err != nil {
 		return "", errors.Wrapf(err, "%v", string(out))
 	}
-	log.Success("Successfully published plugin:", plugin)
+	log.Infof("Successfully published plugin: %v", plugin)
```
chandrareddyp, line 61: Can we start all logs in small letters to be consistent all places

cmd/plugin/builder/command/publish/oci.go
```diff
@@ -70,7 +70,7 @@ func (o *OCIPublisher) PublishDiscovery() error {
 	if err != nil {
 		return errors.Wrapf(err, "%v", string(out))
 	}
-	log.Success("Successfully published CLIPlugin resources to discovery image:", o.OCIDiscoveryImage)
+	log.Infof("Successfully published CLIPlugin resources to discovery image: %v", o.OCIDiscoveryImage)
```
chandrareddyp, line 73:
```suggestion
	log.Infof("successfully published CLIPlugin resources to discovery image: %v", o.OCIDiscoveryImage)
```

pkg/auth/csp/grpc.go
```diff
@@ -144,7 +145,7 @@ func GetAuthOptsOrExit() grpc.CallOption {
 	var err error
 	authOpts, err = WithCredentialDiscovery()
 	if err != nil {
-		log.Fatal("Not logged in. Please retry after logging in")
+		log.Fatal(errors.New("not logged in. Please retry after logging in"), "")
```
chandrareddyp, line 148: I would log the actual error along with messages, not sure why we are skipping the err

chandrareddyp (approved):
Thanks for fixing it; I added a few comments!
LGTM!

pkg/cli/runner.go
```diff
@@ -77,7 +77,7 @@ func (r *Runner) run(ctx context.Context, pluginPath string, stdout, stderr *byt
 		return fmt.Errorf("%q is a directory", pluginPath)
 	}
 
-	log.Debugf("running command path %s args: %+v", pluginPath, r.args)
+	log.Infof("running command path %s args: %+v", pluginPath, r.args)
```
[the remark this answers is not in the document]
anujc25: I have removed this log for now.

cmd/plugin/builder/command/cli_compile.go
```diff
@@ -303,7 +302,7 @@ func buildPlugin(path, id string) (plugin, error) {
 		p.modPath = ""
 	}
 
-	log.Debugy("plugin", p)
+	log.Infof("plugin %v", p)
```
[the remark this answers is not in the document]
anujc25: Updated to introduce timestamp:
```
cd cmd/plugin/builder && go build -o /Users/anujc/Documents/code/tanzu-cli/bin/builder .
build darwin-amd64 plugin with version: v0.0.2-dev
/Users/anujc/Documents/code/tanzu-cli/bin/builder plugin build --version v0.0.2-dev --path ./cmd/plugin --artifacts "/Users/anujc/Documents/code/tanzu-cli/artifacts-admin" --os-arch darwin_amd64
0302 15:57:38.437430 cli_compile.go:142] [ℹ] building local repository at /Users/anujc/Documents/code/tanzu-cli/artifacts-admin, v0.0.2-dev, [darwin_amd64]
0302 15:57:38.438058 cli_compile.go:224] [ℹ] 🦁 - building plugin at path "cmd/plugin/test"
0302 15:57:38.438135 cli_compile.go:224] [ℹ] 🐱 - building plugin at path "cmd/plugin/builder"
0302 15:57:39.838512 cli_compile.go:304] [ℹ] plugin {{builder Build Tanzu components global v0.0.2-dev   Admin  false 0 []  [] <nil>} cmd/plugin/builder  cmd/plugin/builder/README.md  🐱 global}
0302 15:57:39.838633 cli_compile.go:340] [ℹ] 🐱 - $ /usr/local/go/bin/go build -o /Users/anujc/Documents/code/tanzu-cli/artifacts-admin/darwin/amd64/global/builder/v0.0.2-dev/tanzu-builder-darwin_amd64 -ldflags  -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v0.0.2-dev' -tags  ./cmd/plugin/builder
0302 15:57:43.339098 cli_compile.go:304] [ℹ] plugin {{test Test the CLI global v0.0.2-dev   Admin  false 0 []  [] <nil>} cmd/plugin/test cmd/plugin/test/test cmd/plugin/test/README.md  🦁 global}
0302 15:57:43.339217 cli_compile.go:340] [ℹ] 🦁 - $ /usr/local/go/bin/go build -o /Users/anujc/Documents/code/tanzu-cli/artifacts-admin/darwin/amd64/global/test/v0.0.2-dev/tanzu-test-darwin_amd64 -ldflags  -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v0.0.2-dev' -tags  ./cmd/plugin/test
0302 15:57:53.147773 cli_compile.go:340] [ℹ] 🦁 - $ /usr/local/go/bin/go build -o /Users/anujc/Documents/code/tanzu-cli/artifacts-admin/darwin/amd64/global/test/v0.0.2-dev/test/tanzu-test-test-darwin_amd64 -ldflags  -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v0.0.2-dev' -tags  ./cmd/plugin/test/test
0302 15:57:53.850101 cli_compile.go:219] [✔] successfully built local repository
```

cmd/plugin/builder/command/cli_compile.go
anujc25: Introduced timestamp here.

anujc25:
> the changes have impacts on the friendliness of the output:
> 
> 1. All the old `Success` printouts used to have a prefix ✔.  This is now gone.
> 2. Info printouts used to be prefixed with ℹ.  We may want to replace this with `[i]`
> 3. Warning logs used to be prefixed with !.  We may wan to replace it with `[!!]` (I put two `!` as it looked to close to `[i]` with a single one)
> 4. Errors used to be prefixed with ✖.  We may want to replace it with `[X]`
 

I have updated the logger implementation to include this prefix based on the log type.

> Do you also want to remove the use of the old logger for new plugins generated by the `builder` plugin? That is just a couple of lines in:
> 
> * main.go.tmpl
> * gomod.tmpl
> * dotgolangci.txt.tmpl

Fixed this as well.

cmd/plugin/builder/command/cli_compile.go
```diff
@@ -303,7 +301,7 @@ func buildPlugin(path, id string) (plugin, error) {
 		p.modPath = ""
 	}
 
-	log.Debugy("plugin", p)
+	log.V(4).Infof("plugin %v", p)
```
marckhouzam, line 304: This log is still showing all the time.  I think we need to set the `threshold` of the logger for the `V(x)` to work.

marckhouzam (changes requested):
Thanks for all the improvements!

I still feel this change is tricky.  

Now, all the logs are sent to `stderr`.  This makes sense to me if we are logging something that is not the normal output.  For example, it is good for printing an info or warning along with the output of `tz plugin list`.  However, I'm not sure it is so good for cases where the output _is_ the logs.

When doing `tz builder cli compile` for example, I feel like the normal output are the logs, so a user might find it strange that they are sent to `stderr`.  For instance, I may be interested to storing the build logs to compare them with a older run; I would do `tz builder cli compile > build.out`, but the logs would still appear on the screen now.

Also, I find the timestamp may need to special treatment.
Let's compare the `tz builder cli compile` output with what we have before:

Before
```
2023-03-02T22:27:01-05:00 [ℹ]  building local repository at artifacts, , [all]
2023-03-02T22:27:01-05:00 [ℹ]  🦁 - building plugin at path "cmd/plugin/test"
2023-03-02T22:27:01-05:00 [ℹ]  🐱 - building plugin at path "cmd/plugin/builder"
```
After:
```
0302 22:13:03.015035 cli_compile.go:142] [ℹ] building local repository at artifacts, , [all]
0302 22:13:03.016491 cli_compile.go:224] [ℹ] 🐱 - building plugin at path "cmd/plugin/builder"
0302 22:13:03.016512 cli_compile.go:224] [ℹ] 🦁 - building plugin at path "cmd/plugin/test"
```

1. I personally find the `month/day` format cryptic.  I didn't know what the `0302` string actually meant until I looked at the code.
2. I wonder if we should be showing the file:line info in this case.  These printouts are information for plugin devs and not for debugging the `builder` plugin.  This point is similar to the point about sending these printouts to `stdout` instead of `stderr`

cmd/plugin/builder/command/cli_compile.go
anujc25: Yeah. I thought the default threshold was set to `3` but it was not. Updated as part of https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/24.
This log will not be set unless the threshold is set to `>3`.

anujc25:
> Thanks for all the improvements!
> 
> I still feel this change is tricky.
> 
> Now, all the logs are sent to `stderr`. This makes sense to me if we are logging something that is not the normal output. For example, it is good for printing an info or warning along with the output of `tz plugin list`. However, I'm not sure it is so good for cases where the output _is_ the logs.
> 
> When doing `tz builder cli compile` for example, I feel like the normal output are the logs, so a user might find it strange that they are sent to `stderr`. For instance, I may be interested to storing the build logs to compare them with a older run; I would do `tz builder cli compile > build.out`, but the logs would still appear on the screen now.
> 

I have introduced the `log.SetStderr` and `log.SetStdout` functions to allow users to override the default logging location.
For example, if a user wants to write everything to `stdout` for `tz builder cli compile` command, the user will need to override the logger's stderr to stdout by configuring it with `log.SetStderr(os.Stdout)`.
Will this be sufficient interms of flexibility this logger provides?

> Also, I find the timestamp may need to special treatment.
> 
> 1. I personally find the `month/day` format cryptic.  I didn't know what the `0302` string actually meant until I looked at the code.
> 2. I wonder if we should be showing the file:line info in this case.  These printouts are information for plugin devs and not for debugging the `builder` plugin.  This point is similar to the point about sending these printouts to `stdout` instead of `stderr`

I have updated the timestamp format to use the `RFC` format. Also, I have hidden the `file:line` information from the logs we print to stdout/stderr. The `builder cli compile` logs will look like below:

```
$ make build-admin-plugins 
cd cmd/plugin/builder && go build -o /Users/anujc/Documents/code/tanzu-cli/bin/builder .
build darwin-amd64 plugin with version: v0.0.2-dev
/Users/anujc/Documents/code/tanzu-cli/bin/builder plugin build --version v0.0.2-dev --path ./cmd/plugin --artifacts "/Users/anujc/Documents/code/tanzu-cli/artifacts-admin" --os-arch darwin_amd64
2023-03-02T22:54:29-08:00 [i] building local repository at /Users/anujc/Documents/code/tanzu-cli/artifacts-admin, v0.0.2-dev, [darwin_amd64]
2023-03-02T22:54:29-08:00 [i] 🦁 - building plugin at path "cmd/plugin/test"
2023-03-02T22:54:29-08:00 [i] 🐱 - building plugin at path "cmd/plugin/builder"
2023-03-02T22:54:30-08:00 [i] 🐱 - $ /usr/local/go/bin/go build -o /Users/anujc/Documents/code/tanzu-cli/artifacts-admin/darwin/amd64/global/builder/v0.0.2-dev/tanzu-builder-darwin_amd64 -ldflags  -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v0.0.2-dev' -tags  ./cmd/plugin/builder
2023-03-02T22:54:34-08:00 [i] 🦁 - $ /usr/local/go/bin/go build -o /Users/anujc/Documents/code/tanzu-cli/artifacts-admin/darwin/amd64/global/test/v0.0.2-dev/tanzu-test-darwin_amd64 -ldflags  -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v0.0.2-dev' -tags  ./cmd/plugin/test
2023-03-02T22:54:41-08:00 [i] 🦁 - $ /usr/local/go/bin/go build -o /Users/anujc/Documents/code/tanzu-cli/artifacts-admin/darwin/amd64/global/test/v0.0.2-dev/test/tanzu-test-test-darwin_amd64 -ldflags  -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v0.0.2-dev' -tags  ./cmd/plugin/test/test
2023-03-02T22:54:41-08:00 [✔] successfully built local repository
```

marckhouzam:
@anujc25 This PR now needs to be rebased after I merged #78.  Could you also update the log usage that was added here: https://github.com/vmware-tanzu/tanzu-cli/blob/cb3bd4925c0dc444ab342bc96ff39f0c623af9c2/pkg/command/plugin_group.go#L71-L74

Thanks!

anujc25:
> @anujc25 This PR now needs to be rebased after I merged https://github.com/vmware-tanzu/tanzu-cli/pull/78. Could you also update the log usage that was added here.

Updated.

cmd/plugin/builder/command/cli_compile.go
```diff
@@ -191,7 +191,6 @@ func Compile(compileArgs *PluginCompileArgs) error {
 	close(plugins)
 	close(fatalErrors)
 
-	log.BreakHard()
```
marckhouzam, line 194: This was meant to separate the plugin build logs from any fatal errors for some plugins.
All it printed was `=========`.
How about adding our own log: `log.Outputf("========")`?
Note that such a log will go to `stdout`, so we should only add it if we agree to send all the logs of this function to `stdout`, as suggested in the global comments.

marckhouzam:
> I have updated the timestamp format to use the RFC format. Also, I have hidden the file:line information from the logs we print to stdout/stderr. The builder cli compile logs will look like below:
> `2023-03-02T22:54:29-08:00 [i] 🦁 - building plugin at path "cmd/plugin/test"`

Very nice! Thank you.

marckhouzam (commented):
> I have introduced the `log.SetStderr` and `log.SetStdout` functions to allow users to override the default logging location.
>For example, if a user wants to write everything to stdout for `tz builder cli compile` command, the user will need to override the logger's stderr to stdout by configuring it with `log.SetStderr(os.Stdout)`.
>Will this be sufficient in terms of flexibility this logger provides?

That sounds very reasonable to me.

So I think we should use it in the `builder` plugin 😄 
Could we therefore add the lines
```
// Redirect the logs to stdout for the user to consume
log.SetStderr(os.Stdout)
```
to the beginning of `cli_compile.go:Compile()`

marckhouzam:
> So I think we should use it in the `builder` plugin 😄 Could we therefore add the lines
> 
> ```
> // Redirect the logs to stdout for the user to consume
> log.SetStderr(os.Stdout)
> ```
> 
> to the beginning of `cli_compile.go:Compile()`

Actually, maybe we should this to the `init()` function of `cli_compile.go` so it would affect other commands besides the compilation?

pkg/command/context.go
vuil: I am okay with keeping, but
- if it is still somewhat debatable, maybe put it behind a low log level (<=3) instead of unconditionally showing
- would mentioning the env var name itself be more useful?

cmd/plugin/builder/command/cli_compile.go
anujc25: Replaced it with the `log.Info`.

anujc25:
> Actually, maybe we should this to the `init()` function of `cli_compile.go` so it would affect other commands besides the compilation?

Done.

marckhouzam (approved):
LGTM 🎉

PR #92, opened 2023-03-05
Root-targeted plugins now include "global" plugins

### LINKED ISSUE #91: Plugins with "global" target cannot be invoked
**Bug description**

We recently added an explicit "global" target in https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/17.
And the builder plugin started using it in #73 

However, the CLI was not updated to consider the new "global" target to invoke global plugins.  This means the new builder plugin can no longer be invoked.

**Expected behavior**

The builder plugin should be invokable.

```
$ git checkout main
$ make build-all
[...]
$ tz plugin install builder --local ./artifacts-admin/darwin/amd64
ℹ  Installing plugin 'builder:v0.0.2-dev' with target 'global'
✔  successfully installed 'builder' plugin

$ tz builder -h
✖  unknown command "builder" for "tanzu"
```

**Output of `tanzu version`**

version: v0.0.2-dev
buildDate: 2023-03-05
sha: 46474c69b

Local

marckhouzam:

Allows to invoke plugins that use the new "global" target.  For example the `builder` plugin.
Without this, the latest version of the `builder` plugin cannot be invoked from the CLI.

Fixes #91

Before the PR:
```
$ git checkout main
$ make build-all
[...]
$ tz plugin install builder --local ./artifacts-admin/darwin/amd64
ℹ  Installing plugin 'builder:v0.0.2-dev' with target 'global'
✔  successfully installed 'builder' plugin

$ tz builder -h
✖  unknown command "builder" for "tanzu"
```
After the PR:
```
$ make build-all
[...]
$ tz plugin install builder --local ./artifacts-admin/darwin/amd64
ℹ  Installing plugin 'builder:v0.0.2-dev' with target 'global'
✔  successfully installed 'builder' plugin

$ tz builder -h
Build Tanzu components

Usage:
  tanzu builder [command]
[...]
```

### Release note

```release-note

```

### Additional information

I added unit test to verify that the correct targets where invokable.
This (of course 🙄 ) triggered some unexpected failures which I had to chase down.  This is why `manager.go` has a small fix.

### DIFF
--- a/pkg/cli/plugin_cmd.go
+++ b/pkg/cli/plugin_cmd.go
@@ -26,6 +26,7 @@ func GetCmdForPlugin(p *PluginInfo) *cobra.Command {
 		DisableFlagParsing: true,
 		Annotations: map[string]string{
 			"group": string(p.Group),
+			"scope": p.Scope,
 		},
 		Hidden:  p.Hidden,
 		Aliases: p.Aliases,
--- a/pkg/pluginmanager/manager.go
+++ b/pkg/pluginmanager/manager.go
@@ -437,6 +437,7 @@ func DiscoveredFromPlugininfo(p *cli.PluginInfo) discovery.Discovered {
 		RecommendedVersion: p.Version,
 		Source:             p.Discovery,
 		SupportedVersions:  []string{p.Version},
+		Target:             p.Target,
 	}
 	return dp
 }
--- a/pkg/command/plugin_test.go
+++ b/pkg/command/plugin_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/constants"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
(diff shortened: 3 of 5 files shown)


pkg/cli/plugin_cmd.go
```diff
@@ -26,6 +26,7 @@ func GetCmdForPlugin(p *PluginInfo) *cobra.Command {
 		DisableFlagParsing: true,
 		Annotations: map[string]string{
 			"group": string(p.Group),
+			"scope": p.Scope,
```
marckhouzam, line 29: This was added to TF after we moved it to this repo:
https://github.com/vmware-tanzu/tanzu-framework/blob/8354312f13e95a96ccee234d3423ede33eeec09c/cli/core/pkg/cli/cmd.go#L31

pkg/pluginmanager/manager.go
```diff
@@ -437,6 +437,7 @@ func DiscoveredFromPlugininfo(p *cli.PluginInfo) discovery.Discovered {
 		RecommendedVersion: p.Version,
 		Source:             p.Discovery,
 		SupportedVersions:  []string{p.Version},
+		Target:             p.Target,
```
marckhouzam, line 440: The target was not being copied from one struct to the other, which caused some unexpected failures in the new unit tests.

anujc25 (approved):
LGTM. Thank for the fix.

PR #105, opened 2023-03-10
Enable the Central Repo feature by default

marckhouzam:

This PR enables the Central Repo feature by default.
See special notes for an explanation on the approach taken.

Check that the central repo feature is enabled by default:
```
$ rm ~/.config/tanzu/config*
$ tz plugin clean
$ tz config set env.TANZU_CLI_PRE_RELEASE_REPO_IMAGE localhost:9876/tanzu-cli/plugins/central:small

# Notice that the config file does not contain any indication of
# a feature flag for the Central Repo.
$ tz config get
clientOptions:
    cli:
        edition: tkg
        bomRepo: projects.registry.vmware.com/tkg
        compatibilityFilePath: tkg-compatibility
    features:
        global:
            context-target-v2: "true"
    env:
        TANZU_CLI_PRE_RELEASE_REPO_IMAGE: localhost:9876/tanzu-cli/plugins/central:small

# "plugin search" exists, this means the Central Repo feature is enabled
$ tz plugin search
  NAME                DESCRIPTION                  TARGET           VERSION  STATUS         CONTEXT
  isolated-cluster    Desc for isolated-cluster    global           v9.9.9   not installed
  pinniped-auth       Desc for pinniped-auth       global           v9.9.9   not installed
  cluster             Desc for cluster             kubernetes       v9.9.9   not installed
  feature             Desc for feature             kubernetes       v9.9.9   not installed
[...]
```

Test it is possible to disable the central repo feature.
Disable the feature and check that "tanzu plugin search" is no longer accessible.
```
$ tz config set features.global.no-central-repo-test-only true
$ tz plugin search
Manage CLI plugins

Usage:
  tanzu plugin [command]

Available Commands:
  clean       Clean the plugins
  delete      Delete a plugin
  describe    Describe a plugin
  install     Install a plugin
  list        List available plugins
  source      Manage plugin discovery sources
  sync        Sync the plugins
  upgrade     Upgrade a plugin

Flags:
  -h, --help   help for plugin

Use "tanzu plugin [command] --help" for more information about a command.
```
### Release note
```release-note
The Tanzu CLI now discovers and installs all its plugins from a Central Repository of plugins.  This simplifies the user experience and provides a uniform experience for all plugins. 
```

Re-enable the feature by turning the feature flag to "false":
```
$ tz config set features.global.no-central-repo-test-only false

# Notice the "plugin search" command is again accessible.
$ tz plugin search
  NAME                DESCRIPTION                  TARGET           VERSION  STATUS         CONTEXT
  isolated-cluster    Desc for isolated-cluster    global           v9.9.9   not installed
  pinniped-auth       Desc for pinniped-auth       global           v9.9.9   not installed
  cluster             Desc for cluster             kubernetes       v9.9.9   not installed
  feature             Desc for feature             kubernetes       v9.9.9   not installed
[...]
```

I have also performed some plugin lifecycle tests to make sure everything continues to work as before.
- plugin list
- context create
- plugin install
- plugin install --group

We want the central-repo feature to be enabled by default, but we want  a *hidden* way to disable it, just in case.  Therefore, we don't want the feature flag to be present in the config file unless we actually turn off the feature. This is important so that users don't think they can turn off the feature.

However, if a feature flag is not set, it defaults to false.  This means the central-repo feature would be disabled by default.  To work around that, we flip the feature flag to be a way to *disable* the central repo feature instead of enabling it like before.

So the feature flag becomes "no-central-repo-test-only":
    - false -> central-repo feature enabled
    - true  -> central-repo feature disabled

This change implies that in the code, every check for the feature flag has to be flipped to continue working like before.  You will see this in the code changes.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -37,9 +37,7 @@ const (
 	PluginSubCommand     = "tanzu %s"
 
 	// Central repository
-	CentralRepositoryFeatureFlag         = "features.global.central-repository"
-	CentralRepositoryPreReleaseRepoImage = "TANZU_CLI_PRE_RELEASE_REPO_IMAGE"
-	TanzuCliE2ETestCentralRepositoryURL  = "TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL"
+	TanzuCliE2ETestCentralRepositoryURL = "TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL"
 
 	// General constants
 	True      = "true"
--- a/hack/central-repo/README.md
+++ b/hack/central-repo/README.md
@@ -18,7 +18,6 @@ Also, only versions `v0.0.1` and `v9.9.9` for the other plugins can be installed
 The steps to follow to use the test central repo are:
 
 1. Start the test repo with `make start-test-central-repo`.
-1. Enable the central repository using the temporary feature flag: `tz config set features.global.central-repository true`
 1. Configure the plugin source for the test central repo: `tz plugin source add -n default -t oci -u localhost:9876/tanzu-cli/plugins/central:small`
 1. Allow the use of a local registry: `export ALLOWED_REGISTRY=localhost:9876`
 
@@ -28,7 +27,6 @@ Here are the exact commands:
 cd tanzu-cli
 make build
 make start-test-central-repo
-tz config set features.global.central-repository true
 tz plugin source add -n default -t oci -u localhost:9876/tanzu-cli/plugins/central:small
 export ALLOWED_REGISTRY=localhost:9876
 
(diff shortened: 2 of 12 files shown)


test/e2e/framework/framework.go
```diff
@@ -37,9 +37,7 @@ const (
 	PluginSubCommand     = "tanzu %s"
 
 	// Central repository
-	CentralRepositoryFeatureFlag         = "features.global.central-repository"
-	CentralRepositoryPreReleaseRepoImage = "TANZU_CLI_PRE_RELEASE_REPO_IMAGE"
```
marckhouzam, line 41: I also removed this constant which was not used.  We should not duplicate the string value here, but use the global value defined in the CLI.

PR #108, opened 2023-03-12
Use gzip instead of bzip2 for compression

marckhouzam:

bzip2 is not on the ubuntu image used by our CI, so it is easier to simply use gzip for compression.  The size of the registry-content.tgz is of 387K instead of the old 335K, which is not so bad.

```
make start-test-central-repo
\rm -rf hack/central-repo/registry-content
make start-test-central-repo
```


marckhouzam:
It turns out that using `tar xjf` instead of `tar xzf` is sufficient.  I guess the Github ubuntu image does contain `bzip2` unlike that dockerhub image...

So we don't need this change.

PR #111, opened 2023-03-14
Make the `artifacts` and `optional` fields in CLIPlugin CRD optional

anujc25:

- Make the `artifacts` and `optional` fields in CLIPlugin CRD optional.
- With central repository implementation, the plugin location specified under `artifacts` (for context-scoped plugin downloads), will be ignored completely and plugins always get downloaded from the default or configured central repository on the client side. The version of the plugin is selected based on the `recommendedVersion` specified with the CLIPlugin CR.

Fixes #

### Release note

```release-note
Update the `artifacts` and `optional` fields in `CLIPlugin` CRD to be optional
```


marckhouzam (approved):
LGTM
Thanks for thinking about this one!

PR #113, opened 2023-03-14
Add `--validate` flag to `builder inventory plugin add` command to validate plugin exists or not

anujc25:

- This PR adds a `--validate` flag to the `builder inventory plugin add` command to validate the plugin already exists or not in the plugin inventory database.
- It fails if the plugins already exist and succeeds if plugins don't exist in the database.

Fixes #

### Release note

```release-note
Add the `--validate` flag to the `builder inventory plugin add` command to validate plugin exists or not
```


marckhouzam (approved):
I think this is sufficient for now since I believe this `--validate` flag will be used in automation and not by humans.  If we want to support this for humans at a later time, it is just the error output that I don't like too much, as it says that it couldn't insert the plugins although we are just doing validation.

But like I said, this is not important for automation.

PR #136, opened 2023-03-28
Hide the "plugin source" commands and deprecate "config server"

marckhouzam:

For the alpha.0 release of the CLI, plugin discovery sources are not used to configure the Central Repository.

This commit hides the commands to avoid any confusion. These commands were not really used before anyway, so this should not even be noticed by alpha.0 users.

Plugin discovery sources will be enhanced and re-enabled after the alpha.0 release.

Further, the second commit deprecates the "config server" set of commands as they have been replaced with the concept of "context" and the corresponding "context" commands.  Notice that "tanzu login" has already been deprecated in the same fashion.

Fixes N/A

Test for `tanzu plugin source`:
```
# Notice the "source" command is no longer shown under "tanzu plugin"
$ tz plugin -h
Manage CLI plugins

Usage:
  tanzu plugin [command]

Available Commands:
  clean       Clean the plugins
  delete      Delete a plugin
  describe    Describe a plugin
  group       Manage plugin groups
  install     Install a plugin
  list        List available plugins
  search      Search for a keyword or regex in the list of available plugins
  sync        Sync the plugins
  upgrade     Upgrade a plugin

Flags:
  -h, --help   help for plugin

Use "tanzu plugin [command] --help" for more information about a command.

# Notice the "plugin source" command is still accessible
$ tz plugin source -h
Manage plugin discovery sources. Discovery source provides metadata about the list of available plugins, their supported versions and how to download them.

Usage:
  tanzu plugin source [command]

Available Commands:
  add         Add a discovery source
  delete      Delete a discovery source
  list        List available discovery sources
  update      Update a discovery source configuration

Flags:
  -h, --help   help for source

Use "tanzu plugin source [command] --help" for more information about a command.

# Notice the "plugin source" subcommand are still functional
$ tz plugin source list
  NAME  TYPE  SCOPE
```

Test for `tanzu config server`:
```
# Command "config server" is hidden
tz config -h
Configuration for the CLI

Usage:
  tanzu config [command]

Available Commands:
  get         Get the current configuration
  init        Initialize config with defaults
  set         Set config values at the given path
  unset       Unset config values at the given path

Flags:
  -h, --help   help for config

Use "tanzu config [command] --help" for more information about a command.

# Command still accessible
$  tz config server
Command "server" is deprecated, will be removed in version "1.2.0". Use the "context" set of commands instead.
Configured servers

Usage:
  tanzu config server [command]

Available Commands:
  delete      Delete a server from the config
  list        List servers

Flags:
  -h, --help   help for server

Use "tanzu config server [command] --help" for more information about a command.

# Command works
$ tz config server list
  NAME  TYPE               ENDPOINT  PATH                                    CONTEXT
  tkg1  managementcluster            /Users/kmarc/.k3d/kubeconfig-tkg1.yaml  tkg1
```

### Release note

```release-note
The `plugin source` commands are hidden since they have no effect at the moment.
```

### DIFF
--- a/pkg/command/config.go
+++ b/pkg/command/config.go
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
 
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/command"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/component"
 	configlib "github.com/vmware-tanzu/tanzu-plugin-runtime/config"
 	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
@@ -39,6 +40,7 @@ func init() {
 	)
 	serversCmd.AddCommand(listServersCmd)
 	addDeleteServersCmd()
+	command.DeprecateCommandWithAlternative(serversCmd, "1.2.0", "context")
 }
 
 var unattended bool
--- a/pkg/command/discovery_source.go
+++ b/pkg/command/discovery_source.go
@@ -28,9 +28,11 @@ var (
 
 func newDiscoverySourceCmd() *cobra.Command {
 	var discoverySourceCmd = &cobra.Command{
-		Use:   "source",
-		Short: "Manage plugin discovery sources",
-		Long:  "Manage plugin discovery sources. Discovery source provides metadata about the list of available plugins, their supported versions and how to download them.",
+		Use: "source",
+		// TODO(khouzam): not to be used for the alpha.0 release, but will be re-added after
+		Hidden: true,
+		Short:  "Manage plugin discovery sources",
+		Long:   "Manage plugin discovery sources. Discovery source provides metadata about the list of available plugins, their supported versions and how to download them.",
 	}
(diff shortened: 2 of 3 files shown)


pkg/command/config.go
```diff
@@ -243,8 +243,9 @@ var initConfigCmd = &cobra.Command{
 
 // Note: Shall be deprecated in a future version. Superseded by 'tanzu context' command.
 var serversCmd = &cobra.Command{
-	Use:   "server",
-	Short: "Configured servers",
+	Use:        "server",
+	Short:      "Configured servers",
+	Deprecated: fmt.Sprintf("will be removed in version %q. Use the %q command instead.", "1.2.0", "context"),
```
anujc25, line 248: This is fine, however, it would be nice if we can use https://github.com/vmware-tanzu/tanzu-plugin-runtime/blob/main/command/deprecation.go#L20 library to be consistent.

anujc25 (approved):
LGTM.

pkg/command/config.go
marckhouzam: Good point, I wasn't aware if this library.
I have updated both my change and the deprecation of the "login" command to use this lib.

PR #140, opened 2023-03-30
Configure staging central plugin repository as default plugin discovery

anujc25:

This change updates the CLI to use `projects.registry.vmware.com/tanzu_cli_stage/plugins/plugin-inventory:latest` as the default central plugin repository.

To override the default central repository, developers can configure `TANZU_CLI_PRE_RELEASE_REPO_IMAGE`. However, this is not an officially supported method.

CLI will by default point to the public staging central repository (projects.registry.vmware.com/tanzu_cli_stage/plugins/plugin-inventory:latest).
That means, 
- No need to configure TANZU_CLI_PRE_RELEASE_REPO_IMAGE 
- CLI is still be using this variable for our testing pipelines and unit tests but no external user should need to use it and it will not be documented.
- To get plugins from any additional discovery endpoint, please specify it with the TANZU_CLI_ADDITIONAL_PLUGIN_DISCOVERY_IMAGES_TEST_ONLY env variable.

Fixes #

```
$ tz plugin clean

$ tz plugin search
  NAME     DESCRIPTION             TARGET  VERSION                 STATUS         CONTEXT
  builder  Build Tanzu components  global  v0.1.0-dev-8-g6087ea00  not installed
  test     Test the CLI            global  v0.1.0-dev-8-g6087ea00  not installed

$ tz plugin install builder
[i] Installing plugin 'builder:v0.1.0-dev-8-g6087ea00' with target 'global'
t[ok] successfully installed 'builder' plugin

$ tz plugin list
Standalone Plugins
  NAME     DESCRIPTION             TARGET  VERSION                 STATUS
  builder  Build Tanzu components  global  v0.1.0-dev-8-g6087ea00  installed
```

### Release note

```release-note
Configure staging central plugin repository as default plugin discovery
```

### DIFF
--- a/pkg/pluginmanager/manager.go
+++ b/pkg/pluginmanager/manager.go
@@ -200,16 +200,18 @@ func DiscoverPluginGroups() ([]*discovery.DiscoveredPluginGroups, error) {
 }
 
 // getPreReleasePluginDiscovery
-// For our pre-releases we use an environment variable to point to the
-// repository of plugins.  This is because the configuration
-// cfg.ClientOptions.CLI.DiscoverySources
+// For pre-releases CLI points to a default staging central plugin discovery image
+// from where the CLI will discover plugins.
+// For pre-releases, CLI also allows default discovery image to be overridden by using
+// an environment variable and pointing to the different repository of plugins.
+//
+// This is because the configuration cfg.ClientOptions.CLI.DiscoverySources
 // is read by older CLIs so we don't want to modify it.
 // TODO(khouzam): remove before 1.0
 func getPreReleasePluginDiscovery() ([]configtypes.PluginDiscovery, error) {
 	centralRepoTestImage := os.Getenv(constants.ConfigVariablePreReleasePluginRepoImage)
 	if centralRepoTestImage == "" {
-		// Don't set a default value.  This test repo URI is not meant to be public.
-		return nil, fmt.Errorf("you must set the environment variable %s to the URI of the image of the plugin repository.  Please see the documentation", constants.ConfigVariablePreReleasePluginRepoImage)
+		centralRepoTestImage = constants.TanzuCLIDefaultCentralPluginDiscoveryImage
 	} else if centralRepoTestImage == PreReleasePluginRepoImageBypass {
 		return nil, nil
 	}
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -133,6 +133,11 @@ func GetTrustedRegistries() []string {
 		}
 	}
 
+	// Add default central plugin discovery image to trusted registries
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -233,7 +233,7 @@ start-test-central-repo: stop-test-central-repo ## Starts up a test central repo
 	fi
 	@docker run --rm -d -p 9876:5000 --name central \
 		-v $(ROOT_DIR)/hack/central-repo/registry-content:/var/lib/registry \
-		mirror.gcr.io/library/registry:2 > /dev/null && \
+		mirror.gcr.io/library/registry:2.7.1 > /dev/null && \
```
marckhouzam, line 236: Out of curiosity, why do you prefer to fix the version?  I believe using `:2` gets us the latest as I expect that tag to be updated.

pkg/pluginmanager/manager.go
```diff
@@ -200,16 +200,18 @@ func DiscoverPluginGroups() ([]*discovery.DiscoveredPluginGroups, error) {
 }
 
 // getPreReleasePluginDiscovery
-// For our pre-releases we use an environment variable to point to the
-// repository of plugins.  This is because the configuration
-// cfg.ClientOptions.CLI.DiscoverySources
+// For pre-releases CLI points to a default staging central plugin discovery image
+// from where the CLI will discover plugins.
+// For pre-releases, CLI also allows default discovery image to be overriden by using
```
marckhouzam, line 205: As suggested by the linter:
```suggestion
// For pre-releases, CLI also allows default discovery image to be overridden by using
```

marckhouzam:
> To override the default central repository, developers can configure TANZU_CLI_PRE_RELEASE_REPO_IMAGE. However, this is not an officially supported method.

I assume you chose to keep this variable because it is being used by the unit tests to bypass the central repo?  Getting rid of the variable will require reworking this bypassing approach which will require more time.  If that is the case, I agree in your approach.

I just want to confirm if you plan is:
1- don't document TANZU_CLI_PRE_RELEASE_REPO_IMAGE and only use it in the tests
2- document that testers should use `tz config set env.TANZU_CLI_ADDITIONAL_PLUGIN_DISCOVERY_IMAGES_TEST_ONLY` to "override" the default repo

marckhouzam (approved):
LGTM pending the linter fix

Makefile
anujc25: This is because CI was failing constantly with this error.
```
Unable to find image 'mirror.gcr.io/library/registry:2' locally
docker: Error response from daemon: manifest for mirror.gcr.io/library/registry:2 not found: manifest unknown: Failed to fetch "2" from request "/v2/library/registry/manifests/2".
See 'docker run --help'.
```

anujc25: If needed I will be fixing this in a separate PR. But as CI was failing constantly, I wanted to test it with this PR.

vuil (approved):
lgtm

PR #139, opened 2023-03-30
Update plugin-tooling.mk, Update `tanzu builder init` to include `plugin-tooling.mk` to build and publish plugins

anujc25:
### What this PR does / why we need it

- Update `plugin-tooling.mk` within the repo.
- Add a `make plugin-install-local` target to install locally built plugins.
- Update `builder` plugin templates to use the new `plugin-tooling.mk` file to build the plugins.

### Which issue(s) this PR fixes

Fixes #86 

### Describe testing done for PR

```
# Initialize a new repository
$ tanzu builder init hello-world --repo-type github
$ cd hello-world

# Add a new plugin with name foo
$ tanzu builder cli add-plugin foo

# Commit the base changes
$ git add -A
$ git commit -m "Initialize plugin repository"

# Go get the latest tanzu-plugin-runtime from the main branch. [NOTE: This step should not be needed once we have an official tag on the `tanzu-plugin-runtime` repository]
$ go get github.com/vmware-tanzu/tanzu-plugin-runtime@main
$ make gomod

# Building Plugin
$ make plugin-build-local

# Installing plugins from a local source
$ make plugin-install-local

# OR Combined Build and Install target is also available
$ make plugin-build-install-local

# Publish the plugin as oci image to the local oci repository
$ make plugin-publish-packages

# Combined target to build and publish plugins
$ make plugin-build-and-publish-packages

# Initialize a local inventory database by running the below command
$ make inventory-plugin-add

# Add plugin to the local inventory database
$ make inventory-plugin-add

# Configure below environment variable to point to the inventory database as discovery image 
$ tanzu config set env.TANZU_CLI_ADDITIONAL_PLUGIN_DISCOVERY_IMAGES_TEST_ONLY localhost:5001/test/v1/tanzu-cli/plugins/plugin-inventory:latest

# Run plugin search
$ tz plugin search
  NAME                DESCRIPTION                                                        TARGET           LATEST
  foo                 foo plugin v2                                                      global           v0.0.0
```

### Release note

```release-note
Update `tanzu builder init` to include `plugin-tooling.mk` file containing the tooling to build and publish plugins
```

PR #153, opened 2023-04-03
Hide the --local flag of the "plugin install" cmd

### LINKED ISSUE #152: The `--local` flag should not be used in production to install plugins
**Bug description**

With the central repository feature plugins should be installed from the Central Repository.
However, when developing a new plugin version, a developer will want to easily test the plugin without publishing it yet.
This can be done using `tanzu plugin install --local <plugin location>`.  Furthermore, with #139, it will be possible to install a plugin locally by doing `make plugin-install-local`.  Therefore a developer will not need to explicitly use the `--local` flag by using the make target.

We should hide the `--local` flag so that it is not interpreted as being usable in production. 

**Expected behavior**

`plugin install --local` should work but the `--local` flag should be hidden from the help and shell completion.

```
$ tanzu plugin install -h | grep local
  -l, --local string     path to local discovery/distribution source
```

**Output of `tanzu version`**

version: v0.1.0-dev
buildDate: 2023-04-02
sha: 2e0c00aa8

Locally

marckhouzam:

Hide the --local flag of the "plugin install" cmd to avoid giving the impression that it can be used in production.
The `plugin install --local` form should only be used when doing local testing while developing a plugin and it should be used implicitly through `make plugin-install-local` that is introduced in #139 .

Fixes #152 

```
# Notice there is no --local flag shown
$ tz plugin install -h

# Notice there is no shell completion of the --local flag
$ tz plugin install builder --loc<TAB>

# Notice the --local flag still works even though it is hidden
$ tz plugin install builder --local ./artifacts/plugins/darwin/amd64/
[i] Installing plugin 'builder:v0.1.0-dev-23-g23f63f168' with target 'global'
[ok] successfully installed 'builder' plugin
```

### Release note

```release-note
The "tanzu plugin install --local" form is no longer to be used in production.  The "--local" flag is therefore hidden.
```


anujc25 (approved):
LGTM

PR #178, opened 2023-04-04
Fix Docs in Target and Context Management section under `docs/full/README.md`

anujc25:

- Fix Docs in Target and Context Management section under `docs/full/README.md`

Fixes #

### Release note

```release-note

```

### DIFF
--- a/docs/full/README.md
+++ b/docs/full/README.md
@@ -51,7 +51,7 @@ contexts for the same combination of `(user, server)`.
 
 ## Target
 
-A Target refers to a category or tier of control planes that the CLI can interact with. There are currently two supported targets : kubernetes (or k8s) and mission-control (or tmc) which corresponds to the Kubernetes cluster endpoint type and Tanzu Mission Control endpoint type. A context is associated with one of the two supported targets. Plugins are not necessarily but can often be associated with one of the targets as well. To see plugins that apply to a particular Target and not the other, run the command `tanzu <target>'.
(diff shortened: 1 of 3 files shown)


docs/full/README.md
```diff
@@ -51,7 +51,7 @@ contexts for the same combination of `(user, server)`.
...
+A Target refers to a category or tier of control planes that the CLI can interact with. There are currently two supported targets : `kubernetes` (or k8s) and `mission-control` (or tmc) which corresponds to the Kubernetes cluster endpoint type and Tanzu Mission Control endpoint type respectively. A context is associated with one of the two supported targets. Plugins are generally associated with one of the above mentioned targets but if a plugin doesn't fall into any of the above categories developer can create a plugin with 'global' target. A plugin using a `global` target is available as a root Tanzu CLI command. To see plugins that apply to a particular Target and not the other, run the command `tanzu <target>'.
```
marckhouzam, line 54:
```suggestion
A Target refers to a category or tier of control planes that the CLI can interact with. There are currently two supported targets : `kubernetes` (or `k8s`) and `mission-control` (or `tmc`) which corresponds to the Kubernetes cluster endpoint type and Tanzu Mission Control endpoint type respectively. A context is associated with one of the two supported targets. Plugins are generally associated with one of the above mentioned targets but if a plugin doesn't fall into any of the above categories a developer can create a plugin with the 'global' target. A plugin using the `global` target is available as a root Tanzu CLI command. To see plugins that only apply the `kubernetes` Target or only to the `mission-control` Target, run the command `tanzu <target>'.
```

marckhouzam (approved):
Thanks for this!
One small suggested tweek.
LGTM

PR #188, opened 2023-04-06
Disable CEIP prompt for tanzu 'version', 'ceip set' and 'config set' commands

### LINKED ISSUE #189: Disable CEIP participation prompt for tanzu 'verison', 'ceip set' and 'config set' commands
**Bug description**
It's not a good UX to prompt users for intial basic command 'tanzu version' and also if user trying to set the configuration for CEIP using "ceip set" and also for "config set"

**Expected behavior**
Disable CEIP participation prompt for the above said commands

**Output of `tanzu version`**

prkalle:
- Disable CEIP prompt for `tanzu version`, 'tanzu ceip set' and 'tanzu config set'
- Also updated the short/long description for CEIP command

This PR would disable the CEIP prompt for "tanzu version", "tanzu config set" and "tanzu ceip set" 

We need it for the below reasons:

 - 'tanzu version'(common first command to run),
 - 'ceip set' (not a good UX to prompt user who is trying to set the CEIP configuration)
 - 'config set' (would be a chicken and egg issue if user tries to set CEIP configuration using "tanzu config set env.TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER yes")
 
Fixes #189  

Tested 
1. `tanzu version command`
2. `tanzu config set`
3. `tanzu ceip set`
All the above commands doesn't show the CEIP prompt 
```
❯ ./bin/tanzu version
Warning, Masking commands for plugins "login" because a core command or other plugin with that name already exists.
version: v0.0.2-dev
buildDate: 2023-04-06
sha: 83540128-dirty

❯ ./bin/tanzu config set env.TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER yes
Warning, Masking commands for plugins "login" because a core command or other plugin with that name already exists.
❯ vim ~/.config/tanzu/config.yaml

❯ vim ~/.config/tanzu/config-ng.yaml

❯ ./bin/tanzu ceip set false
Warning, Masking commands for plugins "login" because a core command or other plugin with that name already exists.
❯ ./bin/tanzu ceip get
Warning, Masking commands for plugins "login" because a core command or other plugin with that name already exists.
  CEIP-STATUS
  Opt-out
```
Then, removed the CEIP opt-in config from the configuration and tested `tanzu config init` and could see the ceip prompt showing up
```
❯ ./bin/tanzu config init
Warning, Masking commands for plugins "login" because a core command or other plugin with that name already exists.
?
VMware's Customer Experience Improvement Program ("CEIP") provides VMware with information that enables VMware to improve its products and services and fix problems. By choosing to participate in CEIP, you agree that VMware may collect technical information about your use of VMware products and services on a regular basis. This information does not personally identify you.
For more details about the Program, please see http://www.vmware.com/trustvmware/ceip.html

Do you agree to Participate in the Customer Experience Improvement Program?
  [Use arrows to move, type to filter]
> Yes
  No
```

Tested the `tanzu ceip-participation` command help to show the details description for the command

```
❯ ./bin/tanzu ceip-participation
Warning, Masking commands for plugins "login" because a core command or other plugin with that name already exists.
Manage VMware's Customer Experience Improvement Program (CEIP) participation which provides VMware with information that enables VMware to improve its products and services and fix problems

Usage:
  tanzu ceip-participation [command]

Aliases:
  ceip-participation, ceip

Available Commands:
  get         Get the current CEIP opt-in status
  set         Set the opt-in preference for CEIP

Flags:
  -h, --help   help for ceip-participation

Use "tanzu ceip-participation [command] --help" for more information about a command.
```

### Release note

```release-note
Disabled the CEIP participation prompt for tanzu 'version', 'ceip set' and 'config set' commands
```

### DIFF
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -22,29 +22,11 @@ import (
 	"github.com/vmware-tanzu/tanzu-cli/pkg/pluginsupplier"
 )
 
+const cmdNameSet = "set"
+
 // NewRootCmd creates a root command.
 func NewRootCmd() (*cobra.Command, error) {
-	var rootCmd = &cobra.Command{
-		Use: "tanzu",
-		// Don't have Cobra print the error message, the CLI will
-		// print it itself in a nicer format.
-		SilenceErrors: true,
-		// silencing usage for now as we are getting double usage from plugins on errors
-		SilenceUsage: true,
-		// Flag parsing must be deactivated because the root plugin won't know about all flags.
-		DisableFlagParsing: true,
-		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			if cmd.Name() != cobra.ShellCompRequestCmd && cmd.Name() != completionCmd.Name() {
-				// Configure CEIP setting but not if we are doing shell completion stuff.
-				// The shell completion setup is not interactive, so it should not trigger
-				// the ceip prompt.
-				if err := cliconfig.ConfigureCEIPOptIn(); err != nil {
-					return err
-				}
-			}
-			return nil
-		},
-	}
+	var rootCmd = newRootCmd()
 	uFunc := cli.NewMainUsage().UsageFunc()
 	rootCmd.SetUsageFunc(uFunc)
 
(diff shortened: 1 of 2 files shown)


pkg/command/root.go
```diff
@@ -121,6 +101,34 @@ func NewRootCmd() (*cobra.Command, error) {
...
+		// print it itself in a nicer format.
+		SilenceErrors: true,
+		// silencing usage for now as we are getting double usage from plugins on errors
+		SilenceUsage: true,
+		// Flag parsing must be deactivated because the root plugin won't know about all flags.
+		DisableFlagParsing: true,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if cmd.Name() != cobra.ShellCompRequestCmd && cmd.Name() != completionCmd.Name() &&
+				cmd.Name() != newVersionCmd().Name() &&
+				!(cmd.Name() == "set" && (cmd.Parent().Name() == configCmd.Name() || cmd.Parent().Name() == newCEIPParticipationCmd().Name())) {
```
anujc25, line 117: nit: It might be more readable if we can put this condition out in its own function.

anujc25 (approved):
Just a minor nit. We can address it as a followup as well. Otherwise LGTM.

vuil:
Nit: If we don't add as code comments, I suggest we put in the "What this PR does / why we need it"  PR description on the reason we elected to special-case these commands so we don't forget.

pkg/command/root.go
prkalle: Updated

prkalle:
> Nit: If we don't add as code comments, I suggest we put in the "What this PR does / why we need it" PR description on the reason(s) we elected to special-case these commands so we don't forget. (this is a "why we need it" part :-) )

Add the reasoning in the code comments and PR description

PR #190, opened 2023-04-06
Fix the error message for `tanzu plugin install` cmd when all plugins from the group are not mandatory

mpanchajanya:

Better UX for `tanzu plugin install ` command when all plugins from plugin group are not mandatory.

Fixes #

Test test cases from manager_test.go and plugin_test.go - Passed

make test  - Passed

Test Case :  Install account plugin in the group; account is not mandatory in the group and the group has no mandatory plugins

Previously: 
```
tanzu plugin install account --group  vmware-tmc/default:v0.0.1
[x] : plugin 'account' is not part of the group 'vmware-tmc/default:v0.0.1'
```

After the fix:
```
tanzu plugin install account --group  vmware-tmc/default:v0.0.1
[x] : plugin 'account' from group 'vmware-tmc/default:v0.0.1' is not mandatory to install
```

Test Case : Try to install all plugins in the group; plugin group has no mandatory plugins

Previously: 
```
tanzu plugin install all --group  vmware-tmc/default:v0.0.1
[x] : plugin 'account' is not part of the group 'vmware-tmc/default:v0.0.1'
```

After the fix:
```
tanzu plugin install all --group  vmware-tmc/default:v0.0.1
[x] : plugin group 'vmware-tmc/default:v0.0.1' has no mandatory plugins to install
```

### Release note

```release-note

```

### DIFF
--- a/pkg/pluginmanager/manager.go
+++ b/pkg/pluginmanager/manager.go
@@ -825,24 +825,43 @@ func InstallPluginsFromGroup(pluginName, groupID string) error {
 
 	numErrors := 0
 	numInstalled := 0
+	mandatoryPluginsExist := false
+	pluginExist := false
 	for _, plugin := range group.Plugins {
-		if (pluginName == cli.AllPlugins || pluginName == plugin.Name) && plugin.Mandatory {
-			err = InstallStandalonePlugin(plugin.Name, plugin.Version, plugin.Target)
-			if err != nil {
-				numErrors++
-				log.Warningf("unable to install plugin '%s': %v", plugin.Name, err.Error())
-			} else {
-				numInstalled++
+		if pluginName == cli.AllPlugins || pluginName == plugin.Name {
+			pluginExist = true
+			if plugin.Mandatory {
+				mandatoryPluginsExist = true
+				err := InstallStandalonePlugin(plugin.Name, plugin.Version, plugin.Target)
+				if err != nil {
+					numErrors++
+					fmt.Printf("unable to install plugin '%s': %v", plugin.Name, err.Error())
+				} else {
+					numInstalled++
+				}
 			}
 		}
+	}
+
+	if !pluginExist {
+		return fmt.Errorf("plugin '%s' is not part of the group '%s'", pluginName, groupID)
+	}
+
+	if !mandatoryPluginsExist {
+		if pluginName == cli.AllPlugins {
+			return fmt.Errorf("plugin group '%s' has no mandatory plugins to install", groupID)
+		}
+		return fmt.Errorf("plugin '%s' from group '%s' is not mandatory to install", pluginName, groupID)
 	}
 
 	if numErrors > 0 {
 		return fmt.Errorf("could not install %d plugin(s) from group '%s'", numErrors, groupID)
 	}
+
 	if numInstalled == 0 {
 		return fmt.Errorf("plugin '%s' is not part of the group '%s'", pluginName, groupID)
 	}
+
 	return nil
 }
 


vuil:
@mpanchajanya the second "After the fix:" output doesn't seem right. probably just a copy/paste error

mpanchajanya:
> @mpanchajanya the second "After the fix:" output doesn't seem right. probably just a copy/paste error

Updated the output

marckhouzam:
I find that the situation we are dealing with here is tricky.  The below content is not about the PR change but the behaviour when dealing with non-mandatory plugins.

We allow users to install any plugin manually (as a standalone plugin).  So they can do
```
tanzu plugin install account
```

This installs the latest version of the `account` plugin.  If the user wants to install the version of this plugin that is required by TMC v0.0.1, it would seem to make sense to do:
```
tanzu plugin install account --group  vmware-tmc/default:v0.0.1
```
In that optic, why wouldn't the plugin get installed?  I think it should.

anujc25:
> > I find that the situation we are dealing with here is tricky. The below content is not about the PR change but the behaviour when dealing with non-mandatory plugins.
> > We allow users to install any plugin manually (as a standalone plugin). So they can do
> > ```
> > tanzu plugin install account
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This installs the latest version of the `account` plugin. If the user wants to install the version of this plugin that is required by TMC v0.0.1, it would seem to make sense to do:
> > ```
> > tanzu plugin install account --group  vmware-tmc/default:v0.0.1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In that optic, why wouldn't the plugin get installed? I think it should.
> 
> Yea That makes sense to install individual plugin . As per current and previous implementation we do not allow installing a non mandatory plugin. Should we file a enhancement for this ? cc @anujc25 @marckhouzam

I would prefer we file a separate enhancement issue regarding this. I would like to understand the UX for `tanzu plugin group get` to understand more about what will we show as part of the output of that command and if needed we can align this implementation accordingly.

marckhouzam:
> I would prefer we file a separate issue regarding this. I would like to understand the UX for tanzu plugin group get to understand more about what will we show as part of the output of that command and if needed we can align this implementation accordingly. @marckhouzam WDTY?

Sure, let's wait until we get further with `tanzu plugin group get` 👍

anujc25 (approved):

pkg/pluginmanager/manager.go
```diff
@@ -825,24 +825,43 @@ func InstallPluginsFromGroup(pluginName, groupID string) error {
...
-			} else {
-				numInstalled++
+		if pluginName == cli.AllPlugins || pluginName == plugin.Name {
+			pluginExist = true
+			if plugin.Mandatory {
+				mandatoryPluginsExist = true
+				err := InstallStandalonePlugin(plugin.Name, plugin.Version, plugin.Target)
+				if err != nil {
+					numErrors++
+					fmt.Printf("unable to install plugin '%s': %v", plugin.Name, err.Error())
```
marckhouzam, line 838: @mpanchajanya Any reason you changed this from `log.Warningf` to `fmt.Printf`?

The former prints to stderr while the latter prints to stdout so when getting an error, the output is not on the same stream.

Also, `log.Warningf` seems to automatically add a newline at the end, which is important in this case.

I can make a quick PR to fix it but I wanted to know if there was a reason behind it before messing it up.

marckhouzam: Fixed in #226

PR #192, opened 2023-04-06
Do not show command masked warning for `login` plugin

anujc25:
### What this PR does / why we need it

- Do not show command masked warning for the `login` plugin
- As the `login` plugin is now part of the core Tanzu CLI command and not a plugin anymore, skip the `login` plugin from adding it to the maskedPlugins array to avoid the warning message from getting shown to the user on each command invocation.

### Which issue(s) this PR fixes

Fixes #191

### Describe testing done for PR

- Install plugins with old CLI
```
/t/4 $ tanzu-0.28.0 plugin sync
ℹ  Checking for required plugins...
ℹ  Installing plugin 'management-cluster:v0.28.0' with target 'kubernetes'
ℹ  Installing plugin 'package:v0.28.0' with target 'kubernetes'
ℹ  Installing plugin 'pinniped-auth:v0.28.0'

ℹ  Installing plugin 'secret:v0.28.0' with target 'kubernetes'
ℹ  Installing plugin 'telemetry:v0.28.0' with target 'kubernetes'
ℹ  Installing plugin 'isolated-cluster:v0.28.0'
ℹ  Installing plugin 'login:v0.28.0'
ℹ  Successfully installed all required plugins
✔  Done
```

- Run a command with a new CLI WITHOUT bug fix
```
$ tz version
Warning, Masking commands for plugins "login" because a core command or other plugin with that name already exists.
version: v0.81.1
buildDate: 2023-04-06
sha: 73a4c10d-dirty
```

- Run a command with a new CLI WITH bug fix
```
$ tz version
version: v0.81.1
buildDate: 2023-04-06
sha: d3580d4e
```

### Release note

```release-note

```

PR #194, opened 2023-04-07
Update quickstart for login plugin

marckhouzam:

Update the doc after the fix of #192 .

Fixes #

### Release note

```release-note

```


vuil (approved):
lgtm, good call, thanks for the update.

PR #195, opened 2023-04-07
Use x86_64 as the correct RPM arch name

marckhouzam:

The RPM package script was using the wrong arch name for AMD64.  This prevented installing the CLI with YUM on AMD64 machines.

Fixes # N/A

The testing was done by a team member on his AMD64 Mac using docker.
The installation used to fail for him but now it works.

### Release note

```release-note

```


vuil (approved):
lgtm, thanks!

PR #198, opened 2023-04-07
Update documentation regarding plugin-group and update plugin-tooling to use appropriate variable name

anujc25:

- Update documentation to mention users will see `<plugin-group-name>:<plugin-group-version>` separately in the command outputs to align with the proposal.
- Update the `plugin-tooling.mk` file to specify that the variable the user is providing should look like `<plugin-group-name>:<plugin-group-version>` Also update the variable name from `PLUGIN_GROUP_NAME` to `PLUGIN_GROUP_NAME_VERSION`

Fixes #

### Release note

```release-note
NONE
```


vuil (approved):
lgtm, thanks

PR #209, opened 2023-04-11
Fix e2e random failures and add stress tests to context use cases

chandrareddyp:
This PR fixes the E2E random failures issue and adds stress tests to context life cycle use cases.
Before this PR, we execute simultaneously more than one suite of E2E tests which was causing issue to fail E2E tests, because they are trying to change the same config file.
In this PR we have updated the make file target `e2e-cli-core` to run only one E2E test suite at a time.

Fixes #

[Tanzu CLI Core E2E Tests](https://github.com/vmware-tanzu/tanzu-cli/actions/runs/4693107509/jobs/8319654928?pr=209) executed multiple times in this PR

### Release note

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 include ./plugin-tooling.mk
+include ./test/e2e/Makefile
 
 # Ensure Make is run with bash shell as some syntax below is bash-specific
 SHELL := /usr/bin/env bash
@@ -64,22 +65,6 @@ LD_FLAGS += -X 'github.com/vmware-tanzu/tanzu-cli/pkg/buildinfo.Version=$(BUILD_
 ENVS ?= linux-amd64 windows-amd64 darwin-amd64
 
 CLI_TARGETS := $(addprefix build-cli-,${ENVS})
- 
-ifndef TANZU_API_TOKEN
-TANZU_API_TOKEN = ""
-endif
-
-ifndef TANZU_CLI_TMC_UNSTABLE_URL
-TANZU_CLI_TMC_UNSTABLE_URL = ""
-endif
-
-ifndef TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL
-TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL = gcr.io/eminent-nation-87317/tanzu-cli/test/v1/plugins/plugin-inventory:latest
-endif
-
-ifndef TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL
-TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL = localhost:9876/tanzu-cli/plugins/central:small
-endif
 
 ## --------------------------------------
 ## Help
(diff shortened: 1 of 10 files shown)


test/e2e/e2e-test.mk
```diff
@@ -0,0 +1,75 @@
...
+ifndef TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL
+TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL = localhost:9876/tanzu-cli/plugins/central:small
+endif
+
+
+.PHONY: e2e-cli-core ## Execute all CLI Core E2E Tests
+e2e-cli-core: e2e-cli-lifecycle e2e-cli-config e2e-cli-plugin-compatibility-test e2e-context-k8s-tests e2e-context-tmc-test e2e-cli-plugin-lifecycle-sync-test
+
+
+.PHONY: e2e-cli-lifecycle ## Executest CLI life cycle specific e2e tests
```
anujc25, line 29:
```suggestion
.PHONY: e2e-cli-lifecycle ## Execute CLI life cycle specific e2e tests
```

test/e2e/e2e-test.mk
```diff
@@ -0,0 +1,75 @@
...
+e2e-cli-lifecycle:
+	export TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER="Yes" ; \
+	${GO} test ./test/e2e/cli_lifecycle -timeout ${E2E_TEST_TIMEOUT} -race -coverprofile ${E2E_TEST_OUTPUT} ${GOTEST_VERBOSE} ; \
+
+.PHONY: e2e-cli-config ## Executest CLI config life cycle specific e2e tests
```
anujc25, line 34:
```suggestion
.PHONY: e2e-cli-config ## Execute CLI config life cycle specific e2e tests
```

test/e2e/e2e-test.mk
```diff
@@ -0,0 +1,75 @@
...
+	else \
+		export TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL) ; \
+		export TANZU_CLI_PRE_RELEASE_REPO_IMAGE=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL) ; \
+		export TANZU_CLI_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_VERIFICATION_SKIP_LIST=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL) ; \
+		export TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER="Yes" ; \
+		${GO} test ./test/e2e/plugin_lifecycle -timeout ${E2E_TEST_TIMEOUT} -race -coverprofile ${E2E_TEST_OUTPUT} ${GOTEST_VERBOSE} ; \
+		${GO} test ./test/e2e/plugin_sync -timeout ${E2E_TEST_TIMEOUT} -race -coverprofile ${E2E_TEST_OUTPUT} ${GOTEST_VERBOSE} ; \
+	fi
+
+.PHONY: e2e-context-k8s-tests ## Executest CLI context life cycle e2e tests for k8s target
```
anujc25, line 63:
```suggestion
.PHONY: e2e-context-k8s-tests ## Execute CLI context life cycle e2e tests for k8s target
```

test/e2e/e2e-test.mk
```diff
@@ -0,0 +1,75 @@
+# Copyright 2022 VMware, Inc. All Rights Reserved.
```
anujc25, line 1:
```suggestion
# Copyright 2023 VMware, Inc. All Rights Reserved.
```

test/e2e/e2e-test.mk
anujc25, line 1: We should rename this file as `Makefile` as it is already part of the `test/e2e` directory. So, As a user, I can run specific e2e tests from this directory as well by running `make e2e-cli-config` if needed.

test/e2e/e2e-test.mk
```diff
@@ -0,0 +1,75 @@
...
+TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL = gcr.io/eminent-nation-87317/tanzu-cli/test/v1/plugins/plugin-inventory:latest
+endif
+
+ifndef TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL
+TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL = localhost:9876/tanzu-cli/plugins/central:small
+endif
+
+
+.PHONY: e2e-cli-core ## Execute all CLI Core E2E Tests
+e2e-cli-core: e2e-cli-lifecycle e2e-cli-config e2e-cli-plugin-compatibility-test e2e-context-k8s-tests e2e-context-tmc-test e2e-cli-plugin-lifecycle-sync-test
```
anujc25, line 26: For only this external facing test target, the help text should be printed with the target so it is shown with `make help` output.

```suggestion
e2e-cli-core: e2e-cli-lifecycle e2e-cli-config e2e-cli-plugin-compatibility-test e2e-context-k8s-tests e2e-context-tmc-test e2e-cli-plugin-lifecycle-sync-test ## Execute all CLI Core E2E Tests
```

test/e2e/e2e-test.mk
```diff
@@ -0,0 +1,75 @@
...
+		${GO} test ./test/e2e/plugin_lifecycle -timeout ${E2E_TEST_TIMEOUT} -race -coverprofile ${E2E_TEST_OUTPUT} ${GOTEST_VERBOSE} ; \
+		${GO} test ./test/e2e/plugin_sync -timeout ${E2E_TEST_TIMEOUT} -race -coverprofile ${E2E_TEST_OUTPUT} ${GOTEST_VERBOSE} ; \
+	fi
+
+.PHONY: e2e-context-k8s-tests ## Executest CLI context life cycle e2e tests for k8s target
+e2e-context-k8s-tests:
+	export TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER="Yes" ; \
+	${GO} test `go list ./test/e2e/context/... | grep -v test/e2e/context/tmc` -timeout ${E2E_TEST_TIMEOUT} -race -coverprofile ${E2E_TEST_OUTPUT} ${GOTEST_VERBOSE}
+
+.PHONY: e2e-context-tmc-test ## Executest CLI context life cycle e2e tests for tmc target
```
anujc25, line 68:
```suggestion
.PHONY: e2e-context-tmc-test ## Execute CLI context life cycle e2e tests for tmc target
```

Makefile
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 include ./plugin-tooling.mk
+include ./test/e2e/e2e-test.mk
```
anujc25, line 5: Thanks for doing this and cleaning the root level makefile. 👍

mpanchajanya (commented):
@chandrareddyp I wonder if confining ourselves to just one test suite might result in slower test execution and increased build time, particularly as we keep adding more e2e tests that would be another problem we might have to tackle. I'm of the opinion that tests should pass regardless of whether they are run concurrently or in a random order.

Also small nit on documentation. I couldn't find relevant documentation on How do we run these individual tests/ test suites locally. Does it need any additional setup ?

chandrareddyp:
> @chandrareddyp I wonder if confining ourselves to just one test suite might result in slower test execution and increased build time, particularly as we keep adding more e2e tests that would be another problem we might have to tackle. I'm of the opinion that tests should pass regardless of whether they are run concurrently or in a random order.

@mpanchajanya we can not run concurrently these e2e tests because most of tests are modifying the config files which causes issues which is why we need to run one suite at a time.

> Also small nit on documentation. I couldn't find relevant documentation on How do we run these individual tests/ test suites locally. Does it need any additional setup ?

We have another story to make e2e tests framework as an independent module and update the docs, as part of that I will update.

chandrareddyp:
> > @chandrareddyp I wonder if confining ourselves to just one test suite might result in slower test execution and increased build time, particularly as we keep adding more e2e tests that would be another problem we might have to tackle. I'm of the opinion that tests should pass regardless of whether they are run concurrently or in a random order.
> 
> @mpanchajanya we can not run concurrently these e2e tests because most of tests are modifying the config files which causes issues which is why we need to run one suite at a time.
> 
> > Also small nit on documentation. I couldn't find relevant documentation on How do we run these individual tests/ test suites locally. Does it need any additional setup ?
> 
> We have another story to make e2e tests framework as an independent module and update the docs, as part of that I will update.

@mpanchajanya we have created a new make target `make e2e-cli-core` to start the local test repo which is needed to run e2e tests, and for TMC use case i have updated the comment in the make file, otherwise, we need nothing, user can run directly the e2e tests.

anujc25 (approved):
LGTM

PR #218, opened 2023-04-14
Ignore hidden/deactivated plugins

marckhouzam:

This PR teaches the CLI to ignore hidden/deactivated plugins.

The PR adds an `IncludeHidden` boolean field to the plugin filter struct and updates the SQLite query to respect that field.

Furthermore, the pluginfilter argument to `GetPlugins()` when nil, is replaced with an empty filter. This allows the empty value of such a filter to ignore hidden plugins automatically since the field `IncludeHidden` will be false by default.

Please see the "Special notes" section below for different design options.

Also, our test local repository wasn't properly generated to handle hidden plugins, so the second commit of this PR fixes this.  The generating script was using `FALSE` as a boolean when it should have used `false` (the difference in case matters since SQLite doesn't use a boolean type but uses a string type instead).

Fixes # N/A

First publish some plugins to the local test DB:
```
$ make plugin-build-and-publish-packages
$ make inventory-init
$ make inventory-plugin-add

# Configure the CLI to use the local test repository
$ tz config set env.TANZU_CLI_PRE_RELEASE_REPO_IMAGE localhost:5001/test/v1/tanzu-cli/plugins/plugin-inventory:latest
$ tz config set env.TANZU_CLI_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_VERIFICATION_SKIP_LIST localhost:5001/test/v1/tanzu-cli/plugins/plugin-inventory:latest

# Make sure we can see the plugins
$ tz plugin search
[i] Reading plugin inventory for "localhost:5001/test/v1/tanzu-cli/plugins/plugin-inventory:latest", this will take a few seconds.
[!] Skipping the plugins discovery image signature verification for "localhost:5001/test/v1/tanzu-cli/plugins/plugin-inventory:latest"

  NAME     DESCRIPTION             TARGET  LATEST
  builder  Build Tanzu components  global  v0.90.0-alpha.0
  test     Test the CLI            global  v0.90.0-alpha.0
```

Now we deactivate the plugins and check that they are not shown anymore:
```
$ make inventory-plugin-deactivate
$ tz plugin search
[i] Reading plugin inventory for "localhost:5001/test/v1/tanzu-cli/plugins/plugin-inventory:latest", this will take a few seconds.
[!] Skipping the plugins discovery image signature verification for "localhost:5001/test/v1/tanzu-cli/plugins/plugin-inventory:latest"

  NAME  DESCRIPTION  TARGET  LATEST

$ tz plugin install builder
[x] : unable to find plugin 'builder'
```

Reactivate the plugins and see that they can be used again:
```
$ make inventory-plugin-activate

$ tz plugin search
[i] Reading plugin inventory for "localhost:5001/test/v1/tanzu-cli/plugins/plugin-inventory:latest", this will take a few seconds.
[!] Skipping the plugins discovery image signature verification for "localhost:5001/test/v1/tanzu-cli/plugins/plugin-inventory:latest"

  NAME     DESCRIPTION             TARGET  LATEST
  builder  Build Tanzu components  global  v0.90.0-alpha.0
  test     Test the CLI            global  v0.90.0-alpha.0

$ tz plugin install builder
[i] Installing plugin 'builder:v0.90.0-alpha.0' with target 'global'
[ok] successfully installed 'builder' plugin
```

### Release note

```release-note
Plugin that are marked as "deactivated" in the repository of plugins are now ignored by the CLI.
```

The below points are certain decisions I made that I hesitated about:

1. It would have been much easier to simply hard-code in `sqlite_inventory.go` to not show hidden plugins.  And then later one, read an env var directly in `sqlite_inventory.go` to allow showing hidden plugins.  However, I felt it was not the best design. I chose to continue using SQLite queries to properly choose what was extracted from the DB.  By doing that, I made the decision to keep the `plugin_inventory` abstraction separate from any special configuration about showing hidden plugins or not; instead the plugin inventory simply follows the instructions of the filter.

2. To keep the code changes as small as possible, I kept most existing interfaces the same.  However, I actually came very close to:
     1. remove `PluginInventory.GetAllPlugins()` to force every call to get plugins to use a proper filter with `PluginInventory.GetPlugins(filter)`.  Instead, I redirected `PluginInventory.GetAllPlugins()` to call `PluginInventory.GetPlugins(filter)` but only to avoid more code changes
     2. make `GetPlugins(*PluginInventoryFilter)` not take a pointer but an filter object directly like so `GetPlugins(PluginInventoryFilter)`.  This was to avoid using a nil filter which would not automatically ignore hidden plugins.  Instead I check if the filter is nil inside `GetPlugins(filter)` and replace it with an empty filter.
     3. Note that the above two points make things different between the interface for `PluginGroups` where I did do those two points.  

Should we reconcile the two interfaces?


anujc25:
> This commit adds a "ShowHidden" boolean field to the plugin filter
struct and updates the SQLite query to respect that field.

Update `ShowHidden` to `IncludeHidden` in the PR description and the commit message.

marckhouzam:
Alright, this is ready to review and hopefully merge.
The original E2E test failures showed that our test local repository wasn't properly generated to handle hidden plugins, so the second commit of this PR fixes this.

PR #226, opened 2023-04-20
Use log.Warningf() insead of fmt.Printf()

marckhouzam:

`log.Warningf` prints to stderr while `fmt.Printf` stdout. For the case in question, we are processing an error but the output will be on stdout for one part and stderr for the other.

Also, `log.Warningf` automatically adds a newline at the end while `fmt.Printf` does not.  This is an issue in this case as we will have multiple lines of output.

Fixes # N/A

I have a test repo with hidden plugins that are part of the plugin group.  This causes errors at installation time so we can see the printouts.

Before the PR:
```
$ tz plugin install --group vmware-tzcli/test:v0.0.1
unable to install plugin 'builder': unable to find plugin 'builder' for target 'global'unable to install plugin 'test': unable to find plugin 'test' for target 'global'[x] : could not install 2 plugin(s) from group 'vmware-tzcli/test:v0.0.1'

$ tz plugin install --group vmware-tzcli/test:v0.0.1>/dev/null
[x] : could not install 2 plugin(s) from group 'vmware-tzcli/test:v0.0.1'
```
With this PR:
```
$ tz plugin install --group vmware-tzcli/test:v0.0.1
[!] unable to install plugin 'builder': unable to find plugin 'builder' for target 'global'
[!] unable to install plugin 'test': unable to find plugin 'test' for target 'global'
[x] : could not install 2 plugin(s) from group 'vmware-tzcli/test:v0.0.1'
$ tz plugin install --group vmware-tzcli/test:v0.0.1>/dev/null
[!] unable to install plugin 'builder': unable to find plugin 'builder' for target 'global'
[!] unable to install plugin 'test': unable to find plugin 'test' for target 'global'
[x] : could not install 2 plugin(s) from group 'vmware-tzcli/test:v0.0.1'
```

### Release note

```release-note

```

### DIFF
--- a/pkg/pluginmanager/manager.go
+++ b/pkg/pluginmanager/manager.go
@@ -835,7 +835,7 @@ func InstallPluginsFromGroup(pluginName, groupID string) error {
 				err := InstallStandalonePlugin(plugin.Name, plugin.Version, plugin.Target)
 				if err != nil {
 					numErrors++
-					fmt.Printf("unable to install plugin '%s': %v", plugin.Name, err.Error())
+					log.Warningf("unable to install plugin '%s': %v", plugin.Name, err.Error())
 				} else {
 					numInstalled++
 				}


mpanchajanya (approved):
lgtm Thanks for fixing this

PR #232, opened 2023-04-25
Use a plugin source to find the central plugin discovery

marckhouzam:

This PR removes the use of `TANZU_CLI_PRE_RELEASE_REPO_IMAGE` which was a temporary approach.

Instead, the central discovery is automatically created in the config file as a "discoverySource" and "tanzu plugin source update" can be used to override this default value.

This commit also updates the "tanzu plugin source" family of commands.
With the central discovery feature enabled:
1. "tanzu plugin source add" is removed
2. "tanzu plugin source update" no longer has a "--type/-t" flag
3. "tanzu plugin source init" is added

Please see the additional notes at the end of this PR's description to see how "tanzu plugin source delete" is handled.

Fixes # N/A

I confirm that `make e2e-context-tmc-test` passes locally.

```
# Cleanup
$ rm ~/.cache/tanzu/plugin_inventory
$ rm ~/.config/tanzu/config*
$ unset TANZU_CLI_PRE_RELEASE_REPO_IMAGE
$ tz ceip-participation set true

# Notice the default central discovery is automatically set
$ tz plugin source list
  NAME     IMAGE
  default  projects.registry.vmware.com/tanzu_cli_stage/plugins/plugin-inventory:latest
$ tz plugin search
[i] Reading plugin inventory for "projects.registry.vmware.com/tanzu_cli_stage/plugins/plugin-inventory:latest", this will take a few seconds.
  NAME     DESCRIPTION             TARGET  LATEST
  builder  Build Tanzu components  global  v0.90.0-alpha.0
  test     Test the CLI            global  v0.90.0-alpha.0

# Update the default plugin source (meant for air-gapped)
$ tz plugin source list
  NAME     IMAGE
  default  localhost:9876/tanzu-cli/plugins/central:small
$ tz plugin search

  NAME                DESCRIPTION                  TARGET           LATEST
  isolated-cluster    Desc for isolated-cluster    global           v9.9.9
  pinniped-auth       Desc for pinniped-auth       global           v9.9.9
  cluster             Desc for cluster             kubernetes       v9.9.9
[...]

# Reset the default plugin source
$ tz plugin source init
[ok] successfully initialized discovery source
$ tz plugin source list
  NAME     IMAGE
  default  projects.registry.vmware.com/tanzu_cli_stage/plugins/plugin-inventory:latest

# Remove the config files and see that the default source is re-created automatically
$ rm ~/.config/tanzu/config*
$ tz ceip-participation set true
$ tz plugin source list
  NAME     IMAGE
  default  projects.registry.vmware.com/tanzu_cli_stage/plugins/plugin-inventory:latest
```

Now lets test the special "plugin source delete" handling which allows to remove the default central discovery:
```
$ tz plugin source delete default
[ok] deleted discovery source default

# Notice that the plugin source remains deleted (not added back automatically)
$ tz plugin source list
  NAME  IMAGE
$ tz plugin search
  NAME  DESCRIPTION  TARGET  LATEST

# Notice that we can easily restore the default central discovery
$ tz plugin source init
[ok] successfully initialized discovery source
$ tz plugin source list
  NAME     IMAGE
  default  projects.registry.vmware.com/tanzu_cli_stage/plugins/plugin-inventory:latest
```

### Release note

```release-note
The temporary `TANZU_CLI_PRE_RELEASE_REPO_IMAGE` has been removed.  Instead, the use of `tanzu plugin source update` can be used to override the default central discovery location.
```

There is an important behaviour I'd like to get opinions on:
I decided to keep `tanzu plugin source delete`.  The idea is that I've always been a bit uncomfortable with the fact there was no way to remove the default discovery.  By keeping the `tanzu plugin source delete` command it allows us to remove the default central repo and not have the CLI add it back automatically; instead, if a user does a `tanzu plugin source delete default` they can add back the central repo by doing `tanzu plugin source init`.

Note that on first use or removal of the config file, the CLI will notice the config file has no top-level entry for discoverySources and will create the default discovery source; but after a `tanzu plugin source delete`, the CLI will see the discoverySources section exist but is empty, in which case it will not add back the central discovery source.

If this is not desirable, we can easily remove the `tanzu plugin source delete` command as was originally planned.

### DIFF
--- a/pkg/command/discovery_source.go
+++ b/pkg/command/discovery_source.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/vmware-tanzu/tanzu-cli/pkg/cli"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/config"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/constants"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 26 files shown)


pkg/config/init.go
```diff
@@ -33,4 +28,15 @@ func init() {
 	if addedFeatureFlags || addedEdition || addedCompatabilityFile || addedBomRepo || addedContexts {
 		_ = config.StoreClientConfig(c)
 	}
+
+	// We need to release the config lock before call ing PopulateDefaultCentralDiscovery() because
```
anujc25, line 32:
```suggestion
	// We need to release the config lock before calling PopulateDefaultCentralDiscovery() because
```

test/e2e/Makefile
```diff
@@ -42,9 +42,10 @@ e2e-cli-plugin-compatibility-test:
 	@if [ "${TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL}" = "" ]; then \
 		echo "***Skipping Plugin Compatibility test cases because environment variables TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL is not set***" ; \
 	else \
-		export TANZU_CLI_PRE_RELEASE_REPO_IMAGE=$(TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL) ; \
+		export TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL=$(TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL) ; \
 		export TANZU_CLI_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_VERIFICATION_SKIP_LIST=$(TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL) ; \
 		export TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER="Yes" ; \
+		export ALLOWED_REGISTRY=$$(echo ${TANZU_CLI_E2E_TEST_CENTRAL_REPO_URL} | cut -f1 -d/ | cut -f1 -d:) ; \
```
anujc25, line 48: Why do we need to set this environment variable?

pkg/command/discovery_source.go
```diff
@@ -143,23 +164,18 @@ func newUpdateDiscoverySourceCmd() *cobra.Command {
...
 		Args:  cobra.ExactArgs(1),
-		Example: `
-    # Update a local discovery source. If URI is relative path, 
-    # $HOME/.config/tanzu-plugins will be considered base path
-    tanzu plugin source update standalone-local --type local --uri new/path/to/local/discovery
-
-    # Update an OCI discovery source. URI should be an OCI image.
-    tanzu plugin source update standalone-oci --type oci --uri projects.registry.vmware.com/tkg/tanzu-plugins/standalone:v1.0`,
-
 		RunE: func(cmd *cobra.Command, args []string) error {
```
anujc25, line 167: **Possible future enhancement:** Check whether the provided oci image is a valid plugin-inventory image or not if not throw an error without updating the discovery source.

marckhouzam: I like this idea.  It would provide a better UX.
However, after thinking about it some more, there is a scenario this would prevent:
- setup the CLI pointing to an air-gapped registry **before** the air-gapped image is ready

Do you think this is a scenario that an air-gapped operator may use?
If so, we could just print a warning that the OCI image configured is not valid, and the operator could tell their users to ignore this.

test/e2e/Makefile
marckhouzam: The trusted registry list used to automatically include whatever registry was added in `TANZU_CLI_PRE_RELEASE_REPO_IMAGE`.  Now that we use plugin sources instead, I didn't think we should automatically trust whatever registry was added as a plugin source (or else it seems to defeat the purpose of the `ALLOWED_REGISTRY` variable).

So, for the E2E tests, which use a different registry than `projects.registry.vmware.com` we need to set the variable.

Makes sense?

anujc25: > The trusted registry list used to automatically include whatever registry was added in TANZU_CLI_PRE_RELEASE_REPO_IMAGE. Now that we use plugin sources instead, I didn't think we should automatically trust whatever registry was added as a plugin source (or else it seems to defeat the purpose of the ALLOWED_REGISTRY variable).

Why not trust the registry, if the registry was added by the user with the `tanzu plugin source update` command? 
I feel that we should NOT ask users to configure any additional variable except updating the discovery source. Also `ALLOWED_REGISTRY` was added as a backup and we haven't documented this to be used by end users in production even for air-gapped.

pkg/command/discovery_source.go
marckhouzam: Ok, that make sense.  I'll work on this in a follow-up as it will probably require new functions to allow doing this from the command package.

anujc25 (approved):
The changes look good to me. Thanks!

PR #237, opened 2023-04-27
Pass a TANZU_BIN env var to plugins

### LINKED ISSUE #236: Provide plugins with the location of the tanzu CLI binary location

There are scenarios where a plugin needs to trigger a "tanzu plugin sync" or some other "tanzu" cli command.

One possibility to achieve this is for the CLI to set an environment variable with the CLI binary location before calling a plugin.  This is actually the approach helm uses to communicate such information to its plugins: https://helm.sh/docs/topics/plugins/#environment-variables

A plugin could simply call `tanzu`, with the assumption in can be found.  However, this may not correspond to the actual binary the user called to trigger the plugin.

**Additional context**

marckhouzam:

With this PR, when the Tanzu CLI executes a plugin, it injects an additional environment variable `TANZU_BIN` which provides the path to the `tanzu` command (as executed by the user).

If a plugin needs to trigger a Tanzu CLI operation, it can do so by externally calling the `tanzu` binary specified by the `TANZU_BIN` variable.

For example, the plugin can use the following to perform a `tanzu plugin sync`:

```go
cliPath := os.Getenv("TANZU_BIN")
command := exec.Command(cliPath, "plugin", "sync")
command.Stdout = os.Stdout
command.Stderr = os.Stderr
command.Run()
```

The PR also re-generates the local test central repos so that stub plugins now have an `env` sub-command which lists all the environment variables seen by the plugin.  This allows to easily test if a variable is properly passed to a plugin for this PR or any future new variable we may add.

Is `TANZU_BIN` an appropriate name or maybe `TANZU_CLI_BIN` would be better?  Or something else?

Fixes #236

```
$ tz plugin install package

[i] Installing plugin 'package:v9.9.9' with target 'kubernetes'
[ok] successfully installed 'package' plugin

# Try different invocation of the CLI
$ tanzu package env | grep TANZU_BIN
TANZU_BIN=tanzu
$ tz package env | grep TANZU_BIN
TANZU_BIN=/Users/kmarc/git/tanzu-cli/bin/tanzu
$ ./bin/tanzu package env | grep TANZU_BIN
TANZU_BIN=./bin/tanzu
```

I also created a brand new go plugin called `sync` and added the following to the `main.go` file:
```
func main() {
	p, err := plugin.NewPlugin(&descriptor)
	if err != nil {
		log.Fatal(err, "")
	}
	p.Cmd.RunE = func(_ *cobra.Command, _ []string) error {
		cliPath := os.Getenv("TANZU_BIN")
		command := exec.Command(cliPath, "plugin", "sync")
		command.Stdout = os.Stdout
		command.Stderr = os.Stderr
		return command.Run()
	}
	if err := p.Execute(); err != nil {
		os.Exit(1)
	}
}
```
Then I installed it and confirmed it could trigger a plugin sync:
```
$ tz plugin clean
$ tz plugin install sync --local $git/tanzu-sync-plugins/artifacts/plugins/darwin/amd64
[i] Installing plugin 'sync:v0.0.1' with target 'global'
[ok] successfully installed 'sync' plugin
$ tz sync
[i] Installing plugin 'cluster:v0.28.0' with target 'kubernetes'
[i] Installing plugin 'feature:v0.28.0' with target 'kubernetes'
[i] Installing plugin 'kubernetes-release:v0.28.0' with target 'kubernetes'
[i] Successfully installed all required plugins
[ok] Done
$ tz sync
[ok] Done
```

### Release note

```release-note
When calling a plugin, the Tanzu CLI now passes it an environment variable `TANZU_BIN` which provides the path to the `tanzu` command (as executed by the user).
```

The implementation of this feature was done in a way that simplifies the fact that other variables can easily be added if other information should be passed to plugins in the future.
For example, we could add a global log-level flag to the CLI and pass the value of the log-level to plugins so they respect it also.


pkg/cli/plugin_cmd.go
```diff
@@ -112,6 +113,19 @@ func getHelpArguments() []string {
 	return append(helpArgs, "-h")
 }
 
+// setupPluginEnv prepares some extra environment variables
+// that communicate certain information to plugins.
+func setupPluginEnv() {
+	env := make(map[string]string, 10)
```
anujc25, line 119: Why specifically the size of `10` here?

anujc25 (approved):
LGTM. Thanks for adding this.

pkg/cli/plugin_cmd.go
marckhouzam: Actually it technically should be `1` but when we add more variables, I was thinking we would not think to change that value, which would make things slightly less efficient, so I put `10` in preparation for 10 variables.

But as you know, none of this matters at the speeds we work with

Issue #223, opened 2023-04-19
Air-gapped support for tanzu-cli plugins


anujc25:
**Describe the feature request**
Provide a means to migrate tanzu-cli plugins from the default oci registry to the private registry for the airgapped environment.

These can be done by implementing the following or equivalent commands.
- [x] "tanzu plugin download-bundle" base command
- [x] "tanzu plugin upload-bundle" base command

Additional things to do:
- [x] Support plugin migration based on the plugin-group 
- [x] Verify the signature before downloading the plugin bundle
- [x] Write E2E test for the air-gapped environments

**Describe alternative(s) you've considered**
None

**Additional context**

PR #245, opened 2023-05-01
Implement air-gapped E2E tests for plugin bundle download, upload and lifecycle

anujc25:
- Implement air-gapped E2E tests for plugin bundle download, upload, and lifecycle
- Start a new air-gapped registry at `localhost:6001`
- Use `TANZU_CLI_E2E_AIRGAPPED_REPO` in the E2E tests to specify which air-gapped repository to use

**Test Cases Implemented:**

* Download the plugin bundle, Upload the plugin bundle, and plugin lifecycle tests with different plugin groups in the following sequence:
  * With Plugin Group 'vmware-tkg/v0.0.1'
  * With Plugin Group 'vmware-tmc/v9.9.9'
  * With Plugin Group 'vmware-tmc/v0.0.1'
  * Without specifying any plugin group to copy the entire plugin repository
* After each plugin bundle download and upload to the air-gapped repository verify the below things:
  * All migrated plugin groups are available in the air-gapped repository
  * All migrated plugins are available in the air-gapped repository
  * All migrated plugins can be installed from the air-gapped repository

Related to #223

### Release note

```release-note
None
```


hack/central-repo/upload-plugins.sh
```diff
@@ -275,7 +275,6 @@ cat $ROOT_DIR/create_tables.sql | sqlite3 -batch $database
 for name in ${globalPlugins[0]}; do
     addPlugin $name global true v0.0.1
     addPlugin $name global true v9.9.9
-    addGroup vmware tkg v9.9.9 $name global v9.9.9
```
anujc25, line 278: Before this change only the v9.9.9 version of global plugins were getting added to the `vmware-tkg/v9.9.9` group but not v0.0.1 plugins. Removed this to be consistent across versions.

test/e2e/airgapped/airgapped_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	Context("Download plugin bundle, Upload plugin bundle and plugin lifecycle tests with plugin group 'vmware-tkg/v0.0.1'", func() {
+		// Test case: download plugin bundle for plugin-group vmware-tkg/v0.0.1
+		It("download plugin bundle with specific plugin-group vmware-tkg/v0.0.1", func() {
+			err := tf.PluginCmd.DownloadPluginBundle(e2eTestLocalCentralRepoImage, []string{"vmware-tkg/v0.0.1"}, filepath.Join(tempDir, "plugin_bundle_vmware-tkg-v0.0.1.tar.gz"))
+			Expect(err).To(BeNil(), "should not get any error while downloading plugin bundle with specific group")
+		})
+
+		// Test case: upload plugin bundle downloaded using vmware-tkg/v0.0.1 plugin-group to the airgapped repository
+		It("upload plugin bundle that was downloaded using vmware-tkg/v0.0.1 plugin-group to the airgapped repository", func() {
```
chandrareddyp, line 28: is upload plugin bundle functionality overrides the repository if any available already or just appends?
if it appends, then better validate repository is empty before uploading.

test/e2e/framework/plugin_lifecycle_operations.go
```diff
@@ -62,11 +62,20 @@ type PluginGroupOps interface {
 	InstallPluginsFromGroup(pluginNameORAll, groupName string, opts ...E2EOption) error
 }
 
+type PluginDownloadAndUploadOps interface {
+	// DownloadPluginBundle downloads the plugin inventory and plugin bundles to local tar file
+	DownloadPluginBundle(image string, groups []string, toTar string, opts ...E2EOption) error
+
+	// UploadPluginBundle performs the uploading plugin bundle to the remote repository
+	UploadPluginBundle(toRepo string, tar string, opts ...E2EOption) error
```
chandrareddyp, line 70: Could you kindly update the document and clarify whether it should override or append the existing repository?

test/e2e/airgapped/airgapped_test.go
```diff
@@ -0,0 +1,212 @@
...
+		It("validate the plugins from group 'vmware-tkg/v0.0.1' exists", func() {
+			// search plugin groups
+			pluginGroups = pluginlifecyclee2e.SearchAllPluginGroups(tf)
+			// check all expected plugin groups are available in the `plugin group search` output from the airgapped repository
+			expectedPluginGroups := []*framework.PluginGroup{{Group: "vmware-tkg/v0.0.1"}}
+			Expect(framework.IsAllPluginGroupsExists(pluginGroups, expectedPluginGroups)).Should(BeTrue(), "all required plugin groups for life cycle tests should exists in plugin group search output")
+
+			// search plugins and make sure correct number of plugins available
+			// check expected plugins are available in the `plugin search` output from the airgapped repository
+			expectedPlugins := pluginsForPG_TKG_001
```
chandrareddyp, line 51: Is it possible to avoid hard-coding the plugin information? Instead, can we retrieve the plugin details prior to downloading and utilize them to validate the new airgaped repository?

test/e2e/airgapped/plugins.go
chandrareddyp, line 1: nit: as commented in the previous file, let's avoid hard-coding the plugins info if possible.

test/e2e/plugin_lifecycle/plugin_lifecycle_helper.go
```diff
@@ -37,3 +37,10 @@ func SearchAllPluginGroups(tf *framework.Framework, opts ...framework.E2EOption)
 	gomega.Expect(err).To(gomega.BeNil(), "should not get any error for plugin search")
 	return pluginGroups
 }
+
+// ListInstalledPlugins runs the plugin list command and returns all the plugins from the list output
```
chandrareddyp, line 41: nit: It would be best to relocate this to the framework_helper.go file and not validate error with Expect. Instead, it is better to return the error and let user handle to validate error.

chandrareddyp (commented):
Good work, i have added few comments. Thank you.

test/e2e/airgapped/airgapped_test.go
anujc25: It appends the plugins. Good point, I will add a validation that it was empty before uploading.

test/e2e/airgapped/airgapped_test.go
anujc25: I thought about it and the command that I was missing was `tanzu plugin group get` to get all plugins associated with a plugin group.

test/e2e/plugin_lifecycle/plugin_lifecycle_helper.go
anujc25: I removed this helper function and used `installedPluginList, err := tf.PluginCmd.ListInstalledPlugins()` function directly as it seemed unnecessary to add this new function.

test/e2e/airgapped/airgapped_test.go
anujc25: Actually, for our test cases, we start a new registry every time at `localhost:6001` and that registry doesn't have a plugin inventory database image created until we run first `UploadPluginBundle`. So, any attempt to get the list of the plugins might fail.

So to verify this, I added a check in `BeforeSuite` by verifying air-gapped plugin inventory image doesn't exists before the start of the tests.
```
// Make sure airgapped plugin inventory image doesn't exists before starting the tests
_, _, err = carvelhelpers.GetImageDigest(e2eAirgappedCentralRepoImage)
Expect(err).NotTo(BeNil(), "looks like airgapped plugin inventory image already exists before starting the airgapped tests. This should not be the case as it leads to incorrect test results")
```

chandrareddyp: However, since we use the same `e2eAirgappedCentralRepo` for all use cases, plugins may already be available when the second use case (`'plugin group 'vmware-tmc/v9.9.9'`) is running.
Before each use case, the register might need to be cleaned up or deleted. And tests are supposed to fail, except for the first use case, right? any thoughts?

test/e2e/airgapped/airgapped_test.go
chandrareddyp: It could be better to add a TODO statement to the code OR to include some commentary to the story that was created for implementing the command `tanzu plugin group get`.

test/e2e/airgapped/airgapped_test.go
anujc25: No, these tests are supposed to test the merging of the plugins that we are uploading. If we delete of clean the existing plugins we are not testing the correct behavior we want to test.

The test cases are written in a way that it builds the plugin-inventory by adding different plugin-groups in each phase and eventually builds the entire plugin-inventory in the last phase.

chandrareddyp: Okay got it, thank you.

chandrareddyp (approved):
LGTM!

test/e2e/airgapped/airgapped_test.go
anujc25: Sure. Added a TODO under `plugins.go` to remove the hardcoding.

hack/central-repo/upload-plugins.sh
marckhouzam: But then we don't have any group that contains global plugins, which may be useful to have.
How about adding the v0.01 instead of removing the 9.9.9 one?

Makefile
```diff
@@ -207,6 +207,14 @@ start-test-central-repo: stop-test-central-repo setup-custom-cert-for-test-centr
 stop-test-central-repo: ## Stops and removes the local test central repository
 	@docker container stop central > /dev/null 2>&1 && echo "Stopped docker test central repo" || true
 
+.PHONY: airgapped-local-registry
```
marckhouzam, line 210:
```suggestion
.PHONY: start-airgapped-local-registry
```

Makefile
```diff
@@ -207,6 +207,14 @@ start-test-central-repo: stop-test-central-repo setup-custom-cert-for-test-centr
...
+start-airgapped-local-registry: stop-airgapped-local-registry
+	docker run -d -p 6001:5000 --name temp-airgapped-local-registry mirror.gcr.io/library/registry:2
```
marckhouzam, line 212: FYI, but you can leave as is.

I usually like to add `--rm` when running the container so the container is automatically deleted on a stop.
Then you don't have to do `docker rm -v temp-airgapped-local-registry` in the stop target.

test/e2e/Makefile
```diff
@@ -89,3 +96,14 @@ e2e-context-tmc-test:
...
+
+.PHONY: e2e-airgapped-tests ## Execute CLI airgapped tests
+e2e-airgapped-tests:
+	export TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_IMAGE_FOR_AIRGAPPED) ; \
+	export TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_PUBLIC_KEY_PATH=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_PUBLIC_KEY_PATH) ; \
+	export TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_HOST=${TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_HOST} ; \
+	export TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_CA_CERT_PATH=${TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_CA_CERT_PATH} ; \
+	export TANZU_CLI_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_VERIFICATION_SKIP_LIST=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_IMAGE_FOR_AIRGAPPED) ; \
+	export TANZU_CLI_E2E_AIRGAPPED_REPO=$(TANZU_CLI_E2E_AIRGAPPED_REPO) ; \
+	export TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER="Yes" ; \
```
marckhouzam, line 108:
```suggestion
	export TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER="No" ; \
```
@prkalle Changed all of these to `No` to avoid having those tests report telemetry in the future.

hack/central-repo/upload-plugins.sh
anujc25: > But then we don't have any group that contains global plugins, which may be useful to have.
It doesn't matter for air-gapped testing. We need to have plugins that are not part of any plugin group for writing tests.

Makefile
anujc25: Makes sense. 👍

test/e2e/Makefile
marckhouzam: @anujc25 No biggy, but this was forgotten.  Can we do it in a follow-up?

test/e2e/Makefile
```diff
@@ -89,3 +96,14 @@ e2e-context-tmc-test:
...
 		${GO} test ${ROOT_DIR}/test/e2e/context/tmc -timeout ${E2E_TEST_TIMEOUT} -race -coverprofile ${E2E_TEST_OUTPUT} ${GOTEST_VERBOSE} ; \
 	fi
+
+.PHONY: e2e-airgapped-tests ## Execute CLI airgapped tests
+e2e-airgapped-tests:
+	export TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_IMAGE_FOR_AIRGAPPED) ; \
+	export TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_PUBLIC_KEY_PATH=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_PUBLIC_KEY_PATH) ; \
+	export TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_HOST=${TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_HOST} ; \
+	export TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_CA_CERT_PATH=${TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_CA_CERT_PATH} ; \
+	export TANZU_CLI_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_VERIFICATION_SKIP_LIST=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_IMAGE_FOR_AIRGAPPED) ; \
```
marckhouzam, line 106: Why do we skip signing verification here?
We've setup the proper key 3 lines above, so it would work, no?

PR #264, opened 2023-05-05
Fix vcs status error by allowing passing in build go flags to plugin builder

### LINKED ISSUE #263: With upgrade to Go 1.19, Tanzu CLI Plugin Builder returns 'error obtaining VCS status' error when being built in Git submodule (buildweb gobuild)
**Bug description**

After I upgraded all my go modules to use go version 1.19, and when I try to run builder from a Git submodule, I get this error: 

```
building local repository at ./artifacts/plugins, v0.1.0, [all]
ðŸ¦ - building plugin at path "cmd/plugin/feature"
ðŸ¦ - running godep path "cmd/plugin/feature"
ðŸ¦$ /usr/local/go/bin/go mod download
ðŸ¦ - $ /usr/local/go/bin/go build -o /workspace/artifacts/plugins/linux/arm64/kubernetes/feature/v0.1.0/tanzu-feature-linux_arm64 -ldflags  -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v0.1.0' -tags  ./.
ðŸ¦ - error: exit status 1
ðŸ¦ - output: error obtaining VCS status: exit status 128
 	Use -buildvcs=false to disable VCS stamping.
ðŸ¦ - error compiling plugin feature 
```

**Expected behavior**

I expect that plugins can be built from within a Git submodule using Go version 1.19 and above.

**Output of `tanzu version`**

codegold79:

This PR allows for the passing of Go flags to the `go build` command during the building of plugins using Tanzu Plugin Builder.

The reason I need it is because it will allow me to prevent a VCS status error that happens in our downstream build environment: 

```
building local repository at ./artifacts/plugins, v0.1.0, [all]
ðŸ¦ - building plugin at path "cmd/plugin/feature"
ðŸ¦ - running godep path "cmd/plugin/feature"
ðŸ¦$ /usr/local/go/bin/go mod download
ðŸ¦ - $ /usr/local/go/bin/go build -o /workspace/artifacts/plugins/linux/arm64/kubernetes/feature/v0.1.0/tanzu-feature-linux_arm64 -ldflags  -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v0.1.0' -tags  ./.
ðŸ¦ - error: exit status 1
ðŸ¦ - output: error obtaining VCS status: exit status 128
 	Use -buildvcs=false to disable VCS stamping.
ðŸ¦ - error compiling plugin feature 
```

Fixes #263 

I ran `builder plugin build` with and without the `--goflags '-buildvcs=false'` and the plugins built as intended. 

I also tested passing in the argument in our downstream build environment, and it worked without giving the vcs status error.

To test the Makefile changes and multiple go flags, I have also successfully run the commands,
- make plugin-build
- make plugin-build PLUGIN_GO_FLAGS="-buildvcs=false"
- make plugin-build PLUGIN_GO_FLAGS="-buildvcs=false -a"

Here is the output from running the above commands: [2023-05-09_make-plugin-build-log.txt](https://github.com/vmware-tanzu/tanzu-cli/files/11435225/2023-05-09_make-plugin-build-log.txt)

### Release note

```release-note
Allow for passing go flags when building plugins.
```

### DIFF
--- a/cmd/plugin/builder/command/cli_compile.go
+++ b/cmd/plugin/builder/command/cli_compile.go
@@ -14,6 +14,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"sync"
 	"time"
 
@@ -38,6 +39,7 @@ func init() {
 var (
 	version, artifactsDir, ldflags string
 	tags, goprivate                string
+	goflags                        string
 	targetArch                     []string
 	groupByOSArch                  bool
 )
@@ -59,6 +61,7 @@ type PluginCompileArgs struct {
 	ArtifactsDir               string
 	LDFlags                    string
 	Tags                       string
+	GoFlags                    string
 	Match                      string
 	Description                string
 	GoPrivate                  string
@@ -134,6 +137,7 @@ func setGlobals(compileArgs *PluginCompileArgs) {
 	goprivate = compileArgs.GoPrivate
 	targetArch = compileArgs.TargetArch
 	groupByOSArch = compileArgs.GroupByOSArch
+	goflags = compileArgs.GoFlags
 
 	// Append version specific ldflag by default so that user doesn't need to pass this ldflag always.
 	ldflags = fmt.Sprintf("%s -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=%s'", ldflags, version)
@@ -327,7 +331,7 @@ type target struct {
 	args []string
 }
 
-func (t target) build(targetPath, prefix, modPath, ldflags, tags string) error {
+func (t target) build(targetPath, prefix, modPath, ldflags, tags, goflags string) error {
 	cmd := goCommand("build")
 
 	var commonArgs = []string{
@@ -335,6 +339,9 @@ func (t target) build(targetPath, prefix, modPath, ldflags, tags string) error {
 		"-tags", tags,
 	}
 
+	if goflags != "" {
+		cmd.Args = append(cmd.Args, strings.Split(goflags, " ")...)
+	}
 	cmd.Args = append(cmd.Args, t.args...)
 	cmd.Args = append(cmd.Args, commonArgs...)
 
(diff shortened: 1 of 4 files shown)


cmd/plugin/builder/command/cli_compile.go
```diff
@@ -327,14 +330,17 @@ type target struct {
...
+func (t target) build(targetPath, prefix, modPath, ldflags, tags, goflags string) error {
 	cmd := goCommand("build")
 
 	var commonArgs = []string{
 		"-ldflags", ldflags,
 		"-tags", tags,
 	}
 
+	if goflags != "" {
+		cmd.Args = append(cmd.Args, goflags)
```
marckhouzam, line 342: I'm under the impression that as it is proposed, we would only be able to add a single goflag.  Say I want to add both `-buildvcs=false` and `-a`, I don't think I can.

Although we could allow for multiple `--goflags` flags, I think it would be simpler to accept multiple space-separated values for the single `--goflags` flag.
So, in the above line, you can probably just `strings.Split(goflags, " ")`

Then we should be able to do `tanzu builder plugin build --goflags "-buildvcs=false -a"`

marckhouzam (changes requested):
Thanks @codegold79.  This improvement will give the builder plugin more flexibility 👍 
In that optic let's make sure the change is generic.  Please see one comment in-line.

Also, if you don't mind could you integrate this change into `plugin-tooling.mk` so that users can use the makefile and still add goflags?  I suggest:
- at the top, after the `PLUGIN_LD_FLAGS` variable, adding `PLUGIN_GO_FLAGS ?=` to set this new variable to the empty string if it is not provided by the user
- in the `plugin-build-%` make target, add the line `--goflags "$(PLUGIN_GO_FLAGS)" \`

Then, we will be able to call `make plugin-build-local PLUGIN_GO_FLAGS="-buildvcs=false -a"` to include extra goflags.

 And finally, make the exact same makefile changes in `cmd/plugin/builder/template/plugintemplates/plugin-tooling.mk.tmpl`

Thanks!

cmd/plugin/builder/command/cli_compile.go
codegold79: Great suggestion. I have applied it in the latest commit.

marckhouzam (approved):
Very nice! LGTM
Thanks @codegold79 !

cc @anujc25 @vuil

vuil (approved):
lgtm, thanks!

vuil (approved):
lgtm, thanks!

PR #275, opened 2023-05-10
Specify a fixed tag for the registry:2 image

marckhouzam:

On multiple occasions we have seen the registry:2 tag disappear from the registry.  When this happens our CI breaks.  This causes each one of us affected to waste time repeatedly.

We have agreed to use a static tag to protect against this.

Fixes # N/A

Let CI run

### Release note

```release-note

```


anujc25 (approved):
LGTM.

PR #282, opened 2023-05-12
Remove unnecessary operations of "tanzu init"

marckhouzam:

The "tanzu init" command was previously required to install the TKG plugins through a "plugin sync" (when the CLI was tightly-coupled with TKG). Users would be told to call "tanzu init" after installation.

With the Independent CLI, "plugin sync" is no longer used in the same way.  Instead users use "plugin install --group <group>"; but since the group to use is dependent on the product of interest, this installation command cannot be part of a generic "tanzu init" command but has to be run manually by the user.

This commit therefore removes the "plugin sync" call from "tanzu init".

Furthermore, a empty read/write of the plugin catalog was being performed with the goal to normalize the contents of the catalog. This was added when the concept of CLI "targets" was introduced.  However, the logic to handle targets has been revamped and this normalization is no longer required. It has therefore been removed.

The "tanzu init" command is now a no-op.
We have decided to keep it for two reasons:
1. backwards-compatibility (some scripts and even users still call it)
2. the command may be needed by a future feature so we wanted to avoid removing it to re-introduce it some time later.

Fixes # N/A

For months now I have been running the Independent CLI without ever calling "tanzu init".
There have been different upgrade tests performed from old CLIs to the newest version, also without running "tanzu init".
This confirms the command was not doing anything useful.

And have CI pass.

### Release note

```release-note
The "tanzu init" command is now a no-op but is kept for possible future use.
```

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -246,17 +246,6 @@ func testPath() string {
 	return filepath.Join(pluginRoot, "test")
 }
 
-// UpdateCatalogCache when updating the core CLI from v0.x.x to v1.x.x. This is
-// needed to group the standalone plugins by context type.
-func UpdateCatalogCache() error {
-	c, err := getCatalogCache()
-	if err != nil {
-		return err
-	}
-
-	return saveCatalogCache(c)
-}
-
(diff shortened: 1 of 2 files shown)


anujc25 (approved):
LGTM

PR #283, opened 2023-05-12
Add support to configure proxy certificate to interact with registry

### LINKED ISSUE #284: Add support to configure proxy certificate to interact with registry
Add support to configure proxy certificate to interacgt with registry.
- CLI should allow user to add the proxy certificate for the proxy host using the command  `tanzu config cert add --host --ca-certificate /path/to/proxy/ca/cert`
- CLI should resolve if the proxy cert should be used based on environment variables (`http_proxy`, `https_proxy` and `no_proxy`) and if the proxy should be used, it should use the configured proxy cert provided by the user.

**Additional context**

prkalle:

This PR adds support to configure proxy certificate to interact with registry

Summary:
- If proxy is configured between CLI and registry which needs a certificate, user can set the environment variable "PROXY_CA_CERT" with base64 value of the proxy cert. The proxy cert would be used while interacting with registry.

Fixes #284 

Unit tests and CI passed.

### Release note

```release-note
Add support to configure proxy certificate to interact with registry. User has to set the environment variable "TANZU_CLI_PROXY_CA_CERT" value to base64 value of proxy certificate
```

### DIFF
--- a/docs/quickstart/install.md
+++ b/docs/quickstart/install.md
@@ -242,3 +242,9 @@ interacting with the registry.
 Users can update or delete the certificate configuration using the `tanzu config cert update`
 and `tanzu config cert delete` commands.
 Also, users can list the certificate configuration using the `tanzu config cert list` command.
+
+#### Proxy CA certificate
+
+If the user configured a proxy between the Tanzu CLI and the central repository and if the proxy certificate
+needs to be configured, the user should set the environment variable `PROXY_CA_CERT` with base64 value of
+proxy CA certificate.
--- a/pkg/registry/helpers.go
+++ b/pkg/registry/helpers.go
@@ -18,14 +18,14 @@ import (
 	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 )
 
-type RegistryCertOptions struct {
+type CertOptions struct {
 	CACertPaths    []string
 	SkipCertVerify bool
 	Insecure       bool
 }
 
-func GetRegistryCertOptions(registryHost string) (*RegistryCertOptions, error) {
-	registryCertOpts := &RegistryCertOptions{
+func GetRegistryCertOptions(registryHost string) (*CertOptions, error) {
+	registryCertOpts := &CertOptions{
 		SkipCertVerify: false,
 		Insecure:       false,
 	}
@@ -39,6 +39,10 @@ func GetRegistryCertOptions(registryHost string) (*RegistryCertOptions, error) {
 
 	// check if the custom cert data is configured for the registry
 	if exists, _ := configlib.CertExists(registryHost); !exists {
+		err := checkForProxyConfigAndUpdateCert(registryCertOpts)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to check for proxy config and update the cert")
+		}
 		return registryCertOpts, nil
 	}
 	cert, err := configlib.GetCert(registryHost)
@@ -51,11 +55,15 @@ func GetRegistryCertOptions(registryHost string) (*RegistryCertOptions, error) {
 		return nil, errors.Wrapf(err, "failed to updated the registry cert options")
 	}
 
+	err = checkForProxyConfigAndUpdateCert(registryCertOpts)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to check for proxy config and update the cert")
+	}
 	return registryCertOpts, nil
 }
 
 // updateRegistryCertOptions sets the registry options by taking the custom certificate data configured for registry as input
-func updateRegistryCertOptions(cert *configtypes.Cert, registryCertOpts *RegistryCertOptions) error {
+func updateRegistryCertOptions(cert *configtypes.Cert, registryCertOpts *CertOptions) error {
 	if cert.SkipCertVerify != "" {
 		skipVerifyCerts, _ := strconv.ParseBool(cert.SkipCertVerify)
 		registryCertOpts.SkipCertVerify = skipVerifyCerts
@@ -65,28 +73,16 @@ func updateRegistryCertOptions(cert *configtypes.Cert, registryCertOpts *Registr
 		registryCertOpts.Insecure = insecure
 	}
 
-	if cert.CACertData != "" {
-		caCertBytes, err := base64.StdEncoding.DecodeString(cert.CACertData)
-		if err != nil {
-			return errors.Wrap(err, "unable to decode the base64-encoded custom registry CA certificate string")
-		}
-		if len(caCertBytes) != 0 {
-			filePath, err := configpaths.GetRegistryCertFile()
-			if err != nil {
-				return err
-			}
-			err = os.WriteFile(filePath, caCertBytes, 0o644)
-			if err != nil {
-				return errors.Wrapf(err, "failed to write the custom image registry CA cert to file '%s'", filePath)
-			}
-			registryCertOpts.CACertPaths = append(registryCertOpts.CACertPaths, filePath)
-		}
+	err := updateCACertData(cert.CACertData, registryCertOpts)
+	if err != nil {
+		return err
 	}
+
 	return nil
 }
 
 // AddRegistryTrustedRootCertsFileForWindows adds CA certificate to registry options for Windows environments
-func AddRegistryTrustedRootCertsFileForWindows(registryCertOpts *RegistryCertOptions) error {
+func AddRegistryTrustedRootCertsFileForWindows(registryCertOpts *CertOptions) error {
 	filePath, err := configpaths.GetRegistryTrustedCACertFileForWindows()
 	if err != nil {
 		return err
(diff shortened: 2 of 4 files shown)


pkg/registry/helpers.go
```diff
@@ -108,3 +112,68 @@ func GetRegistryName(imageName string) (string, error) {
...
+
+	// check if user provided cert configuration for proxy host, if so, use only the CACertData
+	if exists, _ := configlib.CertExists(proxyURL.Host); !exists {
+		return nil
+	}
+	// get cert provided for the proxy
+	cert, err := configlib.GetCert(proxyURL.Host)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get the custom certificate configuration for host %q", registryHost)
+	}
```
anujc25, line 136: Is it possible to combine these 2 calls into 1 `GetCert` call and `return nil` if an error occurs?

anujc25 (approved):
LGTM.

anujc25:
We should add documentation on how to use a proxy with a self-signed certificate.

marckhouzam (commented):
From what I can see, if a user has both a cert config for the central repo and another for the proxy, the code will take the "insecure" and "skip" settings from the central repo cert config, and the "ca_cert" config from the proxy cert setting.

Is this what we want?

Also, currently we ignore any  "insecure" and "skip" cert config for a proxy.  Do we want to use them instead?  Does it make sense even?

I'm sorry but I'm not very comfortable with the Proxy usecase...

prkalle:
> From what I can see, if a user has both a cert config for the central repo and another for the proxy, the code will take the "insecure" and "skip" settings from the central repo cert config, and the "ca_cert" config from the proxy cert setting.
> 
> Is this what we want?
> 
> Also, currently we ignore any "insecure" and "skip" cert config for a proxy. Do we want to use them instead? Does it make sense even?
> 
> I'm sorry but I'm not very comfortable with the Proxy usecase...

Update: To avoid confusion, after discussion, for now, we decided to go with the environment variable option for user to specify proxy ca cert as it would be inline with the exisiting implementation, because user would be confused to configure "insecure" option for proxy cert configuration, as "insure" option doesn't apply for proxy , it rather apply for end server.

pkg/constants/env_variables.go
```diff
@@ -5,7 +5,7 @@ package constants
 
 // environment variables for http proxy
 const (
-	ProxyCACert = "PROXY_CA_CERT"
+	ProxyCACert = "TANZU_CLI_PROXY_CA_CERT"
```
chandrareddyp, line 8: `TANZU_CLI_PROXY_CA_CERT` is used for self-signed CA and for the proxy use case?

marckhouzam (commented):
I'm curious as to why you chose `TANZU_CLI_PROXY_CA_CERT` instead of the original `PROXY_CA_CERT` name?  Is it for a pass-through proxy that would need a different certificate for the CLI's central repo compared to what TKG would use as a certificate?

marckhouzam (approved):
LGTM

I was actually able to test it a bit (I hope it a valid test):

```
# Prepare the CLI
$ rm ~/.config/tanzu/config*
$ tz ceip-participation set false
$ tz config eula accept
[ok] Marking agreement as accepted.
[...]
$ tz config set env.TANZU_CLI_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_PUBLIC_KEY_PATH $tc/hack/central-repo/cosign-key-pair/cosign.pub

# Verify we can read the DB using the ca_certificate that was added to the config file
$ tz plugin group search
  GROUP                DESCRIPTION                   LATEST
  vmware-tkg/default   Desc for vmware-tkg/default   v9.9.9
  vmware-tmc/tmc-user  Desc for vmware-tmc/tmc-user  v9.9.9

# Extract the ca certificate from the config
$ cert=$(tz config get | \grep caCertData | cut -f8 -d" ")

# Remove the certificate from the config file
$ tz config cert delete localhost:9876
[ok] deleted certificate data for host localhost:9876

# Check that without the certificate, we cannot talk to the registry
$ tz plugin group search
[!] Unable to resolve the plugin discovery image: error getting the image digest: Error while preparing a transport to talk with the registry: Unable to create round tripper: Get "https://localhost:9876/v2/": x509: “localhost.local” certificate is not standards compliant; Get "http://localhost:9876/v2/": net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x15\x03\x01\x00\x02\x02"
[x] Fatal: plugins discovery image resolution failed. Please check that the repository image URL "localhost:9876/tanzu-cli/plugins/central:small" is correct

# Pretend we have a proxy by setting the new variable to the cert we extrated
# and confirm we can now talk to the registry
$ TANZU_CLI_PROXY_CA_CERT=$cert tz plugin group search
  GROUP                DESCRIPTION                   LATEST
  vmware-tkg/default   Desc for vmware-tkg/default   v9.9.9
  vmware-tmc/tmc-user  Desc for vmware-tmc/tmc-user  v9.9.9
```

pkg/registry/helpers.go
```diff
@@ -108,3 +104,39 @@ func GetRegistryName(imageName string) (string, error) {
...
+				return err
+			}
+			err = os.WriteFile(filePath, caCertBytes, 0o644)
+			if err != nil {
+				return errors.Wrapf(err, "failed to write the custom image registry CA cert to file '%s'", filePath)
+			}
+			registryCertOpts.CACertPaths = append(registryCertOpts.CACertPaths, filePath)
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
marckhouzam: I think we need to return nil here because the way the code is written it is this function that ignores an empty ca certificate configuration, which is a valid configuration, for example whenever `TANZU_CLI_PROXY_CA_CERT` is not set.

pkg/registry/helpers_test.go
```diff
@@ -113,5 +113,131 @@ var _ = Describe("config cert command tests", func() {
...
+			cert := &configtypes.Cert{
+				Host:           testHost,
+				CACertData:     caCertDataOptB64,
+				SkipCertVerify: skipCertVerifyOpt,
+				Insecure:       insecureOpt,
+			}
+			err := configlib.SetCert(cert)
+			Expect(err).To(BeNil())
+			os.Setenv("no_proxy", noProxyOpt)
+			os.Setenv("TANZU_CLI_PROXY_CA_CERT", proxyCACertDataB64)
```
anujc25, line 174: nit: Might be better to just use constant here for `TANZU_CLI_PROXY_CA_CERT`.

docs/quickstart/install.md
```diff
@@ -242,3 +242,9 @@ interacting with the registry.
 Users can update or delete the certificate configuration using the `tanzu config cert update`
 and `tanzu config cert delete` commands.
 Also, users can list the certificate configuration using the `tanzu config cert list` command.
+
+#### Proxy CA certificate
+
+If the user configured a proxy between core CLI and the central repository and if the proxy certificate
```
anujc25, line 248:
```suggestion
If the user configured a proxy between the Tanzu CLI and the central repository and if the proxy certificate
```

prkalle:
> I'm curious as to why you chose `TANZU_CLI_PROXY_CA_CERT` instead of the original `PROXY_CA_CERT` name? Is it for a pass-through proxy that would need a different certificate for the CLI's central repo compared to what TKG would use as a certificate?

It is just to be consistent with naming variables with `TANZU_CLI` prefix and also I feel it would be better to have a sperate variable for CLI , just incase if TKG uses the environmnet variable for cluster use cases where it may not be applicable to CLI

pkg/constants/env_variables.go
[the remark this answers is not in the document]
prkalle: Yes, it is just to be consistent with naming variables with TANZU_CLI prefix and also I feel it would be better to have a sperate variable for CLI , just incase if TKG uses the environmnet variable for cluster use cases where it may not be applicable to CLI. Having said that, I see your point, that any organization would be having a single proxy and it would be applicable for all the cases. I can change it to `PROXY_CA_CERT` so that it won't cause any backward compatibility issue with existing CLI

pkg/registry/helpers_test.go
prkalle: updated

docs/quickstart/install.md
prkalle: updated

pkg/registry/helpers.go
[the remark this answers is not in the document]
prkalle: Yes, what @marckhouzam said. Thanks

pkg/constants/env_variables.go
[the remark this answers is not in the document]
anujc25: Yeah. I feel using the old variable name makes sense. Especially considering this will go hand in hand with `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, and `PROXY_CA_CERT`.

anujc25 (approved):
LGTM. Thanks for quickly updating the implementation.

marckhouzam (approved):
LGTM
Thanks @prem!

PR #286, opened 2023-05-13
Fail command if EULA terms are not accepted

vuil:
When invoking a command and it is deemed necessary to display the EULA prompt to solicit acceptance by the user, fail the command unless acceptance is received, with a message stating that the terms have to have to be accepted

Also: updated unit tests for eula show as well as situations when the prompt should be auto-triggered.

Fixes #

Ran unit tests.
Also ran arbitrary command e..g `tanzu context list`
to verify that it fails should EULA not be accepted

```
> tanzu context list
? You must agree to the VMware General Terms in order to install software via
Tanzu CLI.  The term of acceptance of the VMware General Terms will cover all
software installed via the Tanzu CLI until/unless there's a change to VMware
General Terms, a major release of the Tanzu CLI has been installed, or the
underlying software distribution registry has been changed.

To view the VMware General Terms, please see https://network.tanzu.vmware.com/legal_document_agreements/1982364

 No
The Tanzu CLI is only usable with reduced functionality until the General Terms are agreed to.
Please use `tanzu config eula show` to review the terms, or `tanzu config eula accept` to accept them directly
[x] : terms not accepted
```

### Release note

```release-note
Commands that prompts for EULA acceptance fails should EULA not be accepted.
```

### DIFF
--- a/pkg/command/eula_test.go
+++ b/pkg/command/eula_test.go
@@ -8,10 +8,24 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
 
 	"github.com/vmware-tanzu/tanzu-cli/pkg/constants"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
 )
+
+// Executes the command and verify if prompt is expected to be shown or not
+func checkForPromptOnExecute(cmd *cobra.Command, expectPrompt bool) {
+	err := cmd.Execute()
+	// Survey prompts when run without attached tty will fail. Use this fact
+	// to help detect that a prompt is indeed presented.
+	if expectPrompt {
+		Expect(err).ToNot(BeNil())
+		Expect(err.Error()).To(ContainSubstring("prompt failed"))
+	} else {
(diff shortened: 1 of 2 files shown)


pkg/command/eula_test.go
```diff
@@ -52,5 +69,46 @@ var _ = Describe("EULA command tests", func() {
...
+		})
+		Context("When invoking an arbitrary command", func() {
+			It("should invoke the eula prompt if EULA has not been accepted", func() {
+				os.Setenv("TANZU_CLI_EULA_PROMPT_ANSWER", "No")
+				cmd, err := NewRootCmd()
+				Expect(err).To(BeNil())
+				cmd.SetArgs([]string{"context", "list"})
+				err = cmd.Execute()
+				Expect(err).ToNot(BeNil())
+				//Expect(err).To(BeNil())
```
marckhouzam, line 98: Old commented line?

marckhouzam (approved):
I thought you purposefully had chosen to allow commands to run even if the terms weren't accepted and that the repeated EULA prompt would be annoying enough to effectively disable the CLI.

But the more explicit approach of this PR makes it clear to users that the EULA is fundamental.

LGTM

PR #301, opened 2023-05-22
Bug Fix: Add plugin binary File empty check in build  plugins step

mpanchajanya:
- Added plugin binary file empty check in build packages step

Fixes #

- Added unit tests for the util helper functions

### Release note

```release-note
empty plugin binary file will no longer be published
```

### DIFF
--- a/cmd/plugin/builder/helpers/helper.go
+++ b/cmd/plugin/builder/helpers/helper.go
@@ -12,9 +12,12 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+
 	"gopkg.in/yaml.v3"
 
 	"github.com/vmware-tanzu/tanzu-cli/pkg/cli"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/utils"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
 )
 
 // ReadPluginManifest reads the PluginManifest file and returns Manifest object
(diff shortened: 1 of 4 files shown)


cmd/plugin/builder/helpers/helper.go
```diff
@@ -12,8 +12,12 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
+
 	"gopkg.in/yaml.v3"
 
+	"github.com/vmware-tanzu/tanzu-cli/pkg/utils"
+
```
anujc25, line 21: nit: Extra new line.

cmd/plugin/builder/plugin/build-packages.go
```diff
@@ -57,12 +57,25 @@ func (bpo *BuildPluginPackageOptions) BuildPluginPackages() error {
...
+				// Continue to build other plugins and skip if validation on current plugin binary fails
+				if err != nil {
+					continue
+				}
+
+				// Continue to build other plugins and skip if validation on current plugin binary fails
+				if !valid {
+					continue
+				}
+
```
anujc25, line 72: We should return an error here if the binary file exists but it is not a valid file.

In the previous step, we skip it and continue because it is possible that the user has not provided the plugin binary for one of the os_arch combinations. But if binary exists we should throw an error if validation fails.

mpanchajanya: makes sense. Updated to return err if validation fails

cmd/plugin/builder/helpers/helper.go
mpanchajanya: removed the extra lines

anujc25 (approved):
LGTM.

Please update the PR description as well as the release notes.

PR #303, opened 2023-05-23
Remove extra slash from log

mpanchajanya:

Fixes #

### Release note

```release-note

```


pkg/airgapped/plugin_bundle_upload.go
```diff
@@ -82,7 +81,7 @@ func (o *UploadPluginBundleOptions) UploadPluginBundle() error {
 	}
 
 	log.Infof("---------------------------")
-	log.Infof("successfully published all plugin images to %q", fmt.Sprintf("%s%s", o.DestinationRepo, manifest.RelativeInventoryImagePathWithTag))
+	log.Infof("successfully published all plugin images to %q", filepath.Join(o.DestinationRepo, manifest.RelativeInventoryImagePathWithTag))
```
anujc25, line 84: I have a feeling that this might cause issues on the Windows machine as `filepath.Join` uses `\` for windows machines. Can we verify this once on windows?

mpanchajanya: @anujc25  I have looked into filepath.Join() functionality- Its implemention is OS aware.

Here's how it works for Mac, Linux, and Windows:

Mac and Linux: For these Unix-based operating systems, filepath.Join() uses / as the path separator, which is the standard for Unix-like systems. If you're running this function on either Mac or Linux, it will automatically use / to join the paths. For example, filepath.Join("/home", "user", "documents") would return "/home/user/documents".

Windows: Windows uses a different default path separator, the backslash \. When filepath.Join() is used in Windows, it will use \ to join the paths. For example, filepath.Join("C:\\", "Users", "Username", "Documents") would return "C:\\Users\\Username\\Documents".

anujc25: Considering this is a container image path and not the file path. if we use `filepath.Join` on windows it will use `\` in the container image which will make the container image invalid as the container image path should always have `/`. Correct?

pkg/utils/common_test.go
```diff
@@ -27,4 +27,29 @@ var _ = Describe("Unit tests for the common utils", func() {
...
+			{"https://www.example.com", "/test/path/", "https://www.example.com/test/path"},
+			{"https://www.example.com", "test/path/", "https://www.example.com/test/path"},
+			{"https://www.example.com/", "test/path/", "https://www.example.com/test/path"},
+			{"https://www.example.com", "test/path", "https://www.example.com/test/path"},
+			{"https://www.example.com/", "/test/path", "https://www.example.com/test/path"},
+			{"https://www.example.com/", "test/path", "https://www.example.com/test/path"},
+			{"https://www.example.com", "/test/path", "https://www.example.com/test/path"},
+			{"https://www.example.com", "", "https://www.example.com"},
+			{"https://www.example.com/", "", "https://www.example.com/"},
+			{"https://www.example.com/", "/", "https://www.example.com/"},
```
anujc25, line 47: Can we add a unit test that doesn't contain the URL schema (without `https://` at the front)? This is because the container repository will not have the schema defined and I have a feeling that `url.Parse` requires the schema to be defined.

Like 
```
{"www.example.com", "/test/path", "www.example.com/test/path"},
{"example.com/", "/test/path", "example.com/test/path"},
```

mpanchajanya: Done. Added

mpanchajanya: Added more test cases

anujc25: Thanks for verifying.

pkg/airgapped/plugin_bundle_upload.go
```diff
@@ -4,12 +4,15 @@
...
 
 	"github.com/pkg/errors"
+
 	"github.com/verybluebot/tarinator-go"
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/utils"
+
 	"gopkg.in/yaml.v3"
 
 	"github.com/vmware-tanzu/tanzu-cli/pkg/carvelhelpers"
```
anujc25, line 18: nit: It is always better to combine imports of the same repository together.

anujc25 (approved):
LGTM. Thanks

marckhouzam (approved):
LGTM after addressing @anujc25 nit

pkg/airgapped/plugin_bundle_upload.go
mpanchajanya: updated the imports

PR #318, opened 2023-05-26
Add changes to ignore the `tanzu pinniped-auth` command for EULA and CEIP prompts

prkalle:
This PR add changes to ignore the "tanzu pinniped-auth" command for EULA and CEIP prompts.

Changes summary:

- Since the "tanzu pinniped-auth" command would be part of kubeconfig generated with pinniped-auth, the kubectl runs the tanzu pinniped-auth login command to get the cluster credentials dynamically. Since it is executed as part of kubectl, EULA and CEIP prompts won't be visible to user, so ignoring these prompts for tanzu pinniped-auth command.

Fixes #

Ran the kubectl command with pinniped-kubeconfig and the command got stuck forever without the prompt being visible. Later when I exit the command(CTRL+c) i can see the EULA prompt failed

```
❯ kubectl get pods -A --kubeconfig ~/.kube-tanzu/config
[x] : failed to get EULA status: prompt failed: interrupt
Unable to connect to the server: getting credentials: exec: executable tanzu failed with exit code 1
```

With the changes, I ran the same command and it ran successfully 

```
❯ kubectl get pods -A --kubeconfig ~/.kube-tanzu/config
Error from server (Forbidden): pods is forbidden: User "pkalle" cannot list resource "pods" in API group "" at the cluster scope
```

### Release note

```release-note

```

### DIFF
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -23,8 +23,6 @@ import (
 	"github.com/vmware-tanzu/tanzu-cli/pkg/pluginsupplier"
 )
 
-const cmdNameSet = "set"
-
 // NewRootCmd creates a root command.
 func NewRootCmd() (*cobra.Command, error) {
 	var rootCmd = newRootCmd()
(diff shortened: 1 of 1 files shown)


pkg/command/root.go
```diff
@@ -125,9 +127,12 @@ func newRootCmd() *cobra.Command {
 			//   - 'config set' (would be a chicken and egg issue if user tries to set CEIP configuration
 			//      using "tanzu config set env.TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER yes")
 			//   - 'config eula xxx', 'ceip set' (auto prompting when running these commands is confusing)
+			//   - 'tanzu pinniped-auth' ( since this command is being invoked by kubectl exec plugin, user doesn't
+			//      get to see the prompts and kubectl command execution just gets stuck and it is very hard for users to figure out)
 			if cmd.Name() != cobra.ShellCompRequestCmd && cmd.Name() != completionCmd.Name() &&
 				cmd.Name() != newVersionCmd().Name() &&
-				!(cmd.Name() == cmdNameSet && cmd.Parent().Name() == configCmd.Name()) {
+				!(cmd.Name() == cmdNameSet && cmd.Parent().Name() == configCmd.Name()) &&
+				!(cmd.Name() == cmdNamePinnipedAuth && cmd.Parent().Name() == cmdNameTanzu) {
```
marckhouzam, line 135: See next comment first as it may cancel this one.

--

Can a command called "pinniped-auth" be called in any other situation than `tanzu pinniped-auth`?
It may be superfluous to check that the parent is `tanzu`, in which case we don't need to create the `cmdNameTanzu` constant.
I'm asking because if we are to create a constant `cmdNameTanzu` for the name of the root command, we should probably use it at line 114 above also.

pkg/command/root.go
```diff
@@ -125,9 +127,12 @@ func newRootCmd() *cobra.Command {
...
 				if !(cmd.Parent().Name() == newEULACmd().Name()) &&
 					!(cmd.Name() == cmdNameSet && cmd.Parent().Name() == newCEIPParticipationCmd().Name()) {
 					if err := cliconfig.ConfigureEULA(false); err != nil {
```
marckhouzam, line 138: I like the generic approach to this code, using constants and command names.  However, it has grown to a point that I personally find it difficult to understand.  Furthermore, with this PR, three constants used here are simply defined further above, which does not really help make the code generic.

Are we at a point where it would be better to hard-code the commands we want to ignore?  We can use cobra's `cmd.CommandPath()` to help.  We could do something like this:
```
	// Prompt user for EULA and CEIP agreement if necessary, except for
	skipCommands := []string{
		// The shell completion setup is not interactive, so it should not trigger a prompt
		"tanzu __complete",
		"tanzu completion",
		// Common first command to run,
		"tanzu version",
		// It would be a chicken and egg issue if user tries to set CEIP configuration
		// using "tanzu config set env.TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER yes"
		"tanzu config set",
		// Auto prompting when running these commands is confusing
		"tanzu config eula",
		"tanzu ceip-participation set",
		// This command is being invoked by the kubectl exec binary where the user doesn't
		// get to see the prompts and the kubectl command execution just gets stuck and it
		// is very hard for users to figure out what is going wrong
		"tanzu pinniped-auth",
	}
	skipPrompts := false
	for _, cmdPath := range skipCommands {
		if strings.HasPrefix(cmd.CommandPath(), cmdPath) {
			skipPrompts = true
			break
		}
	}
	if !skipPrompts {
            do the prompts
        }
       return nil
```
Do you think this increased readability is worth the reduced future-proof aspect?
@prkalle @vuil @anujc25

marckhouzam (commented):
I suggested a hard-coding approach to improve readability.
I'd like to get your opinion on which you think is better as this stage.

pkg/command/root.go
prkalle: Thanks @marckhouzam. I like this approach and it improves the readability. I will update the PR.

pkg/command/root.go
prkalle: Updated the PR as per the next comment. So it is not applicable anymore.

marckhouzam (approved):
LGTM
Thanks @prkalle

pkg/command/root.go
```diff
@@ -5,11 +5,11 @@
 package command
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"strings"
 
+	"github.com/pkg/errors"
```
marckhouzam, line 12: Actually, is this change ok?

prkalle: That should be fine. But let me revert it to the actual package(as there is no change that requires change in package)

PR #334, opened 2023-06-01
Use an argument to "context create" for the name

marckhouzam:

Instead of `tanzu context create --name <name>` we now prefer the user of an argument: `tanzu context create <name>`
We deprecate the "--name" flag but continue supporting it if the argument is not specified to avoid breaking any scripts.

Fixes # N/A

WILL BE ADDED SHORTLY

### Release note

```release-note
The `tanzu context create` no longer uses a `--name` flag. Instead, the name of the context is to be specified as an argument.
```

### DIFF
--- a/pkg/command/context.go
+++ b/pkg/command/context.go
@@ -89,15 +89,16 @@ func init() {
 }
 
 var createCtxCmd = &cobra.Command{
-	Use:   "create",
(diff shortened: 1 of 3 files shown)


anujc25 (approved):
LGTM

PR #339, opened 2023-06-05
E2E: Boolean flags must use the `=` format

marckhouzam:

Fix E2E test from TMC

Fixes # N/A

Let CI run

### Release note

```release-note

```

### DIFF
--- a/test/e2e/framework/cmd_exec_operations.go
+++ b/test/e2e/framework/cmd_exec_operations.go
@@ -76,7 +76,7 @@ func (co *cmdOps) Exec(command string, opts ...E2EOption) (stdOut, stdErr *bytes
 	cmd.Stderr = &stderr
 	err = cmd.Run()
 	if err != nil {
-		return &stdout, &stderr, fmt.Errorf("error while running '%s', stdOut: %s, stdErr: %s, err: %s", command, stdout.String(), stderr.String(), err.Error())
+		return &stdout, &stderr, fmt.Errorf("error while running '%s' (detailed arguments: %q), stdOut: %s, stdErr: %s, err: %s", command, cmdArgs, stdout.String(), stderr.String(), err.Error())
 	}
 	return &stdout, &stderr, err
 }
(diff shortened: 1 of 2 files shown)


test/e2e/framework/cmd_exec_operations.go
```diff
@@ -76,7 +76,7 @@ func (co *cmdOps) Exec(command string, opts ...E2EOption) (stdOut, stdErr *bytes
 	cmd.Stderr = &stderr
 	err = cmd.Run()
 	if err != nil {
-		return &stdout, &stderr, fmt.Errorf("error while running '%s', stdOut: %s, stdErr: %s, err: %s", command, stdout.String(), stderr.String(), err.Error())
+		return &stdout, &stderr, fmt.Errorf("error while running '%s' (detailed arguments: %q), stdOut: %s, stdErr: %s, err: %s", command, cmdArgs, stdout.String(), stderr.String(), err.Error())
```
marckhouzam, line 79: @chandrareddyp This will change the output for any error in the E2E tests.
Are you ok with that?

chandrareddyp: The `command` includes arguments, so there is no need to add `cmdArgs` in the logs as it is redundant.

marckhouzam: Here is an error message where listing the arguments show an error:

```
"error while running 'tanzu context create --endpoint http://localhost:8080 --staging plugin-sync-tmc-ke9C  --force-csp=true' (detailed arguments: [\"context\" \"create\" \"--endpoint\" \"http://localhost:8080\" \"--staging\" \"plugin-sync-tmc-ke9C\" \"\" \"--force-csp=true\"]), stdOut: , stdErr: [x] : accepts at most 1 arg(s), received 2\n, err: exit status 1"
```
Notice the extra space before `--force-csp` in the command printout.  When trying this on the shell locally, it is not a problem.  But if you look at the list of arguments, which is what the tests really use, there is an extra _**empty**_ argument before `--force-csp=true`, and that makes the code fail.

That is why I suggest printing the list of arguments.

chandrareddyp: Okay got it.

chandrareddyp (approved):

test/e2e/framework/cmd_exec_operations.go
marckhouzam: Thanks.  I reverted the breaking change that illustrated the problem and will merge once CI passes.

PR #349, opened 2023-06-09
Update compatibility matrix with the actual release data instead of examples

anujc25:

This PR updates the Tanzu Plugin Runtime and Tanzu CLI compatibility matrix with the official release information instead of using hypothetical example data.

To easily view the updated policy as part of this PR see [here](https://github.com/vmware-tanzu/tanzu-cli/blob/update-compatibility-matrix/docs/full/policy.md#tanzu-plugin-runtimetanzu-cli-compatibility-policy).

Fixes #

### Release note

```release-note
NONE
```


docs/full/policy.md
```diff
@@ -15,38 +15,30 @@ Given this scheme, the primary vehicle for changes is minor releases. The deprec
...
 - ***FAQ***:
-  - What happens if I upgrade to a newer version of CLI v1.2.0 and use a plugin developed with Plugin Runtime v1.0.5?
-    - As CLI is always compatible with all existing plugins, plugins developed with Plugin Runtime v1.0.5 should continue to work.
-  - If I have developed a plugin with pre-v1.0 (v0.11, 0.25, 0.28) of Tanzu Plugin Runtime (from Tanzu-Framework repository), Will my plugins be compatible with v1.0 Tanzu CLI?
-    - Yes. As shown in the above diagram and the backward compatibility guarantees provided with the v1.0 release of Tanzu CLI, all plugins developed with pre-v1.0 of Tanzu Plugin Runtime will be compatible with v1.0 Tanzu CLI.
-  - As a pre-v1.0 Tanzu CLI plugin developer, What are the possible changes I need to do when consuming v1.0 of Tanzu Plugin Runtime?
-    - We are still working on the design and the possible changes required when consuming v1.0 of Tanzu Plugin Runtime but high-level changes are mentioned in the Impact on Plugin Developers section below.
+  - What happens if I upgrade to a newer version of CLI v0.90.0 and use a plugin developed with Plugin Runtime v0.28.0?
```
marckhouzam, line 35:
```suggestion
  - What happens if I upgrade to the CLI v0.90.0 and use a plugin developed with Plugin Runtime v0.28.0?
```

docs/full/policy.md
```diff
@@ -15,38 +15,30 @@ Given this scheme, the primary vehicle for changes is minor releases. The deprec
...
+    - As CLI is always compatible with all existing plugins, plugins developed with Plugin Runtime v0.28.0 should continue to work.
```
marckhouzam, line 36:
```suggestion
    - As the CLI is always compatible with all existing plugins, plugins developed with Plugin Runtime v0.28.0 will continue to work.
```

docs/full/policy.md
```diff
@@ -15,38 +15,30 @@ Given this scheme, the primary vehicle for changes is minor releases. The deprec
...
+  - If I have developed a plugin with (v0.11, 0.25, 0.28, 0.29) of Tanzu Plugin Runtime (from Tanzu-Framework repository), Will my plugins be compatible with v0.90 Tanzu CLI?
```
marckhouzam, line 37:
```suggestion
  - If I have developed a plugin with (v0.11, 0.25, 0.28, 0.29) of Tanzu Plugin Runtime (from the Tanzu-Framework repository), will my plugins be compatible with v0.90 Tanzu CLI?
```

docs/full/policy.md
```diff
@@ -15,38 +15,30 @@ Given this scheme, the primary vehicle for changes is minor releases. The deprec
...
+    - Example: If a plugin is developed with Tanzu Plugin Runtime v1.3.1 which introduced a new feature X which works in combination with a newer version of Tanzu CLI then if the user is using old Tanzu CLI v1.2.0 (which doesn’t support this new feature X introduced in Plugin Runtime v1.3.1) while invoking a plugin, that new functionality might not work when using old CLI. However, other functionalities of that plugin will continue to work as expected with the old CLI. This will make the new feature in the plugin incompatible with the installed CLI. However, the user can upgrade the CLI to the latest v1.3.0 version to make the new feature compatible again. This type of situation will always be communicated with the above table when this situation arises.
```
marckhouzam, line 41:
```suggestion
    - Example: If a plugin is developed with Tanzu Plugin Runtime v1.3.1 which introduced a new feature X which works in combination with a newer version of Tanzu CLI then if the user is using an old Tanzu CLI v1.2.0 (which doesn’t support this new feature X introduced in Plugin Runtime v1.3.1) while invoking a plugin, that new functionality might not work when using the old CLI. However, other functionalities of that plugin will continue to work as expected with the old CLI. This will make the new feature in the plugin incompatible with the installed CLI. However, the user can upgrade the CLI to the latest v1.3.0 version to make the new feature compatible again. This type of situation will always be communicated with the above table when this situation arises.
```

marckhouzam (approved):
Small grammar nits
LGTM besides that. Thanks!

PR #355, opened 2023-06-14
Refactor E2E module for external consumption

mpanchajanya:
-  To be able to tag the e2e mod properly we need a definitive version of tanzu cli in e2e go mod
-  After this gets merge. Wil create a tag for test/e2e v1.0.0-dev
- Fix documentation links
- Lint issues in test e2e

Fixes #

E2E tests continue to work as expected

### Release note

```release-note
test/e2e/framework is a separate module that can be imported to write e2e tests for cli
```


test/e2e/go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/onsi/ginkgo/v2 v2.9.2
 	github.com/onsi/gomega v1.27.6
 	github.com/pkg/errors v0.9.1
-	github.com/vmware-tanzu/tanzu-cli v0.0.0-00010101000000-000000000000
+	github.com/vmware-tanzu/tanzu-cli v0.90.0
```
anujc25, line 11: I feel it is better to keep using dummy `v0.0.0-00010101000000-000000000000` as we are not really using `v0.90.0` but the latest copy on the same branch to run the e2e tests because of the `replace` statement mentioned above. Updating to `v0.90.0` might give wrong impression to the user that we are using `v0.90.0` version here.


```
replace github.com/vmware-tanzu/tanzu-cli => ../..
```

mpanchajanya: this no longer required

test/e2e/framework/plugin_operations.go
```diff
@@ -13,9 +13,66 @@ import (
...
+type CLIPlugin struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Spec              CLIPluginSpec `json:"spec"`
+}
+
+//+kubebuilder:object:root=true
+
+// CLIPluginList contains a list of CLIPlugin
+type CLIPluginList struct {
```
marckhouzam, line 70: This one does not seem to be used.

mpanchajanya: thanks for catching this. removed the unused struct

mpanchajanya:
> This makes sense to me. I do have some questions around this change that I may have missed during discussions.
> 
> 1. do we have to keep this new e2e/framework module backwards-compatible
A  I dont think we are entitled here since we have not published the or mentioned to any users its available to use. May be after this change we should keep it backward compatible
> 2. why do we want to avoid depending on the CLI (it feels right, but I can't quite clarify it in my head)
Users already impor tanzu cli and importing e2e that depends on cli would cause conflict and not needed. since if they already has cli they cna use the e2e mod directly then I dont see good reason to sep tag and release e2e.
> 3. I am concerned about copying the v1alpha1 types into `plugin_operations.go`, but when I started thinking about it, I started thinking that the original types need to be kept backwards-compatible since they are related to CRDs, and then I started wondering if we should/could make them public...  this is a more general question we should take up with the team.
I think we are duplicating the type from cliplugin_types.go into plugin_operations.go. until we move this to runtime we need to make sure these copies are sync.

marckhouzam:
@chandrareddyp All the logic in `plugin_operations.go`, do we actually use it?  I don't see anything calling any of it in the repo at the moment? Is it something users of the E2E Framework would need even if this repo does not use it?

PR #356, opened 2023-06-15
Fix plugin sync and plugin list not considering updates available state

anujc25:
- Fix plugin sync and plugin list not considering updates available state

Fixes #

### Release note

```release-note

```


pkg/command/plugin.go
```diff
@@ -669,10 +675,10 @@ func displayInstalledAndMissingSplitView(installedStandalonePlugins []cli.Plugin
 		outputWriter.Render()
 	}
 
-	if len(missingContextPlugins) > 0 {
+	if pluginSyncRequired {
 		// Print a warning to the user that some context plugins are not installed, and how to install them
```
vuil, line 679: nit: update comment as well to match the log message?

pkg/command/plugin.go
```diff
@@ -528,18 +528,24 @@ func getInstalledAndMissingContextPlugins() (installed, missing []discovery.Disc
 
 			// Store the installed plugin, which includes the context from which it was installed
 			found = true
+			if serverPlugins[i].RecommendedVersion != installedPlugins[j].Version {
+				serverPlugins[i].Status = common.PluginStatusUpdateAvailable
+				pluginSyncRequired = true
+			} else {
+				serverPlugins[i].Status = common.PluginStatusInstalled
+			}
 			serverPlugins[i].InstalledVersion = installedPlugins[j].Version
```
anujc25, line 537: When this happens, should we be should `RecommendedVersion` or the `InstalledVersion` of the plugin in the `VERSION` column? cc @vuil @prkalle 

This change continues to show the `InstalledVersion` of the plugin along with the `updates available` status.

vuil: I'd say it is more consistent to show the installed version. 

Nit: (feel free to ignore, as I am not sure it is worth tweaking). 'Updates available' suggests to me there is a newer version of the plugin to update to, but the logic allows for syncing down to an older version too.

vuil (approved):
Thanks! Lgtm, left a few comments, none blocking. 
But let's get some test cases implemented as a follow-up as well.

pkg/command/plugin.go
prkalle: > This change continues to show the InstalledVersion of the plugin along with the updates available status.

This sound good to me

anujc25: I do not have better words to describe that state so may be we can go with the existing `updates available` that we used with the legacy CLI and update it in the future if we thing necessary.

mpanchajanya (commented):
lgtm!. Thanks for making the quick fix

prkalle (approved):
LGTM. Thanks

pkg/command/plugin.go
prkalle: Sounds good to me, but it may be confusing to user and it may prompt user to keep trying plugin sync by seeing `updates available` in the status, may be we should document it.

PR #364, opened 2023-06-20
Update JoinURL function when baseURL has no schema

mpanchajanya:

Fixes #

### Release note

```release-note

```


pkg/utils/url.go
```diff
@@ -20,30 +20,33 @@ import (
 // Returns:
 // A string that is the concatenation of baseURL and relativeURL, formatted correctly.
 // An error if there was a problem parsing the base URL.
-func JoinURL(baseURL, relativeURL string) string {
+func JoinURL(baseURL, relativeURL string) (string, error) {
+	// Check if the URL has a schema
```
anujc25, line 24: nit: This comment should be at Line 29.

pkg/utils/url.go
```diff
@@ -20,30 +20,33 @@ import (
...
-	// 2. scheme://userinfo@host/path?query#fragment
-	// If u.Opaque is non-empty, String uses the first form; otherwise it uses the second form.
+	// Join the baseURL and relativeURL
+	parsedBaseURL = parsedBaseURL.JoinPath(relativeURL)
 
-	if parsedBaseURL.Opaque != "" {
-		return path.Join(parsedBaseURL.String(), parsedBaseURL.Path)
+	if schemaNotPresent {
+		// Replace the schema prefix with the original baseUrl and return the joined URL
+		return strings.ReplaceAll(parsedBaseURL.String(), "https://", ""), nil
```
anujc25, line 47: I feel `strings.TrimPrefix` might be better suitable for this scenario.

```suggestion
		return strings.TrimPrefix(parsedBaseURL.String(), "https://"), nil
```

pkg/utils/url.go
mpanchajanya: updated the comment

marckhouzam (approved):
Nit: the manual test in the PR description would benefit in using `127.0.0.1:9876` since this is the case that was failing (`localhost:9876` was working before, but it's good to show it is still working, as you did).

After addressing @anujc25's comments, this LGTM.

pkg/utils/url_test.go
```diff
@@ -15,108 +15,456 @@ func TestJoinUrl(t *testing.T) {
...
-		{"localhost:5001/tanzu-plugins", "test/path", "localhost:5001/tanzu-plugins/test/path"},
-		{"localhost:5001/tanzu-plugins/", "/test/path", "localhost:5001/tanzu-plugins/test/path"},
-		{"localhost:5001/tanzu-plugins/", "test/path", "localhost:5001/tanzu-plugins/test/path"},
-		{"localhost:5001/tanzu-plugins", "/test/path", "localhost:5001/tanzu-plugins/test/path"},
-		{"localhost:5001/tanzu-plugins", "", "localhost:5001/tanzu-plugins"},
-		{"localhost:5001/tanzu-plugins/", "", "localhost:5001/tanzu-plugins"},
-		{"localhost:5001/tanzu-plugins/", "/", "localhost:5001/tanzu-plugins"},
+		// ip address tests
+
+		{"127.0.0.1:5001/tanzu-plugins/", "/test/path/", "127.0.0.1:5001/tanzu-plugins/test/path/", nil},
```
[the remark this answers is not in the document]
anujc25: I think this seems sufficient based on my understanding.

anujc25 (approved):
LGTM

PR #370, opened 2023-06-23
Remove dependency on imgpkg and docker cli for building plugins

### LINKED ISSUE #372: Remove dependency on imgpkg and docker cli for building plugins
**Bug description**

With the Tanzu CLI v0.90.0-beta.0 release, a new prerequisite was introduced. Here is a note from the release announcement:

```
If you are running `make plugin-build-package` to build plugin OCI images, builder plugin now requires docker to be running to generate plugin oci image tar.gz file.
```

Unfortunately, this change has broken Build Tooling for Integrations and how it builds plugin packages and will require significant changes. Perhaps there is another way to not cause such a setback to the Tanzu Runtime Core team's work with Build Tooling for Integrations.

**Expected behavior**
Remove the dependency of docker to build plugin packages.

**Output of `tanzu version`**

anujc25:

Reasons:
- Image tarball generated using `imgpkg copy -i IMAGE --to-tar TARFILE` is not compatible with the docker load or crane push command.
- This might be because the imgpkg tarball format is a special format that only works with imgpkg copy

Considering the above I have POC'd the following approach as part of building and publishing plugin packages.

Building plugin packages:
- Use the `imgpkg` library to push the OCI image to the localhost repository
- Use `crane pull` as a library to generate and save it as a local tarball

Publishing plugin packages:
- Use `crane push` as a library to push the image to the remote repository

This means the builder plugin does not depend on any external binary to build and publish plugin oci images.

Fixes #372

```
$ make prepare-builder

## Build plugins along with building plugin packages and publishing plugins to the local repository
$ make plugin-build-and-publish-packages

$ make inventory-init
$ make inventory-plugin-add
$ make inventory-plugin-activate
$ make inventory-plugin-deactivate
$ make inventory-plugin-activate
$ make inventory-plugin-group-add
$ make inventory-plugin-group-activate
$ make inventory-plugin-group-deactivate
```

### Release note

```release-note
Remove dependency on imgpkg and docker cli for building plugins
```


pkg/carvelhelpers/image_operations.go
```diff
@@ -109,3 +113,38 @@ func (i *ImageOperationOptions) PushImage(imageWithTag string, filePaths []strin
...
+	}
+	reg, err := newRegistry(registryName)
+	if err != nil {
+		return errors.Wrapf(err, "unable to initialize registry")
+	}
+	return reg.ResolveImage(image)
+}
+
+// GetFileDigestFromImage invokes `DownloadImageAndSaveFilesToDir` to fetch the image and returns the digest of the specified file
+func (i *ImageOperationOptions) GetFileDigestFromImage(image, fileName string) (string, error) {
```
marckhouzam, line 131: to be consistent should the first variable be `imageWithTag`?

cmd/plugin/builder/plugin/publish-packages.go
```diff
@@ -51,26 +51,15 @@ func (ppo *PublishPluginPackageOptions) PublishPluginPackages() error {
...
-				imageAfterLoad := fmt.Sprintf("%s/plugins/%s/%s/%s:%s", localRegistry, osArch.OS(), osArch.Arch(), pluginManifest.Plugins[i].Name, version)
 				imageToPush := fmt.Sprintf("%s/%s/%s/%s/%s/%s/%s:%s", ppo.Repository, ppo.Vendor, ppo.Publisher, osArch.OS(), osArch.Arch(), pluginManifest.Plugins[i].Target, pluginManifest.Plugins[i].Name, version)
-				log.Infof("publishing plugin 'name:%s' 'target:%s' 'os:%s' 'arch:%s' 'version:%s'", pluginManifest.Plugins[i].Name, pluginManifest.Plugins[i].Target, osArch.OS(), osArch.Arch(), version)
 
+				log.Infof("publishing plugin 'name:%s' 'target:%s' 'os:%s' 'arch:%s' 'version:%s'", pluginManifest.Plugins[i].Name, pluginManifest.Plugins[i].Target, osArch.OS(), osArch.Arch(), version)
 				if ppo.DryRun {
-					log.Infof("command: 'docker load -i %s'", pluginTarFilePath)
-					log.Infof("command: 'docker tag %s %s'", imageAfterLoad, imageToPush)
-					log.Infof("command: 'docker push %s'", imageToPush)
+					log.Infof("command: 'crane push -i %s %s'", pluginTarFilePath, imageToPush)
```
marckhouzam, line 58: I think:
```suggestion
					log.Infof("command: 'crane push %s %s'", pluginTarFilePath, imageToPush)
```

marckhouzam (commented):
Looks great.  Thanks @anujc25.

Some very minor comments.

cmd/plugin/builder/plugin/publish-packages.go
marckhouzam: This actually still does not work as a command because we first need to unzip the tar file.  But I've asked about why we do the zip/unzip as another comment first.

marckhouzam (changes requested):
`make inventory-init` is panicking with this PR.
I'm looking to see if I'm doing something wrong...

cmd/plugin/builder/inventory.go
```diff
@@ -47,7 +46,6 @@ func newInventoryInitCmd() *cobra.Command {
 				Repository:        piiFlags.Repository,
 				InventoryImageTag: piiFlags.InventoryImageTag,
 				Override:          piiFlags.Override,
-				ImgpkgOptions:     imgpkg.NewImgpkgCLIWrapper(),
```
marckhouzam, line 50: I think this needs to be replaced with 
`ImageOperationsImpl: carvelhelpers.NewImageOperationsImpl(),`

anujc25: Thanks for catching this.

pkg/registry/client.go
```diff
@@ -287,3 +287,23 @@ func (r *registry) PushImage(imageWithTag string, filePaths []string) error {
...
+	var outputBuf, errorBuf bytes.Buffer
+	writerUI := ui.NewWriterUI(&outputBuf, &errorBuf, nil)
+
+	resolveOptions := cmd.NewTagResolveOptions(writerUI)
+	resolveOptions.ImageFlags = cmd.ImageFlags{Image: image}
+	if r.opts != nil {
+		resolveOptions.RegistryFlags = cmd.RegistryFlags{
+			CACertPaths: r.opts.CACertPaths,
+			VerifyCerts: r.opts.VerifyCerts,
+			Insecure:    r.opts.Insecure,
```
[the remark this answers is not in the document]
anujc25: Thanks for catching this. 👍

anujc25:
> I'm trying to understand why do we zip and unzip the tar file. Is the zipped format used anywhere else?

I have removed the additional zip and unzip-related code from here. If any system expects the plugin oci image to be in the gzip archive format an additional command (e.g. `find ${PLUGIN_PACKAGE_ARTIFACTS_DIR} -name "*.tar" | xargs gzip`) can be used to convert the tar file to tar.gz.

cmd/plugin/builder/plugin/publish-packages.go
anujc25: I have removed the additional gzip archive-related code so the mentioned command should work now.

marckhouzam (approved):
Thanks for addressing the comments @anujc25.
LGTM (you can choose to ignore the two latest nits)

PR #373, opened 2023-06-23
Add timeout to k8s client

chandrareddyp:
This pull request updates the default timeout for the Kubernetes client. Previously, the client had a timeout of 32 seconds. With this pull request, the default timeout has been reduced to 5 seconds. Having a timeout of 32 seconds was not a great user experience, hence the change.

Fixes #

I have tested the PR by mocking the http server and running the `tanzu plugin list` before fix and after the fix:

**Before Fix:**
Default timeout is 32s before the fix: 
We can see that the `tanzu plugin list` command is hanging for 32.12 seconds (32.120 total)
```
❯ git status
On branch main
Your branch is up to date with 'origin/main'.
nothing added to commit but untracked files present (use "git add" to track)
❯ 
❯ make build
build darwin-amd64 CLI with version: v1.0.0-dev
mkdir -p bin
cp /Users/cpamuluri/tkg/tasks/cli_core_main/tanzu-cli/artifacts/darwin/amd64/cli/core/v1.0.0-dev/tanzu-cli-darwin_amd64 ./bin/tanzu
❯ 
❯ time t plugin list
[!] there was an error while discovering plugins, error information: 'unable to list plugins from discovery source 'default-k11': Failed to invoke API on cluster : Get "http://127.0.0.1:58650/version?timeout=32s": net/http: request canceled (Client.Timeout exceeded while awaiting headers)'
[!] there was an error while getting installed context plugins, error information: 'unable to list plugins from discovery source 'default-k11': Failed to invoke API on cluster : Get "http://127.0.0.1:58650/version?timeout=32s": net/http: request canceled (Client.Timeout exceeded while awaiting headers)'
Standalone Plugins
  NAME     DESCRIPTION              TARGET      VERSION  STATUS     
  builder  Build Tanzu components   global      v0.90.0  installed  
  secret   Tanzu secret management  kubernetes  v0.29.0  installed  
[x] : unable to list plugins from discovery source 'default-k11': Failed to invoke API on cluster : Get "http://127.0.0.1:58650/version?timeout=32s": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
./bin/tanzu plugin list  0.10s user 0.04s system 0% cpu 32.120 total
❯
```
**After the Fix:**
As part of the fix the default timeout has updated as 5 seconds, we can see the same in the below request call "https://127.0.0.1:58650/version?timeout=5s" and the `tanzu plugin list` is ended after 6.040 seconds 
```
❯ 
❯ git checkout topic/chandra/add-timeout-k8s-client
Switched to branch 'topic/chandra/add-timeout-k8s-client'
❯ 
❯ make build
build darwin-amd64 CLI with version: v1.0.0-dev
mkdir -p bin
cp /Users/cpamuluri/tkg/tasks/cli_core_main/tanzu-cli/artifacts/darwin/amd64/cli/core/v1.0.0-dev/tanzu-cli-darwin_amd64 ./bin/tanzu
❯ 
❯ 
❯ time t plugin list                               
[!] there was an error while discovering plugins, error information: 'unable to list plugins from discovery source 'default-k11': Failed to invoke API on cluster : Get "http://127.0.0.1:58650/version?timeout=5s": net/http: request canceled (Client.Timeout exceeded while awaiting headers)'
[!] there was an error while getting installed context plugins, error information: 'unable to list plugins from discovery source 'default-k11': Failed to invoke API on cluster : Get "http://127.0.0.1:58650/version?timeout=5s": net/http: request canceled (Client.Timeout exceeded while awaiting headers)'
Standalone Plugins
  NAME     DESCRIPTION              TARGET      VERSION  STATUS     
  builder  Build Tanzu components   global      v0.90.0  installed  
  secret   Tanzu secret management  kubernetes  v0.29.0  installed  
[x] : unable to list plugins from discovery source 'default-k11': Failed to invoke API on cluster : Get "http://127.0.0.1:58650/version?timeout=5s": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
./bin/tanzu plugin list  0.09s user 0.04s system 2% cpu 6.040 total
❯ 
```
### Release note

```release-note

```

### DIFF
--- a/pkg/cluster/client.go
+++ b/pkg/cluster/client.go
@@ -6,6 +6,7 @@ package cluster
 
 import (
 	"context"
+	"time"
 
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v3"
@@ -107,8 +108,7 @@ func NewClient(kubeConfigPath, contextStr string, options Options) (Client, erro
 		kubeConfigPath: kubeConfigPath,
 		currentContext: contextStr,
 	}
-
-	err = client.getK8sClients(options.CrtClient, options.DiscoveryClientFactory, options.DynamicClientFactory)
+	err = client.getK8sClients(options.CrtClient, options.DiscoveryClientFactory, options.DynamicClientFactory, options.RequestTimeout)
 	if err != nil {
 		return nil, err
 	}
@@ -199,7 +199,7 @@ func ConsolidateImageRepoMaps(cmList *corev1.ConfigMapList) (map[string]string, 
 	return imageRepoMap, nil
 }
 
-func (c *client) getK8sClients(crtClient CrtClient, discoveryClientFactory DiscoveryClientFactory, dynamicClientFactory DynamicClientFactory) error {
+func (c *client) getK8sClients(crtClient CrtClient, discoveryClientFactory DiscoveryClientFactory, dynamicClientFactory DynamicClientFactory, timeout time.Duration) error {
 	var discoveryClient discovery.DiscoveryInterface
 	config, err := clientcmd.LoadFromFile(c.kubeConfigPath)
 	if err != nil {
@@ -218,6 +218,10 @@ func (c *client) getK8sClients(crtClient CrtClient, discoveryClientFactory Disco
 	// queries per second and the maximum burst for throttle to a high value to avoid throttling of messages
 	restConfig.QPS = constants.DefaultQPS
 	restConfig.Burst = constants.DefaultBurst
+	if timeout != 0 {
+		restConfig.Timeout = timeout
+	}
+
 	mapper, err := apiutil.NewDynamicRESTMapper(restConfig, apiutil.WithLazyDiscovery)
 	if err != nil {
 		return errors.Errorf("Unable to set up rest mapper due to : %v", err)
@@ -292,6 +296,7 @@ type Options struct {
 	CrtClient              CrtClient
 	DiscoveryClientFactory DiscoveryClientFactory
 	DynamicClientFactory   DynamicClientFactory
+	RequestTimeout         time.Duration
 }
 
 // NewOptions returns new options
--- a/pkg/discovery/interface.go
+++ b/pkg/discovery/interface.go
@@ -10,10 +10,13 @@ package discovery
 
 import (
 	"errors"
+	"time"
 
 	"github.com/vmware-tanzu/tanzu-cli/pkg/plugininventory"
 	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 )
+
+var defaultTimeout = 5 * time.Second
 
 // Discovery is the interface to fetch the list of available plugins
 type Discovery interface {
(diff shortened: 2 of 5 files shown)


marckhouzam:
@chandrareddyp This looks reasonable.  Have you been able to reproduce the original delay to see that it is now shorter?

chandrareddyp:
> @chandrareddyp This looks reasonable. Have you been able to reproduce the original delay to see that it is now shorter?
@marckhouzam i did not reproduce the origin 30 seconds hang use case, but i reproduced the timeout use case by setting 1 milliseconds as timeout. I have added the same in the PR testing section.

pkg/cluster/client.go
```diff
@@ -292,6 +299,7 @@ type Options struct {
 	CrtClient              CrtClient
 	DiscoveryClientFactory DiscoveryClientFactory
 	DynamicClientFactory   DynamicClientFactory
+	TimeOut                time.Duration
```
anujc25, line 302: Might be better to use something like `RequestTimeout` so that users understand what the Timeout means here. I would also suggest adding a comment to explain the same.
```suggestion
	RequestTimeout                time.Duration
```

anujc25 (commented):
Overall looks good. But it would have been nice to reproduce the actual issue and make sure that this fix works as expected.

chandrareddyp:
> Overall looks good. But it would have been nice to reproduce the actual issue and make sure that this fix works as expected.

Thanks for the advice, @marckhouzam and @anujc25. I have successfully reproduced the issue by mocking the HTTP server, and I have updated the testing section accordingly.

anujc25 (approved):
LGTM. Thanks.

pkg/cluster/client.go
```diff
@@ -107,8 +110,11 @@ func NewClient(kubeConfigPath, contextStr string, options Options) (Client, erro
 		kubeConfigPath: kubeConfigPath,
 		currentContext: contextStr,
 	}
+	if options.RequestTimeout == 0 {
+		options.RequestTimeout = defaultTimeout
+	}
```
marckhouzam, line 115: Just asking: do we really prefer to do this?  The documentation of the field `restConfig.Timeout` used at line 227 of this file says that "a value of 0 means no timeout".  Why not leave that as a possibility in our own API? 

(If you decide to remove this, you no longer need a `defaultTimeout` constant above)

marckhouzam (approved):
One nit but LGTM thanks!

pkg/cluster/client.go
```diff
@@ -199,7 +205,7 @@ func ConsolidateImageRepoMaps(cmList *corev1.ConfigMapList) (map[string]string,
 	return imageRepoMap, nil
 }
 
-func (c *client) getK8sClients(crtClient CrtClient, discoveryClientFactory DiscoveryClientFactory, dynamicClientFactory DynamicClientFactory) error {
+func (c *client) getK8sClients(crtClient CrtClient, discoveryClientFactory DiscoveryClientFactory, dynamicClientFactory DynamicClientFactory, timeOut time.Duration) error {
```
marckhouzam, line 208: Very small nit: `timeOut` -> `timeout` to be consistent with everywhere else.

pkg/cluster/client.go
chandrareddyp: Having no timeout is not great user experience, I would avoid doing `restConfig.Timeout = timeout` at line 227 if input timeout is 0, so that we will consider it default timeout which is 32 seconds

marckhouzam (approved):
LGTM

PR #377, opened 2023-06-28
Run OCI image package building and publishing concurrently

### LINKED ISSUE #378: Build and Publish plugins OCI images in parallel to improve the speed

As part of the builder plugin we are generating an OCI image package for the plugins and using that package to publish OCI images to the destination container registry.

We can us go routines to do things in parallel to improve the speed.

**Additional context**

anujc25:

- As part of the builder plugin we are generating an OCI image package for the plugins and using that package to publish OCI images to the destination container registry.
- Before this change, the building and publishing of the OCI image were happening sequentially. This change uses go concurrency to run multiple things using go routines to speed up the building and publishing process.

**Performance comparison:**
- The OCI image build to build 7 plugins for 3 os-arch took `40.27s` compared to `164.53s` earlier which is approx **4** times faster.
- The OCI image publishing to publish 7 plugins for 3 os-arch took `21.64s` compared to `195.62s` earlier which is approx **9** times faster.
- The OCI image build and publishing combined of 7 plugins for 3 os-arch took `107.22s` compared to `360.87s` earlier which is approx **3.5** times faster.

Fixes #378

```
Building Plugin Packages: (make plugin-build-packages)

   Without Parallel Execution: 164.53 secs
   With Parallel Execution: 40.27 secs

Publishing Plugin Packages: (make plugin-plugin-packages) (excluding the building packages)
    
   Without Parallel Execution: 195.62 secs
   With Parallel Execution: 21.64 secs

Building & Publishing Plugin Packages Combined: (make plugin-plugin-packages) 

   Without Parallel Execution: 360.87 secs
   With Parallel Execution: 107.22 secs

```

### Release note

```release-note
Run OCI image package building and publishing concurrently when using `builder` plugin
```

### DIFF
--- a/cmd/plugin/builder/helpers/helper.go
+++ b/cmd/plugin/builder/helpers/helper.go
@@ -10,7 +10,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"runtime"
 
 	"github.com/pkg/errors"
 
@@ -20,8 +19,6 @@ import (
 	"github.com/vmware-tanzu/tanzu-cli/pkg/utils"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
 )
-
-var minConcurrent = 2
 
 // ReadPluginManifest reads the PluginManifest file and returns Manifest object
 func ReadPluginManifest(pluginManifestFile string) (*cli.Manifest, error) {
@@ -77,8 +74,8 @@ func GetDigest(filePath string) (string, error) {
 // ValidatePluginBinary validates the plugin binary file
 // Checks the binary file size if its empty
 // returns true if file is not empty and false if file is empty
-func ValidatePluginBinary(pluginBinaryFilePath string) (bool, error) {
-	log.Infof("Validating Plugin Binary file: %v", pluginBinaryFilePath)
+func ValidatePluginBinary(pluginBinaryFilePath, threadID string) (bool, error) {
+	log.Infof("%s Validating Plugin Binary file: %v", threadID, pluginBinaryFilePath)
 	// Check whether the plugin binary is empty by verifying if size is zero
 	fileEmpty, err := utils.IsFileEmpty(pluginBinaryFilePath)
 
@@ -87,36 +84,4 @@ func ValidatePluginBinary(pluginBinaryFilePath string) (bool, error) {
 	}
 
 	return !fileEmpty, nil
-}
-
-func GetMaxParallelism() int {
-	maxConcurrent := runtime.NumCPU() - 2
-	if maxConcurrent < minConcurrent {
-		maxConcurrent = minConcurrent
-	}
-	return maxConcurrent
-}
-
-var Identifiers = []string{
-	string("=> T01]"),
-	string("=> T02]"),
-	string("=> T03]"),
-	string("=> T04]"),
-	string("=> T05]"),
-	string("=> T06]"),
-	string("=> T07]"),
-	string("=> T08]"),
-	string("=> T09]"),
-	string("=> T10]"),
-	string("=> T11]"),
-	string("=> T12]"),
-}
-
-func GetID(i int) string {
-	index := i
-	if i >= len(Identifiers) {
-		// Well aren't you lucky
-		index = i % len(Identifiers)
-	}
-	return Identifiers[index]
 }
--- a/cmd/plugin/builder/helpers/helper.go
+++ b/cmd/plugin/builder/helpers/helper.go
@@ -85,3 +85,15 @@ func ValidatePluginBinary(pluginBinaryFilePath, threadID string) (bool, error) {
 
 	return !fileEmpty, nil
 }
+
+// GetNumberOfIndividualPluginBinariesFromManifest returns the number of plugin binaries
+// into consideration based on the plugin manifest file
+// This includes plugin binaries for all supported os architectures as well as
+// all versions mentioned for the specific plugin.
+func GetNumberOfIndividualPluginBinariesFromManifest(pluginManifest *cli.Manifest) int {
+	numberOfPluginPackages := 0
+	for i := range pluginManifest.Plugins {
+		numberOfPluginPackages += len(pluginManifest.Plugins[i].Versions)
+	}
+	return numberOfPluginPackages * len(cli.AllOSArch)
+}
--- a/cmd/plugin/builder/helpers/helper.go
+++ b/cmd/plugin/builder/helpers/helper.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"github.com/pkg/errors"
 
@@ -19,6 +20,8 @@ import (
 	"github.com/vmware-tanzu/tanzu-cli/pkg/utils"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
 )
+
+var minConcurrent = 2
 
 // ReadPluginManifest reads the PluginManifest file and returns Manifest object
 func ReadPluginManifest(pluginManifestFile string) (*cli.Manifest, error) {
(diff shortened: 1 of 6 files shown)


marckhouzam:
I was hoping to try this out including the change from #370 but there is a merge conflict.
@anujc25 will you be able to rebase this PR once #370 is merged?

cmd/plugin/builder/plugin/publish-packages.go
```diff
@@ -43,40 +46,67 @@ func (ppo *PublishPluginPackageOptions) PublishPluginPackages() error {
 
 	log.Infof("using plugin package artifacts from %q", ppo.PackageArtifactDir)
 
+	// Limit the number of concurrent operations we perform so we don't overwhelm the system.
+	maxConcurrent := helpers.GetMaxParallelism()
+	guard := make(chan struct{}, maxConcurrent)
+	var wg sync.WaitGroup
+	// Mix up IDs so we don't always get the same set.
+	randSkew := rand.Intn(len(helpers.Identifiers)) // nolint:gosec
+	errorList := []error{}
```
marckhouzam, line 55: Same comment about an array probably not being thread-safe here

cmd/plugin/builder/plugin/publish-packages.go
```diff
@@ -43,40 +46,67 @@ func (ppo *PublishPluginPackageOptions) PublishPluginPackages() error {
...
+
+			err := ppo.DockerOptions.LoadImage(pluginTarFilePath)
+			if err != nil {
+				errorList = append(errorList, errors.Wrapf(err, "unable to load image for plugin: %s, target: %s, os: %s, arch: %s, version: %s", p.Name, p.Target, osArch.OS(), osArch.Arch(), version))
+				return
+			}
+			err = ppo.DockerOptions.TagImage(imageAfterLoad, imageToPush)
+			if err != nil {
+				errorList = append(errorList, errors.Wrapf(err, "unable to retag image for plugin: %s, target: %s, os: %s, arch: %s, version: %s", p.Name, p.Target, osArch.OS(), osArch.Arch(), version))
+			}
```
marckhouzam, line 86: Missing `return`

cmd/plugin/builder/plugin/build-packages.go
```diff
@@ -50,50 +53,80 @@ func (bpo *BuildPluginPackageOptions) BuildPluginPackages() error {
...
+
+		if !utils.PathExists(pluginBinaryFilePath) {
+			return
+		}
 
-				if !utils.PathExists(pluginBinaryFilePath) {
-					continue
-				}
+		// Check whether the binary exec file is valid
+		valid, err := helpers.ValidatePluginBinary(pluginBinaryFilePath)
```
marckhouzam, line 78: This function call will be the first output shown to the user(`Validating Plugin Binary file: <path>`.
This output does not have the special thread ID, but it does show the path of the plugin binary, so the user can figure things out.

I suggest that we add a printout at the very beginning of each execution of the new `generatePluginPackage()` to indicate we are starting a new thread.

As for `helpers.ValidatePluginBinary()` I leave it up to you to decide if you want to include the threadId in the output or not.  Or even remove the printout completely (it is not that important).  Actually, I'm thinking removing it is maybe better?

cmd/plugin/builder/helpers/helper.go
```diff
@@ -85,3 +88,35 @@ func ValidatePluginBinary(pluginBinaryFilePath string) (bool, error) {
...
+	string("=> T03]"),
+	string("=> T04]"),
+	string("=> T05]"),
+	string("=> T06]"),
+	string("=> T07]"),
+	string("=> T08]"),
+	string("=> T09]"),
+	string("=> T10]"),
+	string("=> T11]"),
+	string("=> T12]"),
```
marckhouzam, line 112: Why did we choose to use a different ID than for compilation?
Is it to get rid of non-ascii symbols?

cmd/plugin/builder/plugin/build-packages.go
```diff
@@ -50,50 +53,80 @@ func (bpo *BuildPluginPackageOptions) BuildPluginPackages() error {
...
-				pluginBinaryFilePath := filepath.Join(bpo.BinaryArtifactDir, osArch.OS(), osArch.Arch(),
-					pluginManifest.Plugins[i].Target, pluginManifest.Plugins[i].Name, version,
-					cli.MakeArtifactName(pluginManifest.Plugins[i].Name, osArch))
+	// Limit the number of concurrent operations we perform so we don't overwhelm the system.
+	maxConcurrent := helpers.GetMaxParallelism()
+	guard := make(chan struct{}, maxConcurrent)
+	var wg sync.WaitGroup
+	// Mix up IDs so we don't always get the same set.
+	randSkew := rand.Intn(len(helpers.Identifiers)) // nolint:gosec
+	errorList := []error{}
```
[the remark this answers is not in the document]
anujc25: Good catch. I will update it to use channels. I didn't pay attention to these details. 👍

cmd/plugin/builder/plugin/build-packages.go
anujc25: Good point. Updated to pass threadID to the `ValidatePluginBinary` to make logs consistent.

cmd/plugin/builder/helpers/helper.go
anujc25: That was the initial thinking but I am updated to use the existing symbols instead of updating them in this PR. If needed we can update it in a separate PR.

marckhouzam (commented):
Once last tweak about the concurrency and we're good.

cmd/plugin/builder/plugin/build-packages.go
```diff
@@ -47,56 +48,95 @@ func (bpo *BuildPluginPackageOptions) BuildPluginPackages() error {
 
 	log.Infof("Using plugin binary artifacts from %q", bpo.BinaryArtifactDir)
 
+	// Limit the number of concurrent operations we perform so we don't overwhelm the system.
+	maxConcurrent := helpers.GetMaxParallelism()
+	guard := make(chan struct{}, maxConcurrent)
+	var wg sync.WaitGroup
+	fatalErrors := make(chan helpers.ErrInfo, len(pluginManifest.Plugins)*len(cli.AllOSArch))
```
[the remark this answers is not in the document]
chandrareddyp: @anujc25 Have you considered using a slice instead of a channel? I would like to know if there are compelling reasons to choose a channel over a slice.

cmd/plugin/builder/helpers/parallel.go
```diff
@@ -0,0 +1,51 @@
...
+type ErrInfo struct {
+	Err  error
+	Path string
+	ID   string
+}
+
+// GetMaxParallelism return the maximum concurrent threads to use.
+// Limit the number of concurrent operations we perform so we don't overwhelm the system.
+func GetMaxParallelism() int {
+	maxConcurrent := runtime.NumCPU() - 2
```
chandrareddyp, line 20: any specific reason for reducing 2 from the number of cpu count?

cmd/plugin/builder/plugin/build-packages.go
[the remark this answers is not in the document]
anujc25: Thanks @marckhouzam for the details.

> It is a rare case to have more than one version, so I was wondering what was the risk if it actually happened AND there were more errors that we had allocated.

Considering, when we do `make plugin-build` the plugin manifest generated with that will always only have 1 version in it and there is no way to build the plugin with the multiple version in a single run. So it is indeed a rare case where the user updates the plugin manifest manually to have multiple versions.

cmd/plugin/builder/helpers/parallel.go
anujc25: As mentioned in the comment above, this is to make sure we do not overwhelm the system by using all CPUs to build plugins and leave a few CPUs available to do other operations. This is a prior art where we build plugins in parallel and I am using the same functions to build plugin packages and publishing those in parallel.

marckhouzam (approved):
LGTM
Thanks for addressing all the comments @anujc25 !

PR #386, opened 2023-06-29
Add building APT packages on a Fedora image

marckhouzam:

When building the apt packages we assumed an ubuntu-like image.
It is more flexible to also support a redhat-like image even for apt packages.

This commit makes the apt build script generic so they can use either ubuntu or fedora.

Fixes # N/A

Ran `make apt-package` which uses ubuntu and verified locally that the package built and could be installed.

Ran `make apt-package APT_BUILDER_IMAGE=fedora` which instead uses a fedora image and verified locally that the package built and could be installed.

### Release note

```release-note
The APT distribution package can now be built using a Fedora image.
```

### DIFF
--- a/hack/apt/README.md
+++ b/hack/apt/README.md
@@ -34,6 +34,11 @@ repository will be generated as a directory called `_output/apt` which will
 contain the two built packages. Please see the section on publishing the
 repository for more details.
 
+## Using fedora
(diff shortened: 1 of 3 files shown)


vuil (approved):
lgtm, thanks.

PR #391, opened 2023-07-05
Add command invocation metrics collection for telemetry

prkalle:
This PR adds command invocation metrics collection for telemetry

Some Key points:
Flags:
- For all the core commands except `plugin` LCM(life cycle management) commands, all the flag values (except boolean flags) should be masked 
- For any command implemented in any plugin, only the flag names would be captured

Arguments:
- For all the core commands, the default argument(arg[0]) would be hashed and stored in `name_arg` in the DB
- For any command implemented in any plugin, the default arguments would not be capture due to the limitation in the existing infrastructure to capture the plugin args reliably. 

Command:
- For core command, the complete command chain(excluding the `tanzu`) would be captured
   eg: For ` tanzu config cert add --host test.host.com --insecure`, the command captured is `config cert add` 
- For any command implemented in any plugin, currently it would only capture the plugin name due to the way CLI invokes the plugins. This would be updated in the follow-up PR. 

Fixes #

**Testing core commands:** 
- Ran the tanzu core command `tanzu config get` and the metrics is collected successfully in the DB
```
❯ ./bin/tanzu config get
servers:
    - name: tkg-aws-cc-capi115-upg-mc
      type: managementcluster
      managementClusterOpts:
        path: /Users/pkalle/.kube/config
        context: tkg-aws-cc-capi115-upg-mc-admin@tkg-aws-cc-capi115-upg-mc
<SNAPPED>


SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM TanzuCLIOperations;
v1.0.0-dev|darwin|amd64|||config get|aa4c97d5-408e-4cbb-a5f0-986ca5ae79f9|1688581470984|1688581471015|||||||0|1|
```

- Ran the tanzu core command `tanzu config set env.` command. Metrics is collected successfully in the DB and default arg is hashed
```
❯ rm -rf /Users/pkalle/.config/tanzu-cli-telemetry/cli_metrics.db
❯ ./bin/tanzu config set env.TANZU_CLI_ADDITIONAL_PLUGIN_DISCOVERY_IMAGES_TEST_ONLY localhost:9876/tanzu-cli/plugins/sandbox1:small,localhost:9876/tanzu-cli/plugins/sandbox2:small
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM TanzuCLIOperations;
v1.0.0-dev|darwin|amd64|||config set|aa4c97d5-408e-4cbb-a5f0-986ca5ae79f9|1688581603251|1688581603294||||6b3ec009eeaada86a7a9da005dfa3a13f2af2f1157bb73553c819d8503a18953|||0|1|
```

- Running the tanzu core command with flags , the flag values should be masked (except for `plugin` LCM commands all the core command flag values will be asked other than boolan flags)

```
❯ rm -rf /Users/pkalle/.config/tanzu-cli-telemetry/cli_metrics.db
❯
❯ ./bin/tanzu config cert add --host test.host.com --ca-certificate ~/openssl/server.crt
[ok] successfully added certificate data for host test.host.com

SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM TanzuCLIOperations;
v1.0.0-dev|darwin|amd64|||config cert add|aa4c97d5-408e-4cbb-a5f0-986ca5ae79f9|1688583372085|1688583372107||||||{"ca-certificate":"ea4ca78badbf06f493b4710e62171b6796e0f1b7d7092bfa558ea02a849c76ae","host":"d1c5ef9a0984e2877edec4d798450dc2033c68a567fef3761e71167f073f08a7"}|0|1|
```

- Running the plugin LCM command, all the flag values should be in clear text(not hashed)

```
❯ rm -rf /Users/pkalle/.config/tanzu-cli-telemetry/cli_metrics.db
❯ ./bin/tanzu plugin group search --name vmware-tkg/default --show-details
name: vmware-tkg/default
description: Desc for vmware-tkg/default:v22.22.22
latest: v22.22.22
versions:
    - v2.2.0
    - v9.9.9
    - v11.11.11
    - v22.22.22
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM TanzuCLIOperations;
v1.0.0-dev|darwin|amd64|||plugin group search|aa4c97d5-408e-4cbb-a5f0-986ca5ae79f9|1688583670123|1688583670828||||||{"name":"vmware-tkg/default","show-details":"true"}|0|1|


❯ # Installing the plugin group, you can notice the "group" flag value is captured
❯ rm -rf /Users/pkalle/.config/tanzu-cli-telemetry/cli_metrics.db

❯ ./bin/tanzu plugin install --group vmware-tkg/default:v2.2.0
[i] Installing plugin 'isolated-cluster:v0.29.0' with target 'global'
[i] Installing plugin 'management-cluster:v0.29.0' with target 'kubernetes'
[i] Installing plugin 'package:v0.29.0' with target 'kubernetes'
[i] Installing plugin 'pinniped-auth:v0.29.0' with target 'global'
[i] Installing plugin 'secret:v0.29.0' with target 'kubernetes'
[i] Installing plugin 'telemetry:v0.29.0' with target 'kubernetes'
[ok] successfully installed all plugins from group 'vmware-tkg/default:v2.2.0'
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM TanzuCLIOperations;
v1.0.0-dev|darwin|amd64|||plugin install|aa4c97d5-408e-4cbb-a5f0-986ca5ae79f9|1688583805560|1688583850128||||||{"group":"vmware-tkg/default:v2.2.0"}|0|1|
```

- Testing the core command (context plugin)

```
❯ ./bin/tanzu context create tkg-mgmt-vc --kubeconfig ~/temp/tkgCluster_admin.kfg --kubecontext tkg-mgmt-vc-admin@tkg-mgmt-vc
[ok] successfully created a kubernetes context using the kubeconfig /Users/pkalle/temp/tkgCluster_admin.kfg
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM TanzuCLIOperations;
v1.0.0-dev|darwin|amd64|||context create|7dbe3b36-f07b-4206-bd77-0f7728e75cc2|1688586232461|1688586233335||||86fde1e81453c0fa397ddee030f932e2e6e4f2a06fc49e53b23bda225860b3a8||{"kubeconfig":"27c1c0f843b185e6faee830432c1e6994f83d2d252d2fa090f16a9fa6c3adfa4","kubecontext":"b1bbb64960ad9c73469b1654e7992846c02cfdc5ddbd4d3b354bbfae54452fd6"}|0|1|
```

- Testing negative case for core command. Metric is captured successfully(including the exit status )
```
❯ ./bin/tanzu context delete invalid-context -y
[i] Deleting entry for cluster invalid-context
[x] : context invalid-context not found
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM TanzuCLIOperations;
v1.0.0-dev|darwin|amd64|||context delete|7dbe3b36-f07b-4206-bd77-0f7728e75cc2|1688587343050|1688587343063||||5e12a27eb2297df878e8124d622837b703e96cb2385d9471a5f878e6bea0c574||{"yes":"true"}|1|1|

```

**Testing the Plugins:**
- Ran `tanzu management-cluster get ` and you can notice the plugin name, plugin verison, target, endpoint(hash of the context), flags  are captured. 
Note:  command name would only show plugin name, to capture the complete command chain, the work is in progress

```
❯ ./bin/tanzu mc get --show-details
  NAME         NAMESPACE   STATUS   CONTROLPLANE  WORKERS  KUBERNETES        ROLES       PLAN  TKR
  tkg-mgmt-vc  tkg-system  running  1/1           1/1      v1.24.9+vmware.1  management  dev   v1.24.9---vmware.1-tkg.1


Details:

NAME                                                                                READY  SEVERITY  REASON  SINCE  MESSAGE
/tkg-mgmt-vc                                                                        True                     40m
├─ClusterInfrastructure - VSphereCluster/tkg-mgmt-vc-9cc9g                          True                     40m
├─ControlPlane - KubeadmControlPlane/tkg-mgmt-vc-lhr9q                              True                     40m
│ └─Machine/tkg-mgmt-vc-lhr9q-4rt4g                                                 True                     40m
│   ├─BootstrapConfig - KubeadmConfig/tkg-mgmt-vc-lhr9q-6cpnt                       True                     40m
│   └─MachineInfrastructure - VSphereMachine/tkg-mgmt-vc-control-plane-bj7fv-kpsjm  True                     40m
└─Workers
  └─MachineDeployment/tkg-mgmt-vc-md-0-nv25l                                        True                     40m
    └─Machine/tkg-mgmt-vc-md-0-nv25l-69896b7486-642qn                               True                     40m
      ├─BootstrapConfig - KubeadmConfig/tkg-mgmt-vc-md-0-bootstrap-xnmtq-5mzdt      True                     40m
      └─MachineInfrastructure - VSphereMachine/tkg-mgmt-vc-md-0-infra-f9tkd-kpgwm   True                     40m


Providers:

  NAMESPACE                          NAME                            TYPE                    PROVIDERNAME     VERSION  WATCHNAMESPACE
  caip-in-cluster-system             infrastructure-ipam-in-cluster  InfrastructureProvider  ipam-in-cluster  v0.1.0
  capi-kubeadm-bootstrap-system      bootstrap-kubeadm               BootstrapProvider       kubeadm          v1.2.8
  capi-kubeadm-control-plane-system  control-plane-kubeadm           ControlPlaneProvider    kubeadm          v1.2.8
  capi-system                        cluster-api                     CoreProvider            cluster-api      v1.2.8
  capv-system                        infrastructure-vsphere          InfrastructureProvider  vsphere          v1.5.1

❯ 
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM TanzuCLIOperations;
v1.0.0-dev|darwin|amd64|management-cluster|v0.28.0|management-cluster|7dbe3b36-f07b-4206-bd77-0f7728e75cc2|1688586542348|1688586544952|||kubernetes||58fabb359b0e763b10e0ca376f00cf89951251e8b8772601ae3cdff14a14e7f6|{"show-details":""}|0|1|
```

- Running the plugin with target specified before the plugin name (specified `k8s` before `mc` in the cli command). Metrics collection was successful
```
❯ ./bin/tanzu k8s mc get --show-details
  NAME         NAMESPACE   STATUS   CONTROLPLANE  WORKERS  KUBERNETES        ROLES       PLAN  TKR
  tkg-mgmt-vc  tkg-system  running  1/1           1/1      v1.24.9+vmware.1  management  dev   v1.24.9---vmware.1-tkg.1


Details:

NAME                                                                                READY  SEVERITY  REASON  SINCE  MESSAGE
/tkg-mgmt-vc                                                                        True                     138m
├─ClusterInfrastructure - VSphereCluster/tkg-mgmt-vc-rhj2j                          True                     138m
├─ControlPlane - KubeadmControlPlane/tkg-mgmt-vc-2nnbx                              True                     138m
│ └─Machine/tkg-mgmt-vc-2nnbx-c58ks                                                 True                     138m
│   ├─BootstrapConfig - KubeadmConfig/tkg-mgmt-vc-2nnbx-9j5fk                       True                     138m
│   └─MachineInfrastructure - VSphereMachine/tkg-mgmt-vc-control-plane-5zhgz-s9wxx  True                     138m
└─Workers
  └─MachineDeployment/tkg-mgmt-vc-md-0-mx474                                        True                     138m
    └─Machine/tkg-mgmt-vc-md-0-mx474-7b85596464-7mg7v                               True                     138m
      ├─BootstrapConfig - KubeadmConfig/tkg-mgmt-vc-md-0-bootstrap-xp24l-lpfmf      True                     138m
      └─MachineInfrastructure - VSphereMachine/tkg-mgmt-vc-md-0-infra-8r8xh-jlsml   True                     138m


Providers:

  NAMESPACE                          NAME                            TYPE                    PROVIDERNAME     VERSION  WATCHNAMESPACE
  caip-in-cluster-system             infrastructure-ipam-in-cluster  InfrastructureProvider  ipam-in-cluster  v0.1.0
  capi-kubeadm-bootstrap-system      bootstrap-kubeadm               BootstrapProvider       kubeadm          v1.2.8
  capi-kubeadm-control-plane-system  control-plane-kubeadm           ControlPlaneProvider    kubeadm          v1.2.8
  capi-system                        cluster-api                     CoreProvider            cluster-api      v1.2.8
  capv-system                        infrastructure-vsphere          InfrastructureProvider  vsphere          v1.5.1
❯
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM TanzuCLIOperations;
v1.0.0-dev|darwin|amd64|management-cluster|v0.28.0|kubernetes management-cluster|7dbe3b36-f07b-4206-bd77-0f7728e75cc2|1688681354197|1688681356139|||kubernetes||58fabb359b0e763b10e0ca376f00cf89951251e8b8772601ae3cdff14a14e7f6|{"show-details":""}|0|1|
```

- Negative test case where the `tanzu cluster create` command failed. The metric is collected successfully including the exit status,flags etc..
```
❯ ./bin/tanzu cluster create testCluster --tkr v1.24.9---vmware.1-tkg.1 -d
Error: required config variable 'CLUSTER_PLAN' not set
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM TanzuCLIOperations;
v1.0.0-dev|darwin|amd64|cluster|v0.28.0|cluster|7dbe3b36-f07b-4206-bd77-0f7728e75cc2|1688587083310|1688587084109|||kubernetes||58fabb359b0e763b10e0ca376f00cf89951251e8b8772601ae3cdff14a14e7f6|{"d":"","tkr":""}|1|1|
```

### Release note

```release-note
Add command invocation metrics collection for telemetry
```

### Additional information

TODO: 
1. For plugins, a parser need to be implemented to identify the complete command chain executed by the user. This is due to the fact that cobra can only recognize the command path upto plugin name due to the plugin architecture, all the subcommand,args and flags after the plugin name are identified as args. So we need special logic to identify the command chain/path executed by the user. [ Done as part of PR: #398 ]
2. Update the metrics with cspOrgID and account number configured by user using the telemetry plugin
3. Configure and update the metrics with `isInternal` flag to determine whether CLI build is production/official release or customer build.

### DIFF
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -5,11 +5,13 @@
 package command
 
 import (
-	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
@@ -22,10 +24,11 @@ import (
 	cliconfig "github.com/vmware-tanzu/tanzu-cli/pkg/config"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/constants"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/pluginsupplier"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/telemetry"
 )
 
 // NewRootCmd creates a root command.
-func NewRootCmd() (*cobra.Command, error) {
+func NewRootCmd() (*cobra.Command, error) { //nolint:gocyclo
 	var rootCmd = newRootCmd()
 	uFunc := cli.NewMainUsage().UsageFunc()
 	rootCmd.SetUsageFunc(uFunc)
@@ -45,7 +48,9 @@ func NewRootCmd() (*cobra.Command, error) {
 		//       If we decide to fold this functionality into existing 'tanzu telemetry' plugin
 		newCEIPParticipationCmd(),
 	)
-
+	if _, err := ensureCLIInstanceID(); err != nil {
+		return nil, errors.Wrap(err, "failed to ensure CLI ID")
+	}
 	// If the context and target feature is enabled, add the corresponding commands under root.
 	if config.IsFeatureActivated(constants.FeatureContextCommand) {
 		rootCmd.AddCommand(
@@ -66,6 +71,7 @@ func NewRootCmd() (*cobra.Command, error) {
 	if err != nil {
 		return nil, err
 	}
+	telemetry.Client().SetInstalledPlugins(plugins)
 	if err = config.CopyLegacyConfigDir(); err != nil {
 		return nil, fmt.Errorf("failed to copy legacy configuration directory to new location: %w", err)
 	}
@@ -131,6 +137,10 @@ func newRootCmd() *cobra.Command {
 		// Flag parsing must be deactivated because the root plugin won't know about all flags.
 		DisableFlagParsing: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if err := telemetry.Client().UpdateCmdPreRunMetrics(cmd, args); err != nil {
+				telemetry.LogError(err, "")
+			}
+
 			// Prompt user for EULA and CEIP agreement if necessary, except for
 			skipCommands := []string{
 				// The shell completion setup is not interactive, so it should not trigger a prompt
@@ -259,6 +269,18 @@ func duplicateAliasWarning(rootCmd *cobra.Command) {
 		}
 	}
 }
+func ensureCLIInstanceID() (string, error) {
+	cliID, _ := config.GetCLIId()
+	if cliID != "" {
+		return cliID, nil
+	}
+	cliID = uuid.New().String()
+	err := config.SetCLIId(cliID)
+	if err != nil {
+		return "", err
+	}
+	return cliID, nil
+}
 
 // Execute executes the CLI.
 func Execute() error {
@@ -266,5 +288,23 @@ func Execute() error {
 	if err != nil {
 		return err
 	}
-	return root.Execute()
+	executionErr := root.Execute()
+	exitCode := 0
+	if executionErr != nil {
+		exitCode = 1
+		if errStr, ok := executionErr.(*exec.ExitError); ok {
+			// If a plugin exited with an error, we don't want to print its
+			// exit status as a string, but want to use it as our own exit code.
+			exitCode = (errStr.ExitCode())
+		}
+	}
+
+	postRunMetrics := &telemetry.PostRunMetrics{ExitCode: exitCode}
+	if updateErr := telemetry.Client().UpdateCmdPostRunMetrics(postRunMetrics); updateErr != nil {
+		telemetry.LogError(updateErr, "")
+	} else if saveErr := telemetry.Client().SaveMetrics(); saveErr != nil {
+		telemetry.LogError(saveErr, "")
+	}
+
+	return executionErr
 }
--- a/pkg/telemetry/client.go
+++ b/pkg/telemetry/client.go
@@ -1,0 +1,247 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package telemetry collects the CLI metrics and sends the telemetry data to supercollider
+package telemetry
+
+import (
+	"encoding/json"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/buildinfo"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/cli"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
+	configlib "github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
+)
+
+var once sync.Once
+
+var client MetricsHandler
+
+type MetricsHandler interface {
+	// SetInstalledPlugins adds the installed plugins to the handler used to retrieve
+	// the plugin information
+	SetInstalledPlugins(plugins []cli.PluginInfo)
+	// UpdateCmdPreRunMetrics updates the metrics collected before running the command
+	UpdateCmdPreRunMetrics(cmd *cobra.Command, args []string) error
+	// UpdateCmdPostRunMetrics updates the metrics collected after command execution is completed
+	UpdateCmdPostRunMetrics(metrics *PostRunMetrics) error
+	// SaveMetrics saves the metrics to the metrics store/DB
+	SaveMetrics() error
+	// SendMetrics sends the metrics to the destination(metrics data lake)
+	SendMetrics() error
+}
+
+type telemetryClient struct {
+	currentOperationMetrics *OperationMetricsPayload
+	installedPlugins        []cli.PluginInfo
+	metricsDB               MetricsDB
+}
+
+type PostRunMetrics struct {
+	ExitCode int
+}
+type OperationMetricsPayload struct {
+	CliID         string
+	StartTime     time.Time
+	EndTime       time.Time
+	Args          []string
+	NameArg       string
+	CommandName   string
+	ExitStatus    int
+	PluginName    string
+	Flags         string
+	CliVersion    string
+	PluginVersion string
+	Target        string
+	Endpoint      string
+	Error         string
+}
+
+func Client() MetricsHandler {
+	once.Do(func() {
+		client = newTelemetryClient()
+	})
+	return client
+}
+
+func newTelemetryClient() MetricsHandler {
+	opMetrics := &OperationMetricsPayload{}
+	metricsDB := newSQLiteMetricsDB()
+	return &telemetryClient{
+		currentOperationMetrics: opMetrics,
+		metricsDB:               metricsDB,
+	}
+}
+
+func (tc *telemetryClient) SetInstalledPlugins(plugins []cli.PluginInfo) {
+	tc.installedPlugins = plugins
+}
+
+func (tc *telemetryClient) UpdateCmdPreRunMetrics(cmd *cobra.Command, args []string) error {
+	if err := ensureMetricsSource(); err != nil {
+		return errors.Wrap(err, "failed to ensure metrics source in the configuration file")
+	}
+
+	cliID, err := cliInstanceID()
+	if err != nil {
+		return errors.Wrap(err, "unable to get CLI Instance ID")
+	}
+
+	if isCoreCommand(cmd) {
+		return tc.updateMetricsForCoreCommand(cmd, args, cliID)
+	}
+
+	return tc.updateMetricsForPlugin(cmd, args, cliID)
+}
+
+func (tc *telemetryClient) UpdateCmdPostRunMetrics(metrics *PostRunMetrics) error {
+	if metrics == nil {
+		return errors.New("post metrics data is required for update")
+	}
+	tc.currentOperationMetrics.ExitStatus = metrics.ExitCode
+	tc.currentOperationMetrics.EndTime = time.Now()
+	return nil
+}
+
+func (tc *telemetryClient) SaveMetrics() error {
+	// If cli command fail cobra validation, the PersistentPreRunE() wouldn't be invoked where initialization is done
+	// so, it is safe to ignore the metrics for user errors(like typos) at least to an extent where cobra can validate.
+	if tc.currentOperationMetrics.StartTime.IsZero() {
+		return nil
+	}
+
+	err := tc.metricsDB.CreateSchema()
+	if err != nil {
+		return errors.Wrap(err, "unable to create the telemetry schema")
+	}
+
+	return tc.metricsDB.SaveOperationMetric(tc.currentOperationMetrics)
+}
+
+// SendMetrics sends the local stored metrics to super collider
+// TODO: to be implemented
+func (tc *telemetryClient) SendMetrics() error {
+	return nil
+}
+
+func isCoreCommand(cmd *cobra.Command) bool {
+	if cmd.Annotations != nil && cmd.Annotations["type"] == common.CommandTypePlugin {
+		return false
+	}
+	return true
+}
+func (tc *telemetryClient) updateMetricsForCoreCommand(cmd *cobra.Command, args []string, cliID string) error {
+	tc.currentOperationMetrics.CliID = cliID
+	tc.currentOperationMetrics.CliVersion = buildinfo.Version
+	tc.currentOperationMetrics.StartTime = time.Now()
+	tc.currentOperationMetrics.CommandName = strings.Join(strings.Split(cmd.CommandPath(), " ")[1:], " ")
+
+	// CLI recommendation is to have a single argument for a command
+	if len(args) != 0 {
+		tc.currentOperationMetrics.NameArg = hashString(args[0])
+	}
+
+	flagMap := make(map[string]string)
+	hashRequired := isHashRequiredForCmdFlags(cmd.CommandPath())
+	cmd.Flags().Visit(func(flag *pflag.Flag) {
+		// capture the boolean and empty flag values as is
+		if !hashRequired || flag.Value.String() == "" || flag.Value.Type() == "bool" {
+			flagMap[flag.Name] = flag.Value.String()
+		} else {
+			flagMap[flag.Name] = hashString(flag.Value.String())
+		}
+	})
+	if len(flagMap) != 0 {
+		jsonString, _ := json.Marshal(flagMap)
+		tc.currentOperationMetrics.Flags = string(jsonString)
+	}
+
+	return nil
+}
+
+func (tc *telemetryClient) updateMetricsForPlugin(cmd *cobra.Command, args []string, cliID string) error {
+	tc.currentOperationMetrics.CliID = cliID
+	tc.currentOperationMetrics.CliVersion = buildinfo.Version
+	tc.currentOperationMetrics.StartTime = time.Now()
+
+	flagNames := TraverseFlagNames(args)
+	if len(flagNames) > 0 {
+		tc.currentOperationMetrics.Flags = flagNamesToJSONString(flagNames)
+	}
+
+	plugin := tc.pluginInfoFromCommand(cmd)
+	if plugin != nil {
+		tc.currentOperationMetrics.PluginName = plugin.Name
+		tc.currentOperationMetrics.PluginVersion = plugin.Version
+		tc.currentOperationMetrics.Target = string(plugin.Target)
+		tc.currentOperationMetrics.Endpoint = getEndpointSHA(plugin)
+	}
+
+	// TODO : Fix the command chain for plugins by using command chain cache constructed from generate-all-docs command
+	tc.currentOperationMetrics.CommandName = strings.Join(strings.Split(cmd.CommandPath(), " ")[1:], " ")
+
+	return nil
+}
+func (tc *telemetryClient) pluginInfoFromCommand(cmd *cobra.Command) *cli.PluginInfo {
+	var plugin *cli.PluginInfo
+	if cmd.Annotations == nil || cmd.Annotations["pluginInstallationPath"] == "" {
+		return nil
+	}
+	for i := range tc.installedPlugins {
+		if cmd.Annotations["pluginInstallationPath"] == tc.installedPlugins[i].InstallationPath {
+			return &tc.installedPlugins[i]
+		}
+	}
+
+	return plugin
+}
+
+func cliInstanceID() (string, error) {
+	cliID, err := configlib.GetCLIId()
+	if err != nil {
+		return "", err
+	}
+	return cliID, nil
+}
+
+func ensureMetricsSource() error {
+	telemetryOptions, _ := configlib.GetCLITelemetryOptions()
+	dbFile := filepath.Join(common.DefaultCLITelemetryDir, SQliteDBFileName)
+	if telemetryOptions != nil && telemetryOptions.Source == dbFile {
+		return nil
+	}
+
+	err := configlib.SetCLITelemetryOptions(&configtypes.TelemetryOptions{Source: dbFile})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// isHashRequiredForCmdFlags determines if hashing is required for a core command read from command path
+// currently, for each command we are either hashing all the values or none. A possible enhancement would be
+// to return list of flags whose values need to be hashed.
+func isHashRequiredForCmdFlags(cmdPath string) bool {
+	coreCommandsAllowedWithFlagValues := map[string]struct{}{
+		"plugin": struct{}{},
+	}
+
+	cmds := strings.Split(cmdPath, " ")
+	if len(cmds) < 2 {
+		return false
+	}
+
+	if _, exists := coreCommandsAllowedWithFlagValues[cmds[1]]; exists {
+		return false
+	}
+	return true
+}
--- a/pkg/telemetry/client_test.go
+++ b/pkg/telemetry/client_test.go
@@ -1,0 +1,416 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package telemetry
+
+import (
+	"encoding/json"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/cli"
+	configlib "github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
+)
+
+func TestClient(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Telemetry plugin test suite")
+}
+
+type mockMetricsDB struct {
+	createSchemaCalled             bool
+	saveOperationMetricCalled      bool
+	createSchemaReturnError        error
+	saveOperationMetricReturnError error
+}
+
+func (mc *mockMetricsDB) CreateSchema() error {
+	mc.createSchemaCalled = true
+	return mc.createSchemaReturnError
+}
+
+func (mc *mockMetricsDB) SaveOperationMetric(payload *OperationMetricsPayload) error {
+	mc.saveOperationMetricCalled = true
+	return mc.saveOperationMetricReturnError
+}
+
+var _ = Describe("Unit tests for UpdateCmdPreRunMetrics()", func() {
+	const True = "true"
+	var (
+		tc           *telemetryClient
+		metricsDB    *mockMetricsDB
+		rootCmd      *cobra.Command
+		cmd          *cobra.Command
+		configFile   *os.File
+		configFileNG *os.File
+		err          error
+	)
+
+	BeforeEach(func() {
+		metricsDB = &mockMetricsDB{}
+		tc = &telemetryClient{
+			currentOperationMetrics: &OperationMetricsPayload{
+				StartTime: time.Time{},
+			},
+			metricsDB: metricsDB,
+		}
+
+		rootCmd = &cobra.Command{
+			Use: "tanzu",
+			RunE: func(cmd *cobra.Command, args []string) error {
+				return nil
+			},
+		}
+
+		configFile, err = os.CreateTemp("", "config")
+		Expect(err).To(BeNil())
+		os.Setenv("TANZU_CONFIG", configFile.Name())
+
+		configFileNG, err = os.CreateTemp("", "config_ng")
+		Expect(err).To(BeNil())
+		os.Setenv("TANZU_CONFIG_NEXT_GEN", configFileNG.Name())
+
+		err = configlib.SetCLIId("fake-cli-id")
+		Expect(err).ToNot(HaveOccurred(), "failed to set the CLI ID")
+
+	})
+	AfterEach(func() {
+		os.Unsetenv("TANZU_CONFIG")
+		os.Unsetenv("TANZU_CONFIG_NEXT_GEN")
+
+		os.RemoveAll(configFile.Name())
+		os.RemoveAll(configFileNG.Name())
+
+	})
+	Describe("when the command is CLI core command", func() {
+		BeforeEach(func() {
+			cmd = &cobra.Command{
+				Use: "corecommand",
+				RunE: func(cmd *cobra.Command, args []string) error {
+					return nil
+				},
+			}
+			rootCmd.AddCommand(cmd)
+		})
+		Context("when core command has arguments provided but no flags", func() {
+			It("should return success and the metrics should have name arg(first argument) hashed and flags string should be empty", func() {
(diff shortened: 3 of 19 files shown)


pkg/telemetry/client.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	err := configlib.SetCLITelemetryOptions(&configtypes.TelemetryOptions{Source: dbFile})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// isHashRequiredForCmdFlags determines if hashing is required for a core command read from command path
```
vuil, line 259: Nit: I suggest adding a comment about the limitation of this check, which is that for each command we are either hashing all values or none at all. A possible enhancement would be to return a list of flags whose values need to be hashed.

pkg/telemetry/client_test.go
```diff
@@ -0,0 +1,454 @@
...
+				metricsPayload := tc.currentOperationMetrics
+				Expect(metricsPayload.CommandName).To(Equal("corecommand"))
+				Expect(metricsPayload.NameArg).To(Equal(hashString("arg1")))
+				Expect(metricsPayload.Flags).To(BeEmpty())
+				Expect(metricsPayload.CliID).ToNot(BeEmpty())
+				Expect(metricsPayload.StartTime.IsZero()).To(BeFalse())
+
+			})
+		})
+		Context("when the core command(requiring flag values hashed) has arguments and flags provided", func() {
```
vuil, line 118: nit: `command (requiring hashed values)`...
similarly for consistency in L145
`command (not requiring hashed values)`...

pkg/telemetry/metrics_helper.go
```diff
@@ -0,0 +1,60 @@
...
+	ctx, exists := curCtx[configtypes.TargetK8s]
+	if exists {
+		ctxBytes, _ := json.Marshal(ctx)
+		// #nosec using the sha1 just for masking the value and not intended for signatures/certs
+		return hashString(string(ctxBytes))
+	}
+	return ""
+}
+
+// getCurrentTMCContextSHA returns SHA11 of concatenated string of Endpoint and RefreshToken
```
vuil, line 46: SHA256?

pkg/telemetry/metrics_helper.go
```diff
@@ -0,0 +1,60 @@
...
+	if plugin.Target == configtypes.TargetK8s {
+		return getCurrentK8SContextSHA(curCtxMap)
+	} else if plugin.Target == configtypes.TargetTMC {
+		return getCurrentTMCContextSHA(curCtxMap)
+	}
+
+	return ""
+}
+
+// getCurrentK8SContextSHA returns SHA1 of the complete context
```
vuil, line 35: SHA256?

pkg/telemetry/metrics_helper.go
```diff
@@ -0,0 +1,60 @@
...
+func getCurrentK8SContextSHA(curCtx map[configtypes.Target]*configtypes.Context) string {
+	ctx, exists := curCtx[configtypes.TargetK8s]
+	if exists {
+		ctxBytes, _ := json.Marshal(ctx)
+		// #nosec using the sha1 just for masking the value and not intended for signatures/certs
```
vuil, line 40: sha256?

vuil (commented):
Nice work!. I have a few small comments for now, but will give another round of comments shortly.

pkg/telemetry/client.go
```diff
@@ -0,0 +1,274 @@
...
+	AddToPluginsByTarget(target configtypes.Target, plugin *cli.PluginInfo)
+	UpdateCmdPreRunMetrics(cmd *cobra.Command, args []string) error
+	UpdateCmdPostRunMetrics(metrics *PostRunMetrics) error
+	SaveMetrics() error
+	SendMetrics() error
+}
+
+type telemetryClient struct {
+	currentOperationMetrics      *OperationMetricsPayload
+	coreCommandMap               map[string]bool
```
vuil, line 40: Not sure if there are additional plans for coreCommandMap, but as of now it looks like its purpose is purely to capture a table of top-level core commands and rely of checking any Cmd's CommandPath() against it to determine if the command is a core command or not.

But I am thinking if is necessary at all. It's not so much the complexity of the changes since the implementation is straightforward, but I wonder if the telemetry client is making too many assumptions on how to interpret each command. I can imagine a scenario in the future when the second token in the command path is found in this map but the actual sub command could be implemented by a plugin.

Given that the Command tree is fully constructed by the time any Command in the tree is execute-d, have you considered finding a way to distinguish a core Command from a plugin-based Command by examining the Command object directly?  It might be possible because every plugin-based Command is constructed by GetCmdForPlugin(), so there might be some opportunity to explicity identify this category of Commands.

pkg/telemetry/client.go
```diff
@@ -0,0 +1,274 @@
...
+	return nil
+}
+func (tc *telemetryClient) pluginInfoFromCommand(cmd *cobra.Command) *cli.PluginInfo {
+	var plugin *cli.PluginInfo
+
+	cmds := strings.Split(cmd.CommandPath(), " ")
+
+	// check if the second command in command path matches any target,
+	// if so, look for the 3rd command in the plugins map to find the plugin
+	cmdTarget := configtypes.StringToTarget(cmds[1])
```
vuil, line 220: I have similar thoughts here as https://github.com/vmware-tanzu/tanzu-cli/pull/391/files#r1260607600
Do we need to make the assumptions we make here about the command path in order to find the plugin involved (if it is a plugin-based command)
Conceptually, being able to map each plugin-based Command back to the corresponding plugininfo would seem a more direct (and more future-proof) way to implement this function

pkg/telemetry/client.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func (tc *telemetryClient) SaveMetrics() error {
+	// If cli command fail cobra validation, the PersistentPreRunE() wouldn't be invoked where initialization is done
+	// so, it is safe to ignore the metrics for user errors(like typos) at least to an extent where cobra can validate.
+	if tc.currentOperationMetrics.StartTime.IsZero() {
+		return nil
+	}
+
+	err := tc.metricsDB.CreateSchema()
```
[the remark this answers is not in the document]
prkalle: Yes, it should be a no-op if the table already exists.

pkg/telemetry/client.go
```diff
@@ -0,0 +1,274 @@
...
+	if cmdPath != "" {
+		cmds := strings.Split(cmdPath, " ")
+		if len(cmds) >= 2 && tc.coreCommandMap[cmds[1]] {
+			return true
+		}
+	}
+
+	return false
+}
+func (tc *telemetryClient) updateMetricsForCoreCommand(cmd *cobra.Command, args []string, cliID string) error {
```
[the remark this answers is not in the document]
prkalle: Sure. Updated the PR description. Thanks!

pkg/telemetry/metrics_helper.go
prkalle: Thanks for pointing it out. Missed to update the comments after I changed the SHA1 to SHA256.

pkg/telemetry/client.go
```diff
@@ -0,0 +1,274 @@
...
+	tc.currentOperationMetrics.CliID = cliID
+	tc.currentOperationMetrics.CliVersion = buildinfo.Version
+	tc.currentOperationMetrics.StartTime = time.Now()
+	tc.currentOperationMetrics.CommandName = strings.Join(strings.Split(cmd.CommandPath(), " ")[1:], " ")
+
+	// CLI recommendation is to have a single argument for a command
+	if len(args) != 0 {
+		tc.currentOperationMetrics.NameArg = hashString(args[0])
+	}
```
anujc25, line 170: I remember something regarding why only a single argument is recommended got discussed somewhere, but can you point me to that discussion or can you remind me the reasons behind this as I am not able to remember now?

pkg/telemetry/metrics_db_lock.go
```diff
@@ -0,0 +1,85 @@
...
+	var err error
+
+	if cliMetricDBLockFile == "" {
+		cliMetricDBLockFile = filepath.Join(common.DefaultCLITelemetryDir, LocalTanzuCLIMetricsDBFileLock)
+	}
+
+	// using fslock to handle interprocess locking
+	lock, err := getFileLockWithTimeOut(cliMetricDBLockFile, DefaultMetricsDBLockTimeout)
+	if err != nil {
+		panic(fmt.Sprintf("cannot acquire lock for Tanzu CLI metrics DB, reason: %v", err))
```
anujc25, line 47: In the rare scenario of a lock error, I am thinking about, whether or not should we panic here or skip the telemetry collection if this happens for any reason.

pkg/telemetry/metrics_db_lock.go
```diff
@@ -0,0 +1,85 @@
...
+		panic(fmt.Sprintf("cannot release lock for Tanzu CLI metrics DB, reason: %v", errUnlock))
+	}
+
+	cliMetricDBLock = nil
+	// Unlock the mutex to allow other concurrent calls to acquire and configure the cliMetricDBLock
+	cliMetricDBMutex.Unlock()
+}
+
+// getFileLockWithTimeOut returns a file lock with timeout
+func getFileLockWithTimeOut(lockPath string, lockDuration time.Duration) (*fslock.Lock, error) {
```
anujc25, line 70: nit: 
```suggestion
func getFileLockWithTimeout(lockPath string, lockDuration time.Duration) (*fslock.Lock, error) {
```

pkg/telemetry/client.go
```diff
@@ -0,0 +1,274 @@
...
+	if len(flagNames) > 0 {
+		tc.currentOperationMetrics.Flags = flagNamesToJSONString(flagNames)
+	}
+
+	plugin := tc.pluginInfoFromCommand(cmd)
+	if plugin != nil {
+		tc.currentOperationMetrics.PluginName = plugin.Name
+		tc.currentOperationMetrics.PluginVersion = plugin.Version
+		tc.currentOperationMetrics.Target = string(plugin.Target)
+		tc.currentOperationMetrics.Endpoint = getEndpointSHA(plugin)
```
anujc25, line 205: What does the `Endpoint` mean here?

pkg/telemetry/metrics_helper.go
```diff
@@ -0,0 +1,60 @@
...
+	curCtxMap, err := cfg.GetAllCurrentContextsMap()
+	if err != nil || curCtxMap == nil {
+		return ""
+	}
+
+	if plugin.Target == configtypes.TargetK8s {
+		return getCurrentK8SContextSHA(curCtxMap)
+	} else if plugin.Target == configtypes.TargetTMC {
+		return getCurrentTMCContextSHA(curCtxMap)
+	}
```
anujc25, line 30: Can we make this generic and not have separate functions like this for each active context for the target?

In the future, we would like to decouple the context/target, and having a specific function for each target does not make sense considering that.

What is the reason for doing calculating SHA256 for both the `Context` objects differently?

pkg/telemetry/metrics_helper.go
prkalle: Removing this comment as it(#nosec) is not required after i changes SHA1 to SHA256.

pkg/telemetry/client.go
prkalle: updated.

pkg/telemetry/client_test.go
prkalle: updated

pkg/telemetry/client.go
```diff
@@ -0,0 +1,276 @@
...
+	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
+	configlib "github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
+)
+
+var once sync.Once
+
+var client MetricsHandler
+
+type MetricsHandler interface {
```
chandrareddyp, line 29: Please add documentation for each method

pkg/telemetry/client.go
prkalle: Great point!! Thanks for brining this up. Updated the PR to use the command annotation from the Command directly and wouldn't need `coreCommandMap` anymore.

pkg/command/root.go
```diff
@@ -253,12 +269,42 @@ func duplicateAliasWarning(rootCmd *cobra.Command) {
...
+		exitCode = 1
+		if errStr, ok := err.(*exec.ExitError); ok {
+			// If a plugin exited with an error, we don't want to print its
+			// exit status as a string, but want to use it as our own exit code.
+			exitCode = (errStr.ExitCode())
+		}
+	}
+
+	postRunMetrics := &telemetry.PostRunMetrics{ExitCode: exitCode}
+	if err := telemetry.Client().UpdateCmdPostRunMetrics(postRunMetrics); err != nil {
```
chandrareddyp, line 303: Nit:IMO, I would suggest renaming the error variables here for better readability. As we are returning the `err = root.Execute()` at the end.

pkg/telemetry/client.go
prkalle: Thanks! 
Same as above comment. I would be using the Command object (a new annotation `pluginInstallationPath`:<plugin/installation/path>` would be added and would use this as key to find the plugin) used for the plugin command to find the plugin involved.
Updated the PR.

pkg/telemetry/client.go
prkalle: https://github.com/vmware-tanzu/tanzu-cli/blob/main/docs/plugindev/style_guide.md#positional-arguments

pkg/telemetry/client.go
prkalle: `endpoint` is the endpoint that the command is directed at.
It could be kubernetes cluster or SaaS endpoint.

pkg/telemetry/client.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func Client() MetricsHandler {
+	once.Do(func() {
+		client = NewTelemetryClient()
+	})
+	return client
+}
+
+func NewTelemetryClient() MetricsHandler {
```
chandrareddyp, line 70: is there any specific reason we are making it public instead of private, as we have `Client()` to create a single instance of the client? 
If we make it public, we may end up creating multiple instances!!

pkg/telemetry/metrics_db_lock.go
prkalle: I was thinking about it too, but since it is very corner case, I choose to go with `Panic`. Now after thinking a bit more, it make sense to return error(which would be ignored for telemetry unless user chooses to show the telemetry errors) instead of `panic` especially if the operation takes too long to complete, and CLI shouldn't give the impression that operation is failed due to telemetry. I will update the PR.

pkg/telemetry/sqlite_metrics_db.go
```diff
@@ -0,0 +1,152 @@
...
+	nameArg            string
+	endpoint           string
+	flags              string
+	exitStatus         int
+	isInternal         bool
+	error              string
+}
+
+// NewSQLiteMetricsDB returns a new PluginInventory connected to the data found at 'metricsDBFile'.
+func NewSQLiteMetricsDB() MetricsDB {
```
chandrareddyp, line 61: Is there any reason we are making it public? as we are hard-coding the Telemetry SQlite DB file name, i would not make it public! If we are passing the path as input then maybe we can make it public!

pkg/telemetry/sqlite_metrics_db.go
```diff
@@ -0,0 +1,152 @@
...
+	dbFile := filepath.Join(common.DefaultCLITelemetryDir, SQliteDBFileName)
+	return &sqliteMetricsDB{
+		metricsDBFile: dbFile,
+	}
+}
+
+// CreateSchema creates table schemas to the provided database.
+// returns error if table creation fails for any reason
+func (b *sqliteMetricsDB) CreateSchema() error {
+	dirName := filepath.Dir(b.metricsDBFile)
```
chandrareddyp, line 71: We need to consider the following use cases:
What if the SQLite file/schema already exists?
Should we force create, potentially deleting the existing schema and creating a new one?
What should we do if we already have schema and data? We may need to ask the user for permission on what action to take.

pkg/telemetry/metric_db.go
```diff
@@ -0,0 +1,13 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package telemetry
+
+type MetricsDB interface {
+	// CreateSchema creates table schemas to the provided database.
```
chandrareddyp, line 7: We may need to consider more use cases:
Delete Schema
Update Schema
Delete Data
Get/Retrieve Data

pkg/telemetry/metrics_db_lock.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
+
+	"github.com/juju/fslock"
+)
+
+const (
+	LocalTanzuCLIMetricsDBFileLock = ".cli_metrics_db.lock"
+	// DefaultMetricsDBLockTimeout is the default time waiting on the filelock
+	DefaultMetricsDBLockTimeout = 10 * time.Minute
```
chandrareddyp, line 23: IMO, 10 mins is too high!! 
How about configuring it with env variable or config!

pkg/telemetry/metrics_db_lock.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// using fslock to handle interprocess locking
+	lock, err := getFileLockWithTimeOut(cliMetricDBLockFile, DefaultMetricsDBLockTimeout)
+	if err != nil {
+		panic(fmt.Sprintf("cannot acquire lock for Tanzu CLI metrics DB, reason: %v", err))
+	}
+
+	// Lock the mutex to prevent concurrent calls to acquire and configure the cliMetricDBLock
+	cliMetricDBMutex.Lock()
+	cliMetricDBLock = lock
```
chandrareddyp, line 52: How about releasing the lock after a certain period of time, such as 5 minutes, regardless of the `ReleaseTanzuMetricDBLock()` function being called!

pkg/telemetry/sqlite_metrics_db.go
prkalle: I will update it to private as we accessing it from the same package, we can make it public if this is moved to different package under telemetry in future.

pkg/telemetry/client.go
prkalle: Good point. I will update the PR

chandrareddyp (commented):
Great work @prkalle!! I have added few comments. Thanks.

pkg/telemetry/metrics_helper.go
prkalle: > Can we make this generic and not have separate functions like this for each active context for the target?
> 
> In the future, we would like to decouple the context/target, and having a specific function for each target does not make sense considering that.

> What is the reason for doing calculating SHA256 for both the Context objects differently?

If we take SHA256 for the contexts same (taking SHA256 of Context structure) , it would result in a new SHA everytime the `IDToken` or `accessToken` changes and these are shortlived. For Kubernetes context, we can take the SHA of the complete context as there are no shortlived values there.

In future if we decouple the context/target, we have to find out other means to detect the context that plugin uses and hash the right context. So it would be incremental change

pkg/telemetry/sqlite_metrics_db.go
prkalle: If the schema already exists and if we need to update the schema, the new CLI version schema file would have the sql commands to update the schema. This DB is entirely for telemetry and users are not going to access it.

pkg/telemetry/metric_db.go
prkalle: CLI is producer here and not consuming the data from DB. Telemetry plugin would be consuming the data from DB to send it to data lake. So for now `CreateSchema` should be sufficent, we would other operations if needed in future.

pkg/telemetry/metrics_db_lock.go
prkalle: Good point.I would reduce it to 2 seconds to make sure it would not add latency to the user commands as telemetry is best-effort and should not affect the user experience on CLI/plugin functionality

pkg/telemetry/metrics_db_lock.go
prkalle: The client code using this `ReleaseTanzuMetricDBLock()`should have a  timeout mechanism to release the lock( calling the `ReleaseTanzuMetricDBLock()` method after timeout.

pkg/telemetry/metrics_db_lock.go
prkalle: updated

pkg/command/root.go
```diff
@@ -125,6 +131,10 @@ func newRootCmd() *cobra.Command {
 		// Flag parsing must be deactivated because the root plugin won't know about all flags.
 		DisableFlagParsing: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if err := telemetry.Client().UpdateCmdPreRunMetrics(cmd, args); err != nil {
```
chandrareddyp, line 134: Do we really need to make telemetry calls synchronous? How about making the telemetry calls` (UpdateCmdPreRunMetrics, UpdateCmdPostRunMetrics, SaveMetrics, and SendMetrics)` asynchronous using goroutines? Since the CLI command will continue irrespective of these telemetry calls, making them asynchronous can improve performance and responsiveness.
cc: @anujc25

pkg/telemetry/metrics_helper.go
vuil: @anujc25 The goal of the endpoint value computation is to produce the most stable, non-PII value that can be used to accurately distinguish among different endpoints the commands are targeting. The heuristics used here, which relies on using the context data,  needs to be tuned differently for different styles of context for best effect.

@prkalle  That said, perhaps the fn names (getCurrentK8SContextSHA. ...) is setting off a wrong impression that we have multiple ways of computing a SHA.  Maye names like getEndpointSHAForK8SContext, getEndpointSHAForGlobalContext would be more reflective of what the function does.

Alternatively, (similar to what @anujc25 suggested) can we just write a single function computeEndpointSHAForContext that does the right thing depending on what is set in the context? (Some outer code will have to determine that context to compute and pass that into this function)

pkg/command/root.go
prkalle: `UpdateCMDPostRunMetrics`,SaveMetrics` and SendMetrics` are sequential and should be called after the CLI command(core/plugin) is executed, as we need to collect the metrics after the CLI command is execited. `UpdateCmdPreRunMetrics` is also not blocking/time consuming(since we already have command tree cache). So, I don't see a necessity for asynchronous calls at the moment.

pkg/telemetry/metrics_helper.go
prkalle: Updated the PR with single function `computeEndpointSHAForContext`that computes the endpoint SHA for the context,

anujc25 (approved):
This looks good. Thanks @prkalle for addressing all the comments.

chandrareddyp (approved):
LGTM!

PR #398, opened 2023-07-13
 Add command tree parser to parse the plugin command path for plugin telemetry

prkalle:
This PR  adds command tree parser to parse the plugin command path for plugin telemetry.

**Why we need this:**
 With the CLI plugin architecture, CLI attaches the plugin binary as a cobra command. Hence when user runs the CLI plugin command, Cobra has visibility into the args and command path up to the command to which the plugin was attached. IT doesn't have visibility into the command sub-tree that plugin implemented.  However to collect the telemetry, and to make meaningful data, we need the complete command path executed by the user.
_For example:_
 Without command tree parser if user executes the below command 
      > ./bin/tanzu management-cluster ceip-participation set true
CLI can only get command path as `tanzu management-cluster`

With command tree parser , CLI can get the command path as `management-cluster ceip-participation set`

**How are we generating the plugin command tree:**
CLI uses the plugin command `generate_docs` (default command that plugins support) to get the complete command chains supported. However the docs generated doesn't provide the information regarding the aliases of the command/sub-commands. So, CLI would use the help command for each sub-command to extract the aliases supported and fianlly construct the plugin command tree and adds it to cache so that CLI telemetry client(collector) can extract the command chain by parsing the user input against the plugin command tree.

Fixes #

Installed the plugins

```
❯ ./bin/tanzu plugin list
Standalone Plugins
  NAME                DESCRIPTION                                                        TARGET      VERSION  STATUS
  pinniped-auth       Pinniped authentication operations (usually not directly invoked)  global      v0.29.0  installed
  management-cluster  Kubernetes management cluster operations                           kubernetes  v0.29.0  installed
  package             Tanzu package management                                           kubernetes  v0.29.0  installed
  secret              Tanzu secret management                                            kubernetes  v0.29.0  installed
  telemetry           configure cluster-wide settings for vmware tanzu telemetry         kubernetes  v0.29.0  installed

Plugins from Context:  tkg-mgmt-vc
  NAME                DESCRIPTION                           TARGET      VERSION  STATUS
  cluster             Kubernetes cluster operations         kubernetes  v0.28.0  installed
  feature             Operate on features and featuregates  kubernetes  v0.28.0  installed
  kubernetes-release  Kubernetes release operations         kubernetes  v0.28.0  installed
```

Ran the plugin command and the command name now shows the complete path `management-cluster get`
```
❯ ./bin/tanzu mc get --show-details
  NAME         NAMESPACE   STATUS   CONTROLPLANE  WORKERS  KUBERNETES        ROLES       PLAN  TKR
  tkg-mgmt-vc  tkg-system  running  1/1           1/1      v1.24.9+vmware.1  management  dev   v1.24.9---vmware.1-tkg.1


Details:

NAME                                                                                READY  SEVERITY  REASON  SINCE  MESSAGE
/tkg-mgmt-vc                                                                        True                     2d17h
├─ClusterInfrastructure - VSphereCluster/tkg-mgmt-vc-wmt59                          True                     2d17h
├─ControlPlane - KubeadmControlPlane/tkg-mgmt-vc-vjx7j                              True                     2d17h
│ └─Machine/tkg-mgmt-vc-vjx7j-gz9dm                                                 True                     2d17h
│   ├─BootstrapConfig - KubeadmConfig/tkg-mgmt-vc-vjx7j-b5bpf                       True                     2d17h
│   └─MachineInfrastructure - VSphereMachine/tkg-mgmt-vc-control-plane-8btvw-zftsr  True                     2d17h
└─Workers
  └─MachineDeployment/tkg-mgmt-vc-md-0-stbl5                                        True                     25h
    └─Machine/tkg-mgmt-vc-md-0-stbl5-5b6f788564-znb7j                               True                     2d17h
      ├─BootstrapConfig - KubeadmConfig/tkg-mgmt-vc-md-0-bootstrap-jw82n-j8lgv      True                     2d17h
      └─MachineInfrastructure - VSphereMachine/tkg-mgmt-vc-md-0-infra-p7hrt-86t2w   True                     2d17h
<SNAP OUTPUT>

# you can see the command column shows "management-cluster get"
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM tanzu_cli_operations;
v1.0.0-dev|darwin|amd64|management-cluster|v0.29.0|management-cluster get|7dbe3b36-f07b-4206-bd77-0f7728e75cc2|1689313153848|1689313156003|||kubernetes||58fabb359b0e763b10e0ca376f00cf89951251e8b8772601ae3cdff14a14e7f6|{"show-details":""}|0|1|
```

Ran another plugin command with more subcommands and the command path is returned correctly
```
❯ ./bin/tanzu mc ceip-participation set true
tanzu management-cluster ceip-participation is deprecated on TKG 1.6 and beyond - please use tanzu telemetry update --ceip-opt-out/in to manage CEIP settings

# you can observe the command column value is correctly set to "management-cluster ceip-participation set"
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM tanzu_cli_operations;
v1.0.0-dev|darwin|amd64|management-cluster|v0.29.0|management-cluster ceip-participation set|7dbe3b36-f07b-4206-bd77-0f7728e75cc2|1689313430165|1689313430704|||kubernetes||58fabb359b0e763b10e0ca376f00cf89951251e8b8772601ae3cdff14a14e7f6||0|1|
```

Ran a command with invalid subcommand and the metrics was collected upto the subcommand that is valid, and the exit status was zero (for some reason cobra returns sub-command mismatch with exit status zero and shows help.
```
❯ rm ~/.config/tanzu-cli-telemetry/cli_metrics.db
❯ ./bin/tanzu mc ceip-participation set-invalid true
Get or set ceip participation

Usage:
  tanzu management-cluster ceip-participation [command]

Available Commands:
  get         Get the current CEIP opt-in status
  set         Set the opt-in preference for CEIP

Flags:
  -h, --help   help for ceip-participation

Global Flags:
      --log-file string   Log file path
  -v, --verbose int32     Number for the log level verbosity(0-9)

Use "tanzu management-cluster ceip-participation [command] --help" for more information about a command.


# you can see the path is set correctly upto valid command which is "management-cluster ceip-participation" but the exit status is set to 0(zero) because command/plugin return exit status as 0 when the subcommand doesn't match.
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM tanzu_cli_operations;
v1.0.0-dev|darwin|amd64|management-cluster|v0.29.0|management-cluster ceip-participation|7dbe3b36-f07b-4206-bd77-0f7728e75cc2|1689313617433|1689313617747|||kubernetes||58fabb359b0e763b10e0ca376f00cf89951251e8b8772601ae3cdff14a14e7f6||0|1|
```

Ran a command with invalid subcommand and invalid flag name. The metrics was collected upto the subcommand that is valid and the exist status was non-zero(1) 
```
❯ rm ~/.config/tanzu-cli-telemetry/cli_metrics.db
❯ ./bin/tanzu mc ceip-participation set-invalid -x
Time taken for ConstructAndAddTree() is : 26.333µs
Time taken for GetTree() is : 235.625µs
Error: unknown shorthand flag: 'x' in -x
select * from tanzu_cli_operations;
cli_version|os_name|os_arch|plugin_name|plugin_version|command|cli_id|command_start_ts|command_end_ts|csp_org_id|account_number|target|name_arg|endpoint|flags|exit_status|is_internal|error
v1.0.0-dev|darwin|amd64|management-cluster|v0.29.0|management-cluster ceip-participation|7dbe3b36-f07b-4206-bd77-0f7728e75cc2|1689637997205|1689637997455|||kubernetes||58fabb359b0e763b10e0ca376f00cf89951251e8b8772601ae3cdff14a14e7f6|{"x":""}|1|1|
```

Tested with another plugin`cluster` and it was successful

```
❯ ./bin/tanzu cluster mhc control-plane delete  -y
Error: accepts 1 arg(s), received 0
SELECT cli_version,os_name,os_arch,plugin_name,plugin_version,command,cli_id,command_start_ts,command_end_ts,csp_org_id,account_number,target,name_arg,endpoint,flags,exit_status,is_internal,error FROM tanzu_cli_operations;
v1.0.0-dev|darwin|amd64|cluster|v0.28.0|cluster mhc control-plane delete|7dbe3b36-f07b-4206-bd77-0f7728e75cc2|1689313802097|1689313803206|||kubernetes||58fabb359b0e763b10e0ca376f00cf89951251e8b8772601ae3cdff14a14e7f6|{"y":""}|1|1|
```

### Release note

```release-note

```

This PR would be updated based on the comments on it's dependent PR :https://github.com/vmware-tanzu/tanzu-cli/pull/391. But the changes should be minimal and is good for review now


vuil:
Thanks for the changes. I will add review comments specific to the command tree parser commit, and use https://github.com/vmware-tanzu/tanzu-cli/pull/391 for what's before it.

Some quick comments on the PR description:

I suggest add a sentence or two to explain _why_ we needed this change.

The command example and sqllite db output are really useful. But any reason why you are not using `select *` and include the headers? Also the data output seems not up-to-date.
(e.g. Can you check the output for the error case? I would have expected to see the error_status to be non-zero.)

Tip: set `.headers on` in your ~/sqliterc before running your sql query so it is easier to match up column names with row data.

pkg/plugincmdtree/plugins_cache_test.go
```diff
@@ -0,0 +1,253 @@
...
+fi
+`
+
+const expectedPluginCmdTree string = `
+commandTree:
+  ? pluginName: sample-plugin
+    target: kubernetes
+    version: 1.0.0
+  : subcommands:
+      sub-command1:
```
vuil, line 49: nit: sub commands prefixed with 'sub' or 'sub-command' mingled with yaml nodes whose names are 'subcommands' makes  for a rather confusing read :-). I would have picked other set of names with some well know order (e.g. foo, bar, baz, foo1, foo2, ..)

Also I am not sure what happened to sub-command1-sub and the aliases in the expectedPluginCmdTree.

pkg/plugincmdtree/plugins_cache.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func NewCache() (Cache, error) {
+	pct, err := getPluginCommandTree()
+	if err != nil {
+		return nil, err
+	}
+	return &cacheImpl{
+		pluginCommands:      pct,
+		pluginDocsGenerator: generatePluginDocs,
```
vuil, line 62: Nit: somewhere in the cacheImpl definition or implementation, we should briefly explain the approach taken here of leveraging generate-docs and parsing its output. The parsing adds complexity, and might get even messier when we need to learn more about each command beyond its subcommands and aliases. Another alternative would be to introduce plugin-runtime changes generate an even more structure and detailed output of a plugin's command tree.

pkg/plugincmdtree/plugins_cache.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func (c *cacheImpl) GetTree(plugin *cli.PluginInfo) (*CommandNode, error) {
+	// This is just a safety net to generate the command tree if we missed/failed to generate the plugin command tree during plugin install
+	// If the plugin command tree exists, then ConstructAndAddTree is a no-op
+	if err := c.ConstructAndAddTree(plugin); err != nil {
+		return nil, err
+	}
+
+	pkey := pluginKey{
```
vuil, line 73: Something I learned today :-)
That said, I believe in #391  there was some discussion of using the plugininfo's plugin path as a uniq identifier. Would that work here as well for some consistency?

pkg/plugincmdtree/plugins_cache.go
```diff
@@ -0,0 +1,257 @@
...
+
+	return &ctr, nil
+}
+
+func generatePluginDocs(plugin *cli.PluginInfo) error {
+	docsDir := getPluginsDocsCachePath()
+	_ = os.RemoveAll(docsDir)
+	_ = os.MkdirAll(docsDir, 0755)
+
+	args := []string{"generate-docs", "--docs-dir", docsDir}
```
vuil, line 221: It's hard to explain to the curious user what the doc/ subdir is, since contents come and go based on what the latest plugins we invoked the generate-docs command on. How about hiding it using a name like .docs?

pkg/plugincmdtree/plugins_cache.go
```diff
@@ -0,0 +1,257 @@
...
+		return nil, err
+	}
+	return &cacheImpl{
+		pluginCommands:      pct,
+		pluginDocsGenerator: generatePluginDocs,
+	}, nil
+}
+
+func (c *cacheImpl) GetTree(plugin *cli.PluginInfo) (*CommandNode, error) {
+	// This is just a safety net to generate the command tree if we missed/failed to generate the plugin command tree during plugin install
```
vuil, line 67: I like that we get a chance to update the command tree for the plugin if we didn't build it properly. 

(nit) However, do we need to be also concerned about the situation where the subtree exists but is stale/incorrect for some reason?

vuil (commented):
Nice! I am okay with the general approach. My comments are fairly minor.

pkg/plugincmdtree/plugins_cache_test.go
prkalle: ACK. Updated the command name to foo1,bar1 and sub-subcommand to follow the order foo2(sub-coomand of foo1). 
Also fixed the issue with `sub-command1-sub` aliases (doc for the same was not pre-generated in the test hence it was missed)

pkg/plugincmdtree/plugins_cache.go
prkalle: Good point! Updated the docs path to use `.docs`

pkg/plugincmdtree/plugins_cache.go
prkalle: Yes, it works (since CLI installation path includes plugin name,version,target and SHA, it gives enough information to debug if there is any undesired behavior), I updated the PR to use plugininfo's plugin path as a uniq identifier.

pkg/plugincmdtree/plugins_cache.go
prkalle: Sure. Added the brief explanation on the approach taken and the possible future enhancement option to teach plugin through plugin-runtime library to return command tree with exhaustive details that can be used for parsing.

prkalle:
> Thanks for the changes. I will add review comments specific to the command tree parser commit, and use #391 for what's before it.
> 
> Some quick comments on the PR description:
> 
> I suggest add a sentence or two to explain _why_ we needed this change.

ACK. Updated the PR description

> The command example and sqllite db output are really useful. But any reason why you are not using `select *` and include the headers? Also the data output seems not up-to-date. (e.g. Can you check the output for the error case? I would have expected to see the error_status to be non-zero.)
> 
> Tip: set `.headers on` in your ~/sqliterc before running your sql query so it is easier to match up column names with row data.

Thanks for the Tip 👍 (I was not aware of the`.header` and that's the reason I used `select col1,col2` instead of `select *`.
Regarding the data output, it is correct. In the error case, if the subcommand doesn't match, cobra/plugin is return exit status as `0` and showing help. However if we use wrong flag name , it is returning non-zero exit status. I will add another test result in the PR description to show this.

> Finally, it might be good to get some timing numbers for how long a command takes when a command tree needs to be recomputed for a plugin with a fairly complex command tree

Below are the stats( instrumented the code to capture the timing numbers) while a command tree needs to be constructured vs command tree served from cache.

```
❯ #time taken to construct the plugin command tree is 429.803833ms and this is one time taxation(usually during plugin installation for new plugins, if user already has plugins installed, user will observe this latency one time)
❯ time ./bin/tanzu mc ceip-participation set true
Time taken for ConstructAndAddTree() is : 429.494292ms
Time taken for GetTree() is : 429.803833ms
tanzu management-cluster ceip-participation is deprecated on TKG 1.6 and beyond - please use tanzu telemetry update --ceip-opt-out/in to manage CEIP settings
./bin/tanzu mc ceip-participation set true  1.79s user 0.79s system 183% cpu 1.409 total

❯ #time taken to get plugin command tree from cache is 225.45µs
❯ time ./bin/tanzu mc ceip-participation set true
Time taken for ConstructAndAddTree() is : 19.125µs
Time taken for GetTree() is : 225.458µs
tanzu management-cluster ceip-participation is deprecated on TKG 1.6 and beyond - please use tanzu telemetry update --ceip-opt-out/in to manage CEIP settings
./bin/tanzu mc ceip-participation set true  0.37s user 0.10s system 49% cpu 0.948 total
```

pkg/plugincmdtree/plugins_cache.go
```diff
@@ -0,0 +1,247 @@
...
+				if !current.AliasProcessed {
+					// kickoff the goroutine to add the alias to the command
+					aliasErrGroup.Go(func() error {
+						cmdAlias, aliasErr := getPluginCommandAlias(plugin, aliasArgsCopy)
+						if aliasErr != nil {
+							return aliasErr
+						}
+						current.Aliases = cmdAlias
+						return nil
+					})
```
anujc25, line 172: I think we will need to pass the `aliasArgsCopy` and `current` objects to the goroutine otherwise the reference for both might be different within the goroutine at the time of execution.
Reference: https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

```suggestion
					aliasErrGroup.Go(func(aliasArgsCopy []string, current *CommandNode) error {
						cmdAlias, aliasErr := getPluginCommandAlias(plugin, aliasArgsCopy)
						if aliasErr != nil {
							return aliasErr
						}
						current.Aliases = cmdAlias
						return nil
					}(aliasArgsCopy, current))
```

anujc25 (commented):
This is great and amazing work @prkalle to capture the plugin command details. 👏 

I have added one comment but overall looks good.

pkg/plugincmdtree/plugins_cache.go
prkalle: Good point! Thanks @anujc25. As mentioned in the reference we should be good if we declare a new variable with in the loop, we are be good with `aliasArgsCopy` (as we are declaring a new variable for each iteration). So I will update the PR by declaring another variable for `current` and use it in the go routine function because errorgroup `Go` funciton only support `Go( f func() error)` (without parameters). I did test it and its working fine.

anujc25 (approved):
LGTM. Thanks!

PR #403, opened 2023-07-17
Build the RPM repository incrementally and provide 'unstable' package

marckhouzam:
### What this PR does / why we need it

**First commit**
A requirement that we have is that our repo is on a GCP bucket (at least for the moment).  Therefore, all existing versions of the CLI packages are on the bucket and not on the machine where we build the repo.  We don't want to have to download every package from the bucket to be able to build a new repo as this would take more and more time as we accumulate versions.  Instead we want to build the repo incrementally by simply adding a new package version to it.

After investigation, the best we I found to do this was to download the existing repo metadata, locally create empty packages for the existing packages, and the rebuild the metadata with using:
  createrepo --update --skip-stat

This approach re-uses the previous metadata but adds to it the new package versions.

This commit also numbers pre-releases appropriately for RPM. In case we ever put pre-release packages in a repo along with final releases, it is important that the pre-releases be numbered as such for RPM to prioritize the final releases.

The commit also does a little cleanup to use a loop per architecture.

**Second commit**

The second commit (part of this PR because it needs to be based on the first) detects when the version is a pre-release and in that case builds an `tanzu-cli-unstable` package.  This allows to publish pre-release without having final users install them automatically.

### Which issue(s) this PR fixes

Fixes # N/A

### Describe testing done for PR

Build the RPM repo:
```
$ \rm -rf hack/rpm/_output/
$ make rpm-package
docker run --rm -e VERSION=v1.0.0-dev -e RPM_SIGNER= -e RPM_METADATA_BASE_URI= -v /Users/kmarc/git/tanzu-cli:/Users/kmarc/git/tanzu-cli fedora /Users/kmarc/git/tanzu-cli/hack/rpm/build_package.sh
++ uname
+ '[' Linux '!=' Linux ']'

[...]

RPM build warnings:
    source_date_epoch_from_changelog set but %changelog is missing
    Missing build-id in /root/rpmbuild/BUILDROOT/tanzu-cli-1.0.0-0.1_dev.aarch64/usr/bin/tanzu
+ mv /root/rpmbuild/RPMS/aarch64/tanzu-cli-1.0.0-0.1_dev.aarch64.rpm /Users/kmarc/git/tanzu-cli/hack/rpm/_output/rpm/tanzu-cli/
skip rpmsigning packages for aarch64
+ [[ ! -z '' ]]
+ echo skip rpmsigning packages for aarch64

# This is where we start building the repository
[...]

PR #402, opened 2023-07-17
Don't use reprepro for building the apt repo

marckhouzam:
### What this PR does / why we need it

**First commit**

Reprepro does not allow to have more than one version of a package in the same repo.  Since we want to keep multiple version of the CLI packages available, we can no longer use reprepro.

Another requirement that we have is that our repo is on a GCP bucket (at least for the moment).  Therefore, all existing versions of the CLI packages are on the bucket and not on the machine where we build the repo.  We don't want to have to download every package from the bucket to be able to build a new repo as this would take more and more time as we accumulate versions.  Instead we want to build the repo incrementally by simply adding a new package version to it.

After investigating different alternatives, mostly `aptly`, it became apparent that building a repository incrementally was not easy to do with existing tools.  On the other hand, building the repo manually ourselves turned out to be relatively easy.

Therefore, this commit replaces `reprepro` with a homemade script that allows us to add a new version to an existing repo without having to have the entire set of packages on the local machine.

**Second commit**

The second commit (part of this PR because it needs to be based on the first) detects when the version is a pre-release and in that case builds an `tanzu-cli-unstable` package. This allows to publish pre-releases without having final users install them automatically.

### Which issue(s) this PR fixes

Fixes # N/A

### Describe testing done for PR

Build the APT repo using a fedora image:
```
$ \rm -rf artifacts hack/apt/_output
$ make build-cli-linux-amd64
build linux-amd64 CLI with version: v1.0.0-dev

$ make apt-package APT_BUILDER_IMAGE=fedora
docker run --rm -e VERSION=v1.0.0-dev -e DEB_SIGNER= -v /Users/kmarc/git/tanzu-cli:/Users/kmarc/git/tanzu-cli fedora /Users/kmarc/git/tanzu-cli/hack/apt/build_package.sh
++ uname
+ '[' Linux '!=' Linux ']'
++ command -v apt-get
++ command -v dnf
+ PKG_MGR=/usr/bin/dnf
+ '[' -z /usr/bin/dnf ']'
+ '[' -z v1.0.0-dev ']'
+ VERSION=1.0.0-dev

[...]

+ echo skip debsigning
skip debsigning

# This part is where we build the repository
docker run --rm -e VERSION=v1.0.0-dev -e DEB_SIGNER= -e DEB_METADATA_BASE_URI= -v /Users/kmarc/git/tanzu-cli:/Users/kmarc/git/tanzu-cli fedora /Users/kmarc/git/tanzu-cli/hack/apt/build_package_repo.sh
[...]

PR #404, opened 2023-07-19
Documentation on how to install older CLIs

marckhouzam:

This is simply an update to the installation documentation to provide information on how to install older versions of the CLI using package managers.

This is related to #402 and #403 

Fixes #

Look at the new doc at https://github.com/marckhouzam/tanzu-cli/blob/feat/docMultiPackage/docs/quickstart/install.md

### Release note

```release-note
Document how to install older versions of the Tanzu CLI.
```

### DIFF
--- a/docs/quickstart/install.md
+++ b/docs/quickstart/install.md
@@ -40,6 +40,19 @@ brew update
 brew install vmware-tanzu/tanzu/tanzu-cli
 ```
 
+At the time of writing, Homebrew did not officially support installing older
+versions of a formula, however the following approach works:
+
+```console
+brew tap-new local/tap
+brew tap local/tap
+brew extract --version=0.90.1 vmware-tanzu/tanzu/tanzu-cli local/tap
+brew install tanzu-cli@0.90.1
+
+# To uninstall such an installation
+brew uninstall tanzu-cli@0.90.1
+```
+
 ### Chocolatey (Windows)
 
 ```console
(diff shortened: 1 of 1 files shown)


docs/quickstart/install.md
```diff
@@ -49,22 +62,37 @@ choco install tanzu-cli
...
 
 ```console
-sudo apt-get update
-sudo apt-get install -y ca-certificates curl gpg
+sudo apt update
+sudo apt install -y ca-certificates curl gpg
 curl -fsSL https://packages.vmware.com/tools/keys/VMWARE-PACKAGING-GPG-RSA-KEY.pub | sudo gpg --dearmor -o /etc/apt/keyrings/tanzu-archive-keyring.gpg
 echo "deb [signed-by=/etc/apt/keyrings/tanzu-archive-keyring.gpg] https://storage.googleapis.com/tanzu-cli-os-packages/apt tanzu-cli-jessie main" | sudo tee /etc/apt/sources.list.d/tanzu.list
-sudo apt-get update
-sudo apt-get install -y tanzu-cli
```
anujc25, line 67: Was this a bug in the existing documentation?  Or is it that both `apt` and `apt-get` work?

anujc25 (approved):
Added one comment to confirm the existing documentation. 
But otherwise LGTM.

docs/quickstart/install.md
marckhouzam: `apt` is the new version of `apt-get` as far as I can tell.
I had documented `apt-get` because I thought it was used probably more, but with this PR I found that `apt list` does not exist for `apt-get`, so instead of using both in the doc, I switched everything to `apt`

PR #380, opened 2023-06-28
Rename the --local flag to --local-source

marckhouzam:
### What this PR does / why we need it

This is a UX improvement that was recommended by our UX team.
It renames the `--local` flag to `--local-source` for both `tanzu plugin install` and `tanzu plugin search` in a backwards compatible fashion.

For backwards-compatibility the "--local" form is kept but is marked deprecated.

Further the "--local-source" flag is now hidden for "plugin search". Since the same flag is hidden for "plugin install" because that scenario is not supported in production, I felt the same approach should be taken for "plugin search".

### Which issue(s) this PR fixes

Fixes # N/A

### Describe testing done for PR

```
# Notice that both `--local` and `--local-source` are hidden in the help
$ tz plugin install -h
Install a specific plugin by name or specify all to install all plugins of a group

Usage:
tanzu plugin install [PLUGIN_NAME] [flags]

Examples:

    # Install all plugins of the vmware-tkg/default plugin group version v2.1.0
    tanzu plugin install --group vmware-tkg/default:v2.1.0

    # Install all plugins of the latest version of the vmware-tkg/default plugin group
    tanzu plugin install --group vmware-tkg/default

    # Install the latest version of plugin "myPlugin"
	# If the plugin exists for more than one target, an error will be thrown
    tanzu plugin install myPlugin

    # Install the latest version of plugin "myPlugin" for target kubernetes
    tanzu plugin install myPlugin --target k8s

    # Install version v1.0.0 of plugin "myPlugin"
    tanzu plugin install myPlugin --version v1.0.0

Flags:
      --group string     install the plugins specified by a plugin-group version
  -h, --help             help for install
  -t, --target string    target of the plugin (kubernetes[k8s]/mission-control[tmc]/global)
  -v, --version string   version of the plugin (default "latest")

# Notice that both `--local` and `--local-source` are hidden in the help
$ tz plugin search -h
Search provides the ability to search for plugins that can be installed.
The command lists all plugins currently available for installation.
The search command also provides flags to limit the scope of the search.

Usage:
tanzu plugin search [flags]

Flags:
  -h, --help            help for search
  -n, --name string     limit the search to plugins with the specified name
  -o, --output string   output format (yaml|json|table)
      --show-details    show the details of the specified plugin, including all available versions
  -t, --target string   limit the search to plugins of the specified target (kubernetes[k8s]/mission-control[tmc]/global)

# Test the new flag
$ tz plugin install --local-source artifacts/plugins/darwin/amd64 builder
[...]

PR #408, opened 2023-07-20
Remove --local from "plugin list"

marckhouzam:

This cleanup was agreed upon here: https://github.com/vmware-tanzu/tanzu-cli/pull/380#issuecomment-1643026987

The `--local` flag was kept in the `v0.90` versions for `tanzu plugin list` to help the user transition to use `tanzu plugin search`, but now that we are preparing for `v1.0`, we don't need to have such special handling as the user should be familiar with the fact that `tanzu plugin search --local` is to way to list a local discovery.

So this commit completely remove the `--local` flag from `tanzu plugin list`.

Fixes # N/A

Reminder that before the PR we had:
```
$ /opt/homebrew/bin/tanzu plugin list --local hello
[x] : the '--local' flag does not apply to this command. Please use 'tanzu plugin search --local'

With this PR:
```
$ tz plugin list --local hello
[x] : unknown flag: --local

# Some regression testing for code that got moved
$ tz plugin install --group vmware-tkg/default
[i] Installing plugin 'isolated-cluster:v0.29.0' with target 'global'
[...]
```

### Release note
<!--
     Please add a short text (limit to 1 to 2 sentences if possible) in the release-note block below if
     there is anything in this PR that is worthy of mention in the next release.

     See https://github.com/vmware-tanzu/tanzu-cli/blob/main/docs/release/release-notes.md#does-my-pull-request-need-a-release-note
     for more details.
-->
```release-note
The `--local` flag for `tanzu plugin list` is removed. This flag was previously disabled with a warning to users.
```

### DIFF
--- a/pkg/command/plugin.go
+++ b/pkg/command/plugin.go
@@ -67,22 +67,11 @@ func newPluginCmd() *cobra.Command {
 	discoverySourceCmd := newDiscoverySourceCmd()
 
 	listPluginCmd.Flags().StringVarP(&outputFormat, "output", "o", "", "Output format (yaml|json|table)")
-	listPluginCmd.Flags().StringVarP(&local, "local", "l", "", "path to local plugin source")
-	if !config.IsFeatureActivated(constants.FeatureDisableCentralRepositoryForTesting) {
-		// The --local flag no longer applies to the "list" command.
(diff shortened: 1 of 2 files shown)


anujc25 (approved):
LGTM.

PR #414, opened 2023-07-24
Add an e2e tests using a sha

marckhouzam:

This PR adds a new `plugin-with-sha` plugin to the local test central repos.
It also has the E2E tests install that plugin to confirm installation using a sha reference works in the normal scenario but also in an airgap scenario.

These tests are meant to make sure we can move to using shas in the future, if we choose too.

Fixes # N/A

I ran `make start-test-central-repo` and checked that the following repos all had a `plugin-with-sha` plugin that used a sha in the DB:
1. localhost:9876/tanzu-cli/plugins/central:small
2. localhost:9876/tanzu-cli/plugins/central:large
3. localhost:9876/tanzu-cli/plugins/airgapped:small
4. localhost:9876/tanzu-cli/plugins/airgapped:large

Then, since this PR only adds E2E tests, I verified that CI passed and checked that the new tests have been properly run by looking at the logs.

### Release note

```release-note
E2E test of plugin installation using a sha reference have been added.
```


marckhouzam:
This is ready for review

anujc25 (approved):
LGTM. Thanks for adding this.

PR #416, opened 2023-07-24
Update the max number of telemetry data rows in the database

prkalle:
This PR updates the max number of telemetry data rows that would be collected in local database before telemetry collection pauses collecting the telemetry data.

Summary:
 The value of max rows was updated from 1k to 10k and the profiling shows the average DB file size is around 5MB and that should be fine on any client machine. This max value is updated to cover the air-gapped environment case where user may choose to send/export the telemetry data collected in the local DB though other means.

Fixes #

All CI and unit test passed and the below is the profiling data

Inserted the around 10k rows into the DB and the size is around 5MB on mac.

```
❯ ~/temp/10k_insert.sh
-- Loading resources from /Users/pkalle/.sqliterc
-- Loading resources from /Users/pkalle/.sqliterc
-- Loading resources from /Users/pkalle/.sqliterc
-- Loading resources from /Users/pkalle/.sqliterc
-- Loading resources from /Users/pkalle/.sqliterc
-- Loading resources from /Users/pkalle/.sqliterc
-- Loading resources from /Users/pkalle/.sqliterc
<SNAP>

select count(*) from tanzu_cli_operations;
count(*)
10023
^C
❯ ls -lah ~/.config/tanzu-cli-telemetry/cli_metrics.db
-rw-r--r--  1 pkalle  staff   4.4M Jul 24 23:03 /Users/pkalle/.config/tanzu-cli-telemetry/cli_metrics.db
```

### Release note

```release-note

```

### DIFF
--- a/pkg/telemetry/sqlite_metrics_db.go
+++ b/pkg/telemetry/sqlite_metrics_db.go
@@ -29,7 +29,7 @@ const (
 	SQliteDBFileName = "cli_metrics.db"
 
 	// TanzuCLITelemetryMaxRowCount Max metric instances to be accumulated before pausing the collection
-	TanzuCLITelemetryMaxRowCount = 1000
+	TanzuCLITelemetryMaxRowCount = 10000
 
 	// cliOperationMetricRowClause is the SELECT section of the SQL query to be used when querying the Metric DB row count.
 	cliOperationMetricRowClause = "SELECT count(*) FROM tanzu_cli_operations"


anujc25 (approved):
LGTM

marckhouzam (approved):
LGTM

One thing I was wondering for the future.  When the DB is full, should we replace the oldest entry with a new one? So as to keep the newest telemetry in the DB?

PR #419, opened 2023-07-24
Allow linux plugins to be built with different CGO_ENABLED configuration using the builder plugins

### LINKED ISSUE #418: Builder plugin does not support CGO_ENABLED=1 for linux builds
**Bug description**

It seems like the `builder` plugin does not support CGO_ENABLED=1 for Linux builds.

This is because of the hard coding done in the builder plugin code to use CGO_ENABLED=0 for all Linux builds.

**Expected behavior**
Builder Plugin should provide flexibility in this configuration to allow plugin developers to build the plugin binary with CGO_ENABLED=1 for Linux.

**Output of `tanzu version`**
v0.90.1 (builder plugin)

anujc25:
* It seems like the builder plugin is hard-coding to use `CGO_ENABLED=0` while building a plugin for Linux.
* This change removes the hard-coded configuration which allows plugin developers to pass this as an environment variable as needed.

Fixes #418

* Build plugins with `CGO_ENABLED=1` and without configuring `CGO_ENABLED`.

### Release note

```release-note
Allow Linux plugins to be built with different `CGO_ENABLED` configurations while building plugins
```

### DIFF
--- a/cmd/plugin/builder/command/cli_compile.go
+++ b/cmd/plugin/builder/command/cli_compile.go
@@ -334,7 +334,7 @@ var archMap = map[cli.Arch]targetBuilder{
 	cli.Linux386: func(pluginName, outPath string) target {
 		return target{
 			env: []string{
-				"CGO_ENABLED=0",
+				CGOEnabled(),
 				"GOARCH=386",
 				"GOOS=linux",
 			},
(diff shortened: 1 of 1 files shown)


zach96guan (approved):
LGTM. Thanks for the fix!

marckhouzam (commented):
I don’t understand the impact of having`CGO_ENABLED=0` by default or not. I’ve only had time to do a very quick search on the Internet and I’ve seen that it’s more portable or something. I’m not sure if that applies to us so much.

But to play it safe since this is late in the cycle. How about we check if the variable is set in the environment and only if it’s not then do we set it to zero like we do currently.

marckhouzam (approved):
LGTM
Thanks!

PR #420, opened 2023-07-25
Update installation documentation for pre-releases

marckhouzam:

After discussion it was agreed that we would document the way to install pre-releases using package managers.
This commit updates the doc accordingly.
It also replaces the terminology "install an older version" with "install a specific version".

Fixes # N/A

Read through the doc.

### Release note

```release-note
Document how to install pre-releases using package managers.
```


anujc25 (approved):
LGTM. Thanks

PR #425, opened 2023-07-25
Update the telemetry send metrics timeout

prkalle:
This PR updates the telemetry send metrics operation timeout from 1 sec to 2 sec
While testing the telemetry on VPN from different country it was observed 1sec timeout is not sufficient to send the metrics to supercollider

Fixes #

CI check and unit tests passed

### Release note

```release-note

```


marckhouzam (approved):
LGTM.  Thanks @prkalle

PR #427, opened 2023-07-25
Fix the last tag generation in release yaml

mpanchajanya:

When tagging a major release, the release notes workflow did not work properly.
Example:
tag 1.0.0 would cause a comparison with a release tag of 0.0.0 which did not exist
tag 2.0.0 would cause a comparison with 1.0.0 which would ignore any 1.x.y releases.

This change looks for the last tag of the previous release.
So 1.0.0 would find 0.90.1 since it is our last release.

Fixes #

### Release note

```release-note

```


marckhouzam (approved):
LGTM as long as CI passes

PR #442, opened 2023-07-29
Mutually exclusive --local-source flags

marckhouzam:
**Only the top commit is part of this PR.  The first commit is part of #440 **

### What this PR does / why we need it

Running a "tanzu install" with both the "-local-source" and "-group" flag does not throw an error.

### Which issue(s) this PR fixes

Fixes #441

### Describe testing done for PR

```
$ tz plugin install  --group vmware-tkg/default --local-source artifacts/plugins/darwin/amd64
[x] : if any flags in the group [group local-source] are set none of the others can be; [group local-source] were all set

$ tz plugin install  --group vmware-tkg/default --local artifacts/plugins/darwin/amd64
Flag --local has been deprecated, this was done in the v1.0.0 release, it will be removed following the deprecation policy (6 months). Use the --local-source flag instead.

[x] : if any flags in the group [group local] are set none of the others can be; [group local] were all set

$ tz plugin search --local-source artifacts/plugins/darwin/amd64 --show-details
[x] : if any flags in the group [local-source show-details] are set none of the others can be; [local-source show-details] were all set

$ tz plugin search --local-source artifacts/plugins/darwin/amd64 --name builder
[x] : if any flags in the group [local-source name] are set none of the others can be; [local-source name] were all set

$ tz plugin search --local-source artifacts/plugins/darwin/amd64 --target tmc
[x] : if any flags in the group [local-source target] are set none of the others can be; [local-source target] were all set

$ tz plugin search --local-source artifacts/plugins/darwin/amd64
  NAME     DESCRIPTION             TARGET  LATEST
  builder  Build Tanzu components  global  v1.0.0-dev
  test     Test the CLI            global  v1.0.0-dev
```

### Release note

```release-note
Mark the `--group` and `--local-source` flags as mutually exclusive 
```

PR #434, opened 2023-07-28
Use --local-source to locally installing a plugin

marckhouzam:
### What this PR does / why we need it

The previously used --local flag has been deprecated.

### Which issue(s) this PR fixes

Fixes # N/A

### Describe testing done for PR

Before this PR:
```
$ cd tanzu-cli
$ export PLUGIN_NAME=builder
$ make plugin-build-install-local
/Users/kmarc/git/tanzu-cli/bin/builder plugin build \
		--path ./cmd/plugin \
		--binary-artifacts /Users/kmarc/git/tanzu-cli/artifacts/plugins \
		--version v1.0.0-dev \
		--ldflags "-X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Date=2023-07-28' -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.SHA=26c97e72f' -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v1.0.0-dev'" \
		--goflags "" \
		--os-arch darwin_amd64 \
		--match "builder" \
		--plugin-scope-association-file ./cmd/plugin/plugin-scope-association.yaml
2023-07-28T10:30:57-04:00 [i] building local repository at /Users/kmarc/git/tanzu-cli/artifacts/plugins, v1.0.0-dev, [darwin_amd64]
2023-07-28T10:30:57-04:00 [i] 🐼 - building plugin at path "cmd/plugin/builder"
2023-07-28T10:31:02-04:00 [i] 🐼 - $ /usr/local/go/bin/go build -o /Users/kmarc/git/tanzu-cli/artifacts/plugins/darwin/amd64/global/builder/v1.0.0-dev/tanzu-builder-darwin_amd64 -ldflags -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Date=2023-07-28' -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.SHA=26c97e72f' -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v1.0.0-dev' -X 'github.com/vmware-tanzu/tanzu-plugin-runtime/plugin/buildinfo.Version=v1.0.0-dev' -tags  ./cmd/plugin/builder
2023-07-28T10:31:04-04:00 [i] ========
2023-07-28T10:31:04-04:00 [i] saving plugin group manifest...
[...]

PR #448, opened 2023-07-31
[release 1.0]: Cherry picks from main

marckhouzam:

Cherry-picks: #434 #442 #414

I originally was not going to cherry-pick #414, but then I thought that having those tests on the release-1.0 branch will help avoid breaking that scenario for a future 1.0.x patch release.

Did a `git diff main` and confirmed the two branches were the same.

Let CI run

### Release note

```release-note
Update make target to use the `--local-source` flag instead of `--local` as well as mark the `--local-source` flags as mutually exclusive with some other flags.
```


anujc25 (approved):
LGTM.

PR #472, opened 2023-08-24
Fix Catalog corruption issue when running CLI in parallel

### LINKED ISSUE #471: Running different tanzu cli commands concurrently can corrupt the catalog cache
**Bug description**

In certain circumstances, running different tanzu cli commands concurrently can corrupt the catalog cache.

The CLI reads and writes from a `catalog.yaml` file which is shared between different `tanzu` invocation. At the moment there is no inter-process concurrency logic protecting this file from concurrent `tanzu` invocations and we have seen the file become corrupted, leading to plugins not being accessible.

The problem happens when the new "telemetry/global" and the old "telemetry/k8s" plugins are installed at the same time AND that the CLI is invoked concurrently.

The code that causes the problem is the following:
https://github.com/vmware-tanzu/tanzu-cli/blob/334e35d8f0054050a63d36dfdda1877c37304aa6/pkg/command/root.go#L120

which eventually always writes the `catalog.yaml` file, and without concurrency protection, ends up corrupting it.

Although we haven't reproduced it, other situations could cause similar concurrency problems for the catalog such as:
1. concurrent plugin install commands
1. concurrent plugin delete commands
1. concurrent plugin clean commands (more rare use case)

**Expected behavior**

Running concurrent calls to different commands of the `tanzu` binary should not cause corruption in the `catalog.yaml` file.

```
# Start with no plugins installed
# The essential plugins are automatically installed (that "telemetry" plugin must be installed to trigger the bug)
$ tanzu plugin list
[i] The tanzu cli essential plugins have not been installed and are being installed now. The install may take a few seconds.

Standalone Plugins

# Now install the other "telemetry" plugin (required to trigger the bug)
$ tanzu plugin install telemetry --target k8s
[i] Installing plugin 'telemetry:v0.30.1' with target 'kubernetes'
[i] Plugin binary for 'telemetry:v0.30.1' found in cache
[ok] successfully installed 'telemetry' plugin

# Also install another stand-alone plugin (not required, but shows the impact better)
$ tanzu plugin install management-cluster --target k8s
[i] Installing plugin 'management-cluster:v0.30.1' with target 'kubernetes'
[i] Plugin binary for 'management-cluster:v0.30.1' found in cache
[ok] successfully installed 'management-cluster' plugin
$ tanzu plugin list
Standalone Plugins
  NAME                DESCRIPTION                                                 TARGET      VERSION  STATUS
  telemetry           configure cluster-wide settings for vmware tanzu telemetry  global      v1.1.0   installed
  management-cluster  Kubernetes management cluster operations                    kubernetes  v0.30.1  installed
  telemetry           configure cluster-wide settings for vmware tanzu telemetry  kubernetes  v0.30.1  installed

# run "tanzu version" 100 times concurrently (by running it in the background)
$ for i in {1..50}; do
tanzu version>/dev/null&
done

[3] 43972
[4] 43973
[...][101] 44070
[102] 44071
$
[4]    43973 done       tanzu version > /dev/null
[8]    43977 done       tanzu version > /dev/null
[...]
[54]  + 44023 done       tanzu version > /dev/null
[10]  + 43979 done       tanzu version > /dev/null

# Notice that the "catalog.yaml" file is now corrupt and no longer lists plugins
$ cat ~/.cache/tanzu/catalog.yaml
serverPlugins:
    tkg1: {}

# Notice that the "telemetry" plugin is gone so it is being installed again automatically
# and that the "telemetry" plugin for target "kubernetes" and the "management-cluster
# plugins are also gone.
$ tanzu plugin list
[i] The tanzu cli essential plugins have not been installed and are being installed now. The install may take a few seconds.

Standalone Plugins
```

**Output of `tanzu version`**

```
$ tanzu version
version: v1.0.0-dev
buildDate: 2023-08-21
sha: 334e35d8f
```

Locally

anujc25:

* Fix the Catalog corruption issue when running CLI in parallel
* Implements `NewContextCatalog` (for the reading) and `NewContextCatalogUpdater` (for reading/writing) as separate APIs.
* Uses a new `lockedfile` API to lock the file. Inspired from: https://go.googlesource.com/proposal/+/master/design/33974-add-public-lockedfile-pkg.md

Fixes #471

* Wrote a bash script to manually verify the fix
```
#!/bin/bash

TANZU=tanzu
$TANZU plugin install --group vmware-tkg/default
for i in {1..10}
do
    for x in {1..50}
    do
        $TANZU version &
        echo "running: $x $!"        
    done
    echo "waiting"
    wait
    echo "done"    
done

$TANZU plugin list
```

* Also added E2E tests to run tanzu commands in parallel when 2 telemetry plugins are present
* Also added E2E tests to install multiple Plugins in parallel

* User can verify the E2E test by running tests with latest `tanzu` binary with the fix and the old `tanzu binary without the fix as below:
```
cd test/e2e/

export TANZU_CLI_E2E_TEST_BINARY_PATH=tanzu-old ## Without the fix
make e2e-catalog-tests. ## This should fail


export TANZU_CLI_E2E_TEST_BINARY_PATH=tanzu-new ## With the fix
make e2e-catalog-tests ## This should be successful

```

### Release note

```release-note
Fix the Catalog corruption issue (missing installed plugins) when running CLI in parallel
```


pkg/pluginmanager/manager.go
```diff
@@ -1230,6 +1231,7 @@ func doDeletePluginFromCatalog(pluginName string, target configtypes.Target, cat
 		if err != nil {
 			return fmt.Errorf("plugin %q could not be deleted from cache", pluginName)
 		}
+		c.Unlock()
```
marckhouzam, line 1234: I think we should call the `c.Unlock()` right after the delete, even if there is an error.  The catalog file is locked even if the delete fails.

pkg/pluginmanager/manager.go
```diff
@@ -1104,13 +1104,14 @@ func doInstallTestPlugin(p *discovery.Discovered, pluginPath, version string) er
...
 func updatePluginInfoAndInitializePlugin(p *discovery.Discovered, plugin *cli.PluginInfo) error {
-	c, err := catalog.NewContextCatalog(p.ContextName)
+	c, err := catalog.NewContextCatalogUpdater(p.ContextName)
 	if err != nil {
 		return err
 	}
 	if err := c.Upsert(plugin); err != nil {
 		log.Info("Plugin Info could not be updated in cache")
 	}
+	c.Unlock()
```
marckhouzam, line 1114: I think the `c.Unlock()` should be called even if the `Upsert` fails.
Maybe using a `defer` would be simplest, right after the catalog is obtained successfully.

marckhouzam (commented):
This look very nice to me.
There are, I think, some important comments to address, but this is a very clean solution.
Thanks @anujc25 !

pkg/catalog/catalog.go
```diff
@@ -77,25 +92,16 @@ func (c *ContextCatalog) Upsert(plugin *cli.PluginInfo) error {
...
 	if plugin.Target == configtypes.TargetGlobal || plugin.Target == configtypes.TargetK8s {
-		c.deleteOldTargetEntries(PluginNameTarget(plugin.Name, configtypes.TargetUnknown))
+		delete(c.plugins, PluginNameTarget(plugin.Name, configtypes.TargetUnknown))
 	} else if plugin.Target == configtypes.TargetUnknown {
 		// This could be a global plugin or a k8s plugin (but not both), so let's
 		// delete either pre-existing entry from the catalog
-		c.deleteOldTargetEntries(PluginNameTarget(plugin.Name, configtypes.TargetGlobal))
-		c.deleteOldTargetEntries(PluginNameTarget(plugin.Name, configtypes.TargetK8s))
+		delete(c.plugins, PluginNameTarget(plugin.Name, configtypes.TargetUnknown))
+		delete(c.plugins, PluginNameTarget(plugin.Name, configtypes.TargetUnknown))
```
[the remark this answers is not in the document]
anujc25: I noticed that we were using `deleteOldTargetEntries` to delete the old target entries.
```
func (c *ContextCatalog) deleteOldTargetEntries(key string) {
	if oldInstallationPath, exists := c.plugins[key]; exists {
		delete(c.sharedCatalog.IndexByPath, oldInstallationPath)
		delete(c.plugins, key)
		delete(c.sharedCatalog.IndexByName, key)
	}
}
```

Now, this function deletes the entry from the shared catalog that it operates on with `delete(c.plugins, key)`.
However, deleting the entries from `IndexByPath` and `IndexByName` should not be done during this step.
```
    delete(c.sharedCatalog.IndexByPath, oldInstallationPath)
    delete(c.sharedCatalog.IndexByName, key)
```
Because we are trying to delete the plugin entry from a specific context catalog and it is possible that the old shared catalog (meaning other existing contexts)  still refers to the pluginInstallionPath that we are trying to delete here.
So, we cannot delete the entry from `IndexByPath` and `IndexByName` at this step. If needed, we should run periodic checks and cleanup unused entries from the catalog cache under `IndexByPath` and `IndexByName`.

pkg/catalog/cleanup.go
```diff
@@ -13,40 +13,34 @@ import (
...
+
 			// The "unknown" target was previously used in two scenarios:
 			// 1- to represent the global target (>= v0.28 and < v0.90)
 			// 2- to represent either the global or kubernetes target (< v0.28)
 			// If we have a plugin with the "global" or "k8s" target we should remove any similar plugin using
 			// the "unknown" target.
-			if plugins[i].Target == configtypes.TargetGlobal || plugins[i].Target == configtypes.TargetK8s {
-				c.deleteOldTargetEntries(PluginNameTarget(plugins[i].Name, configtypes.TargetUnknown))
+			if pluginInfo.Target == configtypes.TargetGlobal || pluginInfo.Target == configtypes.TargetK8s {
+				delete(c.StandAlonePlugins, PluginNameTarget(pluginInfo.Name, configtypes.TargetUnknown))
```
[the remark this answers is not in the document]
anujc25: > Why is this changed to only delete the StandAlonePlugins section?
I'm not 100% but couldn't there be context-scoped plugins that have an entry using the old empty string to represent a k8s target?

Good catch. I should be using `pluginAssociations[i]` instead of `c.StandAlonePlugins`.


> Also, don't we also want to cleanup the IndexByPath and IndexByName sections as we used to do?
I think deleting IndexByPath and IndexByName can be considered a bug to me. I tried to explain the rationale behind that [here](https://github.com/vmware-tanzu/tanzu-cli/pull/472/files#r1308871509)

pkg/catalog/catalog.go
```diff
@@ -161,21 +171,43 @@ func newSharedCatalog() (*Catalog, error) {
...
+		lockedFile, err = lockedfile.Edit(getCatalogCachePath())
+		if err != nil {
+			return nil, nil, err
+		}
+		unlock = func() {
+			if lockedFile != nil {
+				lockedFile.Close()
+			}
+		}
+		b, err = os.ReadFile(getCatalogCachePath())
```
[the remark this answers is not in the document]
anujc25: Makes sense. I have replaced `os.ReadFile(getCatalogCachePath())` with `io.ReadAll(lockedFile)` which should use the lockedFile interface to read the file.

vuil:
Do we have any e2e or integ tests (or doing multiple concurrent catalog reads/writes) that can help uncover the issue we are trying to fix here? Having them earlier would have help catch the issue, having them would have helped giving confidence that the issue is sufficiently addressed. Let's make sure we at least file a TODO to provide them.

pkg/pluginmanager/manager.go
vuil: It might be good to paraphrase the above as a comment for the benefit of other readers of the code.

pkg/catalog/catalog.go
```diff
@@ -161,21 +172,43 @@ func newSharedCatalog() (*Catalog, error) {
...
+				lockedFile.Close()
+			}
+		}
+		b, err = io.ReadAll(lockedFile)
+	} else {
+		b, err = lockedfile.Read(getCatalogCachePath())
+	}
+
 	if err != nil {
 		catalog, err = newSharedCatalog()
```
vuil, line 201: It looks like that could still be a potential race if multiple callers enter the newSharedCatalog case (because there is no write lock acquired by either)

anujc25:
> Do we have any e2e or integ tests (or doing multiple concurrent catalog reads/writes) that can help uncover the issue we are trying to fix here? Having them earlier would have help catch the issue, having them would have helped giving confidence that the issue is sufficiently addressed. Let's make sure we at least file a TODO to provide them.

I am planning to add e2e tests for this. I wanted to get high-level feedback on the approach before I deep dive into writing tests for this PR and hence a draft PR. If the approach looks good, I will add some e2e tests for this change.

pkg/catalog/catalog.go
anujc25: It shouldn't be a problem because we will lock the file at `L185` or `L197`.

Note: `newSharedCatalog` creates a new in-memory catalog object and does not create a new file.

vuil:
> > Do we have any e2e or integ tests (or doing multiple concurrent catalog reads/writes) that can help uncover the issue we are trying to fix here? Having them earlier would have help catch the issue, having them would have helped giving confidence that the issue is sufficiently addressed. Let's make sure we at least file a TODO to provide them.
> 
> I am planning to add e2e tests for this. I wanted to get high-level feedback on the approach before I deep dive into writing tests for this PR and hence a draft PR. If the approach looks good, I will add some e2e tests for this change.

sgtm. The not-so-great part is an expectation that any valid unlock function returned has to be called by the caller, but the approach is a reasonable compromised.

pkg/catalog/catalog.go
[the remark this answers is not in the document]
marckhouzam: Ah yes, good catch!  I understand now.  Thanks for the explanation.

pkg/catalog/catalog.go
```diff
@@ -77,25 +93,16 @@ func (c *ContextCatalog) Upsert(plugin *cli.PluginInfo) error {
...
 	// the "unknown" target and vice versa.
 	if plugin.Target == configtypes.TargetGlobal || plugin.Target == configtypes.TargetK8s {
-		c.deleteOldTargetEntries(PluginNameTarget(plugin.Name, configtypes.TargetUnknown))
+		delete(c.plugins, PluginNameTarget(plugin.Name, configtypes.TargetUnknown))
 	} else if plugin.Target == configtypes.TargetUnknown {
 		// This could be a global plugin or a k8s plugin (but not both), so let's
 		// delete either pre-existing entry from the catalog
-		c.deleteOldTargetEntries(PluginNameTarget(plugin.Name, configtypes.TargetGlobal))
-		c.deleteOldTargetEntries(PluginNameTarget(plugin.Name, configtypes.TargetK8s))
+		delete(c.plugins, PluginNameTarget(plugin.Name, configtypes.TargetUnknown))
```
marckhouzam, line 100: This line and the one below are the same.
I believe these two lines should not use `configtypes.TargetUnknown` but should use the `configtypes.TargetGlobal` and `configtypes.TargetK8s` like before.

I had to think hard to remember which case this addressed.  Maybe you want to add a better comment here.  Something like:

```
// An older plugin binary may not specify a target (through its 'info' command).
// Therefore the plugin could be a global plugin or a k8s plugin (but not both).
// We need to delete either pre-existing entries from the catalog to avoid having
// a double entry.
```

anujc25: Thanks for catching this. Fixed.

pkg/catalog/catalog.go
```diff
@@ -161,21 +174,49 @@ func newSharedCatalog() (*Catalog, error) {
...
+
+	if setWriteLock {
+		var lockedFile *lockedfile.File
+		if utils.PathExists(getCatalogCachePath()) {
+			lockedFile, err = lockedfile.Edit(getCatalogCachePath())
+		} else {
+			// Create directory path if missing before locking the file
+			_ = os.MkdirAll(filepath.Dir(getCatalogCachePath()), 0777)
+			lockedFile, err = lockedfile.Create(getCatalogCachePath())
+		}
```
anujc25, line 194: Please notice this latest change where I am using 2 different functions based on the existence of the catalog file. cc @marckhouzam @vuil

marckhouzam: Oh, nice find! 
Did the test fail, or did you just notice by looking at the code?  I'm asking because I can't find a way to cause a problem before this change

pkg/catalog/catalog.go
```diff
@@ -161,21 +174,49 @@ func newSharedCatalog() (*Catalog, error) {
...
+		if err != nil {
+			return nil, nil, err
+		}
+		unlock = func() {
+			if lockedFile != nil {
+				lockedFile.Close()
+			}
+		}
+		b, err = io.ReadAll(lockedFile)
```
marckhouzam, line 203: If this fail with an error, we should not fall out and create a `newSharedCatalog()`, since it will not be locked.
An error here is a real error and we should return.

pkg/catalog/catalog.go
anujc25: One of the unit tests failed which was testing the catalog update when no file or directory existed.

marckhouzam (commented):
Nice effort.  This is complicated stuff.
I haven't looked at the E2E tests yet because more concurrency questions came to my mind.

I'm wondering, in `saveCatalogCache()` do we need to use the `lockedFile` file descriptor to write to the file?  I think doing that would make unit tests proper fail if `Unlock()` is called too early (which we had before but you corrected in your latest changes), because a call to `Upsert()` after the `Unlock()` would not be able to write to the closed file.

pkg/catalog/catalog.go
anujc25: I did not get what you mean by `since it will not be locked`? The lock is acquired when we do `lockedFile, err = lockedfile.Edit(getCatalogCachePath())` right?

So, if the `ReadAll` fails the lock is already acquired and we will create a new sharedCatalog.

This is no different from the below line where we do `lockedfile.Read(getCatalogCachePath())` which internally does
```
Open()
io.ReadAll(lockedFile)
Close()
```

marckhouzam: You are right. My comment about “not being locked” doesn’t make sense. 

The risk however is that if the readAll fails, we’ll overwrite the existing catalog with a new empty one.  But I don’t know what could make the read fail…

But you are right that it’s the same as the line `lockedfile.Read(getCatalogCachePath())`.  The code assumes that if there is an error with the read it’s because the file doesn’t exist and therefore we create a new empty catalog. But what if the read fails for some other reason?  Maybe this could happen with a lock timeout?

pkg/catalog/catalog.go
```diff
@@ -77,25 +93,18 @@ func (c *ContextCatalog) Upsert(plugin *cli.PluginInfo) error {
...
-		c.deleteOldTargetEntries(PluginNameTarget(plugin.Name, configtypes.TargetK8s))
+		// An older plugin binary may not specify a target (through its 'info' command).
+		// Therefore the plugin could be a global plugin or a k8s plugin (but not both).
+		// We need to delete either pre-existing entries from the catalog to avoid having
+		// a double entry.
+		delete(c.plugins, PluginNameTarget(plugin.Name, configtypes.TargetGlobal))
+		delete(c.plugins, PluginNameTarget(plugin.Name, configtypes.TargetK8s))
 	}
-
 	return saveCatalogCache(c.sharedCatalog)
```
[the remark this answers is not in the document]
anujc25: > I was wondering if we could save the catalog as part of the created unlock() function?
I'm thinking that that is when we know that the caller is done modifying the catalog.

Thanks for the suggestion. I have updated the code to throw some meaningful errors when this happens. Also added some unit tests to capture this.

pkg/catalog/catalog.go
```diff
@@ -131,10 +140,14 @@ func (c *ContextCatalog) Delete(plugin string) error {
 	if ok {
 		delete(c.plugins, plugin)
 	}
-
 	return saveCatalogCache(c.sharedCatalog)
 }
 
+// Unlock unlocks the catalog for other process to read/write
```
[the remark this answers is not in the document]
anujc25: > We could even code a protection for this by setting a boolean showing the catalog has been unlocked, and then failing any operation on it. Something to discuss.

Thanks for the suggestion. I have updated the code to throw some meaningful errors when this happens. Also added some unit tests to capture this.

pkg/catalog/catalog.go
anujc25: > But you are right that it’s the same as the line lockedfile.Read(getCatalogCachePath()). The code assumes that if there is an error with the read it’s because the file doesn’t exist and therefore we create a new empty catalog. But what if the read fails for some other reason? Maybe this could happen with a lock timeout?

Yeah makes sense. I have updated the code to check the type of error and only create the `newSharedCatalog` if `os.IsNotExist(err)` is `true`.

marckhouzam (approved):
LGTM
Thanks for this complicated and important change!

pkg/catalog/catalog.go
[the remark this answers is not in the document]
anujc25: Currently, we do not have a place in the code where we create a `contextCatalog` and then use the same catalog to update operations periodically. But if we do that, we would like to ensure that the file persists as soon as possible because subsequent operations can fail or the user can terminate the process in between. It is always good to persist the file as soon as possible to get consistent behavior.

Also, if the `Update`/`Delete` fails while writing to the file we would like to know it as soon as possible during that function call to return a meaningful error message. If we only write during `Unlock` it would be hard to show meaningful error message to the user as we do not know which operation failed actually.

vuil (approved):
Thanks for the updates and improvements after the reviews.
There is a nit a typo in the comments, but changes lgtm.

PR #476, opened 2023-09-11
Update GO v1.21

mpanchajanya:

 - Update go.mod to point to go v1.21
 - Update tools like linters to latest
 - Fix lint issues
 
Fixes #

 - Build the Tanzu CLI
 - Perform basic lifecycle ops
 - Unit tests, Lint checks
 
### Release note

```release-note
Tanzu CLI is updated to latest go v1.21
```


.golangci.yaml
```diff
@@ -102,49 +154,43 @@ linters:
...
       linters:
         - staticcheck
-      text: "SA1019:"
-
+      text: 'SA1019:'
   include:
-    - EXC0011 # disable excluding of issues about missing package comments from stylecheck
-
-  # Show all errors for all linters. Setting these to 0 disables limiting error reporting.
+    - EXC0011
```
anujc25, line 187: Can we keep the comment as it is.
```suggestion
    - EXC0011 # disable excluding of issues about missing package comments from stylecheck
```

.golangci.yaml
```diff
@@ -75,14 +131,10 @@ linters:
     - typecheck
     - unconvert
     - unused
-    - varcheck
     - whitespace
     - bodyclose
     - noctx
     - rowserrcheck
-    - structcheck
-    - unparam
```
anujc25, line 84: Are these removals intentional?

pkg/auth/csp/selfmanaged.go
```diff
@@ -96,6 +96,7 @@ func runLocalListener() error {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/callback", callbackHandler)
 	tokenExchange, tokenExchangeComplete = context.WithCancel(context.TODO())
+	//nolint:gosec
```
anujc25, line 99: If possible might be better to add a comment on why we are skipping this.

pkg/registry/client.go
```diff
@@ -104,7 +104,7 @@ func getFilesFromLayer(imgLayer regv1.Layer) (map[string][]byte, error) {
 			}
 			return files, err
 		}
-		if hdr.Typeflag == tar.TypeReg || hdr.Typeflag == tar.TypeRegA {
+		if hdr.Typeflag == tar.TypeReg {
```
anujc25, line 107: Is this change intentional?

.golangci.yaml
vuil: same question. If they are, in the future it would be good to mention in the PR description or a comment in this file

.golangci.yaml
```diff
@@ -1,4 +1,63 @@
 linters-settings:
+  depguard:
+    rules:
+      main:
+        files:
+          - $all
+          - "!$test"
+        allow:
+          - $gostd
+          - github.com/AlecAivazis/survey/v2
```
vuil, line 10: this sounds like sounding that is rather cumbersome to maintain. Can you summarize when a pattern needs to be added to the allow list, because it does not seem to cover all of go.mod

.golangci.yaml
mpanchajanya: Yes these are deprecated in new golangci-lint version. Updated the PR description

mpanchajanya:
Addressing the comments as part of https://github.com/vmware-tanzu/tanzu-cli/pull/490

PR #484, opened 2023-09-19
Cleanup pluginmanager of unused feature flag code

marckhouzam:
### What this PR does / why we need it

This commit removes all the unused code around feature flags from the `manager.go` file.  

**Code cleanup details:**
1. Remove `ValidatePlugin()` which was never used.  There is a separate copy of this function in the tanzu-plugin-runtime repository.
2. Remove `discoverPlugins()` and directly use `discoverSpecificPlugins()`
3. Remove `discoverServerPluginsBasedOnCurrentServer` which is no longer used.
4. Include the logic of `discoverServerPluginsBasedOnAllCurrentContexts()` directly into `DiscoverServerPlugins()` which had become a one-liner.
5. Remove the following functions that were no longer used:
          - `AvailablePlugins()`
          - `AvailablePluginsFromLocalSource()`
          - `availablePlugins()`
          - `combineDuplicatePlugins()`
          - `getInstalledButNotDiscoveredStandalonePlugins()`
          - `DiscoveredFromPlugininfo()`
          - `availablePluginsFromStandaloneAndServerPlugins()`
          - `pluginIndexForName()`
          - `legacyPluginInstall()`
          - `GetRecommendedVersionOfPlugin()`

Completely remove the feature flag constant: `FeatureDisableCentralRepositoryForTesting`

**Update the unit tests for the plugin manager.**

This commit also improves our ability to test the pluginmanager.
It does this using two techniques:
1. Creating a test plugin inventory DB in the cache and requesting the inventory code to always use the cache (without checking the digest). To force the use of the cache the commit Introduces a test variable `TANZU_CLI_USE_DB_CACHE_ONLY`.  This approach allows to discover plugins and groups without needing a real OCI registry.

2. Creating "fake" plugin binaries in the plugin binary cache.  This allows installation of plugins to find the binaries in the cache and therefore never have to go to an OCI registry to download them.

With these two techniques, the unit tests can discovery and install plugins, and can also do so using groups.  The plugin sync can also be tested with this.

### Which issue(s) this PR fixes

Part of #477

### Describe testing done for PR

```
# Remove the existing plugin DB
$ \rm -rf ~/.cache/tanzu/plugin_inventory
# Confirm the DB gets downloaded
$ tz plugin search --name apps
[i] Reading plugin inventory for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.
  NAME  DESCRIPTION                 TARGET      LATEST
[...]

PR #486, opened 2023-09-19
Allow installing plugins with `vMAJOR` or `vMAJOR.MINOR` or `vMAJOR.MINOR.PATCH` versions

### LINKED ISSUE #485: Support installing latest minor and patch version of the plugin when vMAJOR or vMAJOR.MINOR is specified
Tanzu CLI supports installing plugins based on the specified plugin version. However, today it only supports installing plugins when the exact version is selected.

This issue proposes the plugin installation UX be enhanced to provide flexibility in installing the latest patch version of a plugin when vMAJOR.MINOR is specified.

Example:
```
tanzu plugin install <plugin-name>  (installs the latest version of the plugin - supported today)
tanzu plugin install <plugin-name> -v vMAJOR (installs the latest minor.patch for the specified major - NOT supported today)
tanzu plugin install <plugin-name> -v vMAJOR.MINOR (installs the latest patch for the specified minor - NOT supported today)
tanzu plugin install <plugin-name> -v vMAJOR.MINOR.PATCH (installs the specific version - supported today)
```

**Additional context**

anujc25:
* Allow installing plugins with `vMAJOR` or `vMAJOR.MINOR` or `vMAJOR.MINOR.PATCH` versions

Fixes #485

* Added Unit Tests and E2E Tests

```
~ $ tz plugin search --show-details --name insight                                                                                                                         
name: insight
description: Post & query image, package, source, and vulnerability data
target: kubernetes
latest: v1.6.3
versions:
    - v1.3.9
    - v1.4.6
    - v1.5.1
    - v1.6.0
    - v1.6.3
    - 
~ $ tz plugin install insight -v v1                                                                                                                                        (base)
[i] Installing plugin 'insight:v1.6.3' with target 'kubernetes'
[ok] successfully installed 'insight' plugin

~ $ tz plugin install insight -v v1.3                                                                                                                                      (base)
[i] Installing plugin 'insight:v1.3.9' with target 'kubernetes'
[ok] successfully installed 'insight' plugin

~ $ tz plugin install insight -v v1.6                                                                                                                                      (base)
[i] Installing plugin 'insight:v1.6.3' with target 'kubernetes'
[i] Plugin binary for 'insight:v1.6.3' found in cache
[ok] successfully installed 'insight' plugin

~ $ tz plugin install insight -v v1.6.3                                                                                                                                    (base)
[i] Installing plugin 'insight:v1.6.3' with target 'kubernetes'
[i] Plugin binary for 'insight:v1.6.3' found in cache
[ok] successfully installed 'insight' plugin

~ $ tz plugin install insight -v v1.5                                                                                                                                      (base)
[i] Installing plugin 'insight:v1.5.1' with target 'kubernetes'
[ok] successfully installed 'insight' plugin

~ $ tz plugin install insight -v v2                                                                                                                                        (base)
[x] : unable to find plugin 'insight' with version 'v2' for target 'kubernetes'

~ $ tz plugin list                                                                                                                                                      (base)
Standalone Plugins
  NAME       DESCRIPTION                                                  TARGET      VERSION     STATUS
  telemetry  configure cluster-wide settings for vmware tanzu telemetry   global      v1.1.0      installed
  insight    post & query image, package, source, and vulnerability data  kubernetes  v1.5.1      installed
```

### Release note

```release-note
Allow installing plugins with `vMAJOR` or `vMAJOR.MINOR` or `vMAJOR.MINOR.PATCH` versions with `tanzu plugin install` and `tanzu plugin sync` commands
```


marckhouzam:
Thanks @anujc25 for this.  While I review I wanted to mention an alternative implementation.  I'm not suggesting using it, I'm just bringing it to your attention.

Currently when we look for a plugin in the DB:
1. the `RecommendedVersion` field returned is set to the highest semantic version found by the query
2. the query will either look for all versions (not filter on version) or will look for a specific version using `WHERE Version=<vX.Y.Z>`

For this feature, we could use SQL's matching functionality and have the version filter become `WHERE Version like '<version>%'`.  Then the query would return all versions that match the specified version, even if it is `vX.Y` or `vX`.  And the latest version matching would be in `RecommendedVersion`.

For example, requesting to install a version of `v1.2` would return the highest version that matches `v1.2`, which means the latest patch of `v1.2`.

Pros:
- should be almost transparent to the code as the sql query does this already

Cons:
- we would be dependent on using the `RecommendedVersion` field so we would need to ignore that value if the DB ever provided one.

marckhouzam:
@anujc25 my apologies but after merging #484 this PR needs a rebase

pkg/plugininventory/sqlite_inventory.go
```diff
@@ -222,7 +222,7 @@ func createPluginWhereClause(filter *PluginInventoryFilter) (string, error) {
 				whereClause = fmt.Sprintf("%s Version=RecommendedVersion AND", whereClause)
 			} else {
 				// We want a specific version of the plugin
-				whereClause = fmt.Sprintf("%s Version='%s' AND", whereClause, filter.Version)
+				whereClause = fmt.Sprintf("%s Version Like '%s%%' AND", whereClause, filter.Version)
```
marckhouzam, line 225: We have to be careful that a version of e.g., `v0.2` doesn’t match e.g.,`v0.28.1` but only matches `v0.2.PATCH`

This type of query probably deserves a comment if you agree.

marckhouzam (commented):
Thanks for the update using SQL!

pkg/plugininventory/sqlite_inventory.go
anujc25: Thanks for catching this. I have updated the query and added a comment explaining this condition.

marckhouzam (commented):
1. For error messages that say `unable to find plugin '%v' with version '%v'` we will now show "short" versions like `v1`.  How about changing the error message to use the word `matching to say `unable to find plugin '%v' matching version '%v'`?

2. Is it ok for the special plugins that have actual binaries with versions `v0.25`/`v0.28`/`v0.29` to now install the latest patch instead of the specific binary with that version?  I hope so, because the behaviour of the PR is what I'm hoping we can keep.

3. I think we need to look at how this feature can be used for context-scoped plugins.
I created a CR for the `cluster` plugin with a version of `v0.30`.
When the plugin is not installed I see version `v0.30`, which is probably ok
```
Plugins from Context:  tkg1
  NAME                DESCRIPTION                           TARGET      VERSION  STATUS
  cluster             Kubernetes cluster operations         kubernetes  v0.30    not installed
```

when I do a plugin sync I then see `v0.30.1` which is ok, but is shows there is an update available, which is wrong, and the plugin sync will re-install the same version
```
Plugins from Context:  tkg1
  NAME                DESCRIPTION                           TARGET      VERSION  STATUS
  cluster             Kubernetes cluster operations         kubernetes  v0.30.1  update available
```

pkg/plugininventory/sqlite_inventory.go
```diff
@@ -221,8 +221,12 @@ func createPluginWhereClause(filter *PluginInventoryFilter) (string, error) {
...
 				whereClause = fmt.Sprintf("%s Version=RecommendedVersion AND", whereClause)
 			} else {
-				// We want a specific version of the plugin
-				whereClause = fmt.Sprintf("%s Version='%s' AND", whereClause, filter.Version)
+				// We want a specific version of the plugin or the plugin version that matches vMAJOR or vMAJOR.MINOR pattern
+				// In following condition, "Version Like '%s.%%'" condition should handle the cases where only major or major.minor version is specified
+				// e.g. If specified version is `v1` it matches with all versions like v1.MINOR.PATCH
+				//      If specified version is `v1.2` it matches will all versions like v1.2.PATCH
+				// And "Version='%s'" condition will try to match with the exact same match for the version
+				whereClause = fmt.Sprintf("%s ( Version Like '%s.%%' OR Version='%s' ) AND", whereClause, filter.Version, filter.Version)
```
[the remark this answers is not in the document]
anujc25: Thanks. I didn't know about this.

anujc25:
> * For error messages that say `unable to find plugin '%v' with version '%v'` we will now show "short" versions like `v1`.  How about changing the error message to use the word `matching to say `unable to find plugin '%v' matching version '%v'`?

Good point. Updated the error message.

> * Is it ok for the special plugins that have actual binaries with versions v0.25/v0.28/v0.29 to now install the latest patch instead of the specific binary with that version? I hope so, because the behaviour of the PR is what I'm hoping we can keep.

Yes. That was the goal with v0.25/v0.28/v0.29 versions but we did not have this feature in past and hence we had to release it separately.

> * I think we need to look at how this feature can be used for context-scoped plugins.
I created a CR for the cluster plugin with a version of v0.30.
When the plugin is not installed I see version v0.30, which is probably ok
when I do a plugin sync I then see v0.30.1 which is ok, but is shows there is an update available, which is wrong, and the plugin sync will re-install the same version

I have updated the `DiscoverServerPlugins` function a bit to handle this scenario. However, when the plugin is not installed it will also show the latest matching version (in this case v0.30.1). I feel it should be reasonable as it is more explicitly saying which version will be installed with the `tanzu plugin sync` but would be nice to get your opinion on this.

pkg/pluginmanager/manager.go
```diff
@@ -997,13 +993,13 @@ func InstallPluginsFromLocalSource(pluginName, version string, target configtype
 	}
 
 	if len(matchedPlugins) == 1 {
-		return installOrUpgradePlugin(&matchedPlugins[0], FindVersion(matchedPlugins[0].RecommendedVersion, version), installTestPlugin)
+		return installOrUpgradePlugin(&matchedPlugins[0], matchedPlugins[0].RecommendedVersion, installTestPlugin)
```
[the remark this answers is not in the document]
marckhouzam: Ok, I understand.  So moving `FindVersion()` is good, but you need to leave `InstallPluginsFromLocalSource()` like before where it calls `installOrUpgradePlugin()` passing it `version` instead of `matchedPlugins[i].RecommendedVersion`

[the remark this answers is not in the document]
anujc25: My bad. Got it. I misunderstood earlier.

pkg/pluginmanager/manager_test.go
```diff
@@ -213,6 +213,26 @@ func Test_InstallStandalonePlugin(t *testing.T) {
...
+	// Install login (standalone) plugin with just vMajor.Minor as version
+	err = InstallStandalonePlugin("login", "v0.2", configtypes.TargetUnknown)
+	assertions.Nil(err)
+	// Verify installed plugin
+	installedPlugins, err = pluginsupplier.GetInstalledPlugins()
+	assertions.Nil(err)
+	assertions.Equal(1, len(installedPlugins))
+	assertions.Equal("login", installedPlugins[0].Name)
+	assertions.Equal("v0.2.0", installedPlugins[0].Version)
+
```
marckhouzam, line 235: Would it be possible to add a login plugin of version `v0.20.0` in the DB (in `manager_helper_test.go`) to make sure that installing `v0.2` will not match the newer `v0.20.0`

marckhouzam (approved):
One small suggestion to improve the tests.
LGTM

Nice job!

pkg/pluginmanager/manager_test.go
anujc25: Yup. Meant to do that earlier as well. Just added some tests. Thanks

marckhouzam (approved):
Nice that you also added a pre-release version in the tests👍

LGTM

marckhouzam:
@anujc25 I think this feature is impactful enough that it should be mentioned in the documentation.  I'm thinking of at least two areas:
1. using plugin groups with "shortened" plugin versions
2. context-scope plugins with "shortened" plugin versions

What do you think?

anujc25:
Added some basic documentation. I will be adding more as part of follow-up PR.

PR #496, opened 2023-09-25
Add e2e tests for a DB that has an extra column

marckhouzam:

This PR adds E2E tests to verify that the latest code for the CLI can handle a plugin inventory DB with an extra column (both for the plugin table and for the plugin group table).

The approach taken is simple:
1. have a test central repo that has an extra column in `localhost:9876/tanzu-cli/plugins/extra:small`
2. add a make target `e2e-extra-column-tests` which sets the test repo to the above image with the extra column and then call the standard `make e2e-plugin-lifecycle-tests`

The biggest part of this PR is updating the code that generates the test central repos.  To be able to run the same `make e2e-plugin-lifecycle-tests` on the `extra:small` image, that image must contain the same set of plugins and groups as the `central:small` image.  To make this future-proof, the PR generates both images in a loop using the same code with small tweaks.

Fixes # N/A

I made sure both E2E sets of tests passed:
```
# run the new tests
make e2e-extra-column-tests 
# run the old tests
make e2e-plugin-lifecycle-tests
```

I also hacked the CLI to not properly handle an extra command (by using `SELECT * FROM ...` and ran `make e2e-plugin-lifecycle-tests` to see the tests properly fail.

And then there is the CI run.

### Release note

```release-note

```


anujc25 (approved):
LGTM. Thanks

PR #504, opened 2023-09-28
Show log message on plugin installation based on various plugin installation states

anujc25:

* Updated the PR to show proper log messages based on different plugin installation states

Here is how the log messages are shown:

**Scenario 1: If the plugin binary is NOT available in the cache OR the plugin is NOT installed (as per catalog cache)**
    * `Installing plugin 'pluginName:version'`

**Scenario 2: If the plugin binary is available in the cache AND the plugin is NOT installed (as per catalog cache)**
    * `Installing plugin 'pluginName:version' (from cache)`

**Scenario 3: If the plugin binary is available in the cache AND the plugin is installed (as per catalog cache)**
    * `Plugin 'pluginName:version' is already installed. Reinitializing...`

Fixes #

```
$ tz plugin list
Standalone Plugins

## Install `package` plugin for the first time
$ tz plugin install package
[i] Installing plugin 'package:v0.31.0' with target 'kubernetes'
[ok] successfully installed 'package' plugin

## Reinstall `package` plugin when it is already installed
$ tz plugin install package
[i] Plugin 'package:v0.31.0' with target 'kubernetes' is already installed. Reinitializing...
[ok] successfully installed 'package' plugin


## Delete the package plugin
$ tz plugin delete package
Deleting Plugin 'package' for target 'kubernetes'. Are you sure? [y/N]: y
[ok] successfully deleted plugin 'package'

## Reinstall the `package` plugin after deletion (should fetch the plugin from cache)
$ tz plugin install package
[i] Installing plugin 'package:v0.31.0' with target 'kubernetes' (from cache)
[ok] successfully installed 'package' plugin

## Remove the plugin binary directly from file system without updating catalog cache
$ rm /Users/anujc/Library/ApplicationSupport/tanzu-cli/package/v0.31.0_7439a6a7dcb8d584b13c7780abb5b181c234d0d6d642e516e8e402c4ac416052_kubernetes

## Reinstall the `package` plugin after deleting binary
$ tz plugin install package
[i] Installing plugin 'package:v0.31.0' with target 'kubernetes'
[ok] successfully installed 'package' plugin
```

For context-scoped plugins the behavior remain as it is.
```
## Activate a context (after cleaning all installed plugins)
$ tz context use kind
[i] Installing plugin 'kubernetes-release:v0.31.0' with target 'kubernetes'

$ tz plugin sync
[ok] Done

## Delete the kubernetes-release plugin
$ tz plugin delete kubernetes-release
Deleting Plugin 'kubernetes-release' for target 'kubernetes'. Are you sure? [y/N]: y
[ok] successfully deleted plugin 'kubernetes-release'

## Run sync after plugin deletion should install plugins from cache
$ tz plugin sync
[i] Installing plugin 'kubernetes-release:v0.31.0' with target 'kubernetes' (from cache)
[i] Successfully installed all required plugins
[ok] Done
```

### Release note

```release-note
Show log message on plugin installation based on various plugin installation states
```

* Thanks for the feedback on the draft PR. I have updated the PR to show logs based on different plugin installation condition based on the feedback from @vuil and @marckhouzam

### DIFF
--- a/pkg/pluginmanager/manager.go
+++ b/pkg/pluginmanager/manager.go
@@ -665,14 +665,14 @@ func InstallPluginsFromGroup(pluginName, groupIDAndVersion string, options ...Pl
 func logPluginInstallationMessage(p *discovery.Discovered, version string, isPluginInCache, isPluginAlreadyInstalled bool) {
 	withTarget := ""
 	if p.Target != configtypes.TargetUnknown {
-		withTarget = fmt.Sprintf("with target '%v'", p.Target)
+		withTarget = fmt.Sprintf("with target '%v' ", p.Target)
 	}
 
 	if isPluginInCache {
 		if !isPluginAlreadyInstalled {
-			log.Infof("Installing plugin '%v:%v' %v (from cache)", p.Name, version, withTarget)
+			log.Infof("Installing plugin '%v:%v' %v(from cache)", p.Name, version, withTarget)
 		} else {
-			log.Infof("Plugin '%v:%v' %v is already installed. Reinitializing...", p.Name, version, withTarget)
+			log.Infof("Plugin '%v:%v' %vis already installed. Reinitializing...", p.Name, version, withTarget)
 		}
 	} else {
 		log.Infof("Installing plugin '%v:%v' %v", p.Name, version, withTarget)
--- a/pkg/pluginmanager/manager.go
+++ b/pkg/pluginmanager/manager.go
@@ -662,25 +662,44 @@ func InstallPluginsFromGroup(pluginName, groupIDAndVersion string, options ...Pl
 	return groupIDAndVersion, nil
 }
 
+func logPluginInstallationMessage(p *discovery.Discovered, version string, isPluginInCache, isPluginAlreadyInstalled bool) {
+	withTarget := ""
+	if p.Target != configtypes.TargetUnknown {
+		withTarget = fmt.Sprintf("with target '%v'", p.Target)
+	}
+
+	if isPluginInCache {
+		if !isPluginAlreadyInstalled {
+			log.Infof("Installing plugin '%v:%v' %v (from cache)", p.Name, version, withTarget)
+		} else {
+			log.Infof("Plugin '%v:%v' %v is already installed. Reinitializing...", p.Name, version, withTarget)
+		}
+	} else {
+		log.Infof("Installing plugin '%v:%v' %v", p.Name, version, withTarget)
+	}
+}
+
 func installOrUpgradePlugin(p *discovery.Discovered, version string, installTestPlugin bool) error {
 	// If the version requested was the RecommendedVersion, we should set it explicitly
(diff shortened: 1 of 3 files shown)


marckhouzam:
> Currently the plugin installation is determined by catalog cache so if the plugin binary gets deleted underneath and when a user tries to reinstall the same version of the plugin which is present in the catalog cache, CLI might skip the installation because of this check.

Good thinking.
1. One can argue that if the cache has been modified but not the catalog, we are not dealing with a supported case; a user would need to first delete the plugin to force a re-install.
2. Or, if we want the CLI to deal with it, we can use the `getPluginFromCache()` function to check if the binary of plugin that is installed is present.  This becomes a bit more complicated as we now have to check both ARM64 then AMD64 on Mac if we want to use this function.

BUT, the next point will affect the above.

> Currently we already have a check to use the plugin binary cache if we find it in the system instead of downloading it from the internet, considering this optimization, do we really need to skip the plugin installation?

I agree.  But I do like that we clarify to the user that the plugin is already installed.

Maybe that is what you have in mind: we print that the plugin is already installed, but we still go ahead with the installation.  It becomes clearer for the user, and we are protected against a missing binary and such.

If we do this, how about having a printout saying:
`[i] Plugin 'isolated-cluster:v0.31.0' with target 'global' is already installed. Verifying the installation...`

So we would have:
```
$ tz plugin install isolated-cluster -v v0.31.0 -t global
[i] Plugin 'isolated-cluster:v0.31.0' with target 'global' is already installed. Verifying the installation...
[i] Installing plugin 'isolated-cluster:v0.31.0' with target 'global'
[i] Plugin binary for 'isolated-cluster:v0.31.0' found in cache
[ok] successfully installed 'isolated-cluster' plugin
```

marckhouzam:
When we find that the plugin is already installed, we could also disable the digest verification, which would make the re-install almost instantaneous.  We can use `TEST_TANZU_CLI_USE_DB_CACHE_ONLY=1` for that, although I think we need a more elegant way to force the cache in such cases.

anujc25:
> When we find that the plugin is already installed, we could also disable the digest verification, which would make the re-install almost instantaneous. We can use `TEST_TANZU_CLI_USE_DB_CACHE_ONLY=1` for that, although I think we need a more elegant way to force the cache in such cases.

I also feel that we should do it more generically and I would like to keep the existing behavior in place but show correct log message to the user based on the current situation. I have updated the PR based on the feedback and @vuil's (online) suggestions.

anujc25:
Updated the PR and PR description based on the discussion.

marckhouzam (approved):
I like this a lot.
One nit but otherwise LGTM

marckhouzam:
Nit: for scenario 1 in the PR description, it should be 
`Scenario 1: If the plugin binary is NOT available in the cache OR the plugin is NOT installed (as per catalog cache)`

`OR` instead of `AND`

PR #510, opened 2023-09-29
Update tanzu context list to include UCP data

vuil:
### What this PR does / why we need it

Shows additional information relevant to the UCP context variant.
Not that the tabular output conditionally hides the UCP-specific columns if no such contexts are available.
An upcoming rename exercise will fix up the terminology to use TAE, till then only `context list --target ucp`
works as a means to filter TAE contexts.

### Which issue(s) this PR fixes

Fixes #N/A

### Describe testing done for PR

```
tanzu-cli (context-list)> tz context list
Target:  kubernetes
  NAME    ISACTIVE  TYPE     ENDPOINT                                                                                           KUBECONFIGPATH                KUBECONTEXT       PROJECT  SPACE
  kucp    true      cluster                                                                                                     /Users/vuichiap/.kube/config  kind-ucp
  foo     false     cluster                                                                                                     /Users/vuichiap/.kube/config  kind-foo
  stucp   false     TAE      https://api-dev.tanzu.cloud.vmware.com/org/7753c388-914c-4085-1111-111111111111                    /Users/vuichiap/.kube/config  tanzu-cli-stucp   projA
  sprucp  true      TAE      https://ucp-aria.stacks.bluesky.tmc-dev.cloud.vmware.com/org/ee04bfae-a665-4f20-2222-222222222222  /Users/vuichiap/.kube/config  tanzu-cli-sprucp
Target:  mission-control
  NAME       ISACTIVE  ENDPOINT
  vui-tmc-2  true      unable.tmc-dev.cloud.vmware.com:443

tanzu-cli (context-list)> tz context list --target ucp
Target:  kubernetes
  NAME    ISACTIVE  TYPE  ENDPOINT                                                                                           KUBECONFIGPATH                KUBECONTEXT       PROJECT  SPACE
  stucp   false     TAE   https://api-dev.tanzu.cloud.vmware.com/org/7753c388-914c-4085-1111-111111111111                    /Users/vuichiap/.kube/config  tanzu-cli-stucp   projA
  sprucp  true      TAE   https://ucp-aria.stacks.bluesky.tmc-dev.cloud.vmware.com/org/ee04bfae-a665-4f20-2222-222222222222  /Users/vuichiap/.kube/config  tanzu-cli-sprucp

tanzu-cli (context-list)> tz context list --target k8s
Target:  kubernetes
  NAME  ISACTIVE  TYPE     ENDPOINT  KUBECONFIGPATH                KUBECONTEXT
[...]

PR #512, opened 2023-10-01
Fix compilation error for unit tests

marckhouzam:

This PR updates some IDs in `context_test.go` which didn't compile.

Note that the unit test still fail, but it is because of a change in https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/103.
At least the tests compile, which also fixes the linter.

Fixes # N/A

`make lint`

Ran the tests in `context_test.go`: they now fail but they compile; they used to not compile.

### Release note

```release-note

```


vuil:
@marckhouzam thanks, I noticed as well and I am taking care of it in my change

marckhouzam:
Resolved by #510

PR #523, opened 2023-10-05
Ensure the k8s and application-engine current contexts are mutual exclusive

prkalle:
This PR ensures the k8s and application-engine current contexts are mutual exclusive at the end of end cli command/plugin invocation and also specifically for `SyncPlugin()` command as this command can be invoked by plugin through plugin-runtime API before finishing plugin command execution(eg: management-cluster create command creates management cluster and would invoke SyncPlugins within the command)

**TODO:**
- This PR has dependency on plugin-runtime PR #https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/113 and this PR needs to be updated once the other PR is merged.

Fixes #

Unit test ran locally passed 

### Release note

```release-note
Ensure the kubernetes and application-engine current contexts are mutual exclusive
```

### DIFF
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -22,6 +22,7 @@ import (
 	"github.com/vmware-tanzu/tanzu-cli/pkg/pluginmanager"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/pluginsupplier"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/telemetry"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/utils"
 
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
 	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
@@ -134,6 +135,12 @@ func newRootCmd() *cobra.Command {
 		// Flag parsing must be deactivated because the root plugin won't know about all flags.
 		DisableFlagParsing: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			// Ensure mutual exclusion in current contexts just in case if any plugins with old
+			// plugin-runtime sets k8s context as current when tae context is already set as current
+			if err := utils.EnsureMutualExclusiveCurrentContexts(); err != nil {
+				return err
+			}
+
 			if !shouldSkipTelemetryCollection(cmd) {
 				if err := telemetry.Client().UpdateCmdPreRunMetrics(cmd, args); err != nil {
 					telemetry.LogError(err, "")
@@ -160,6 +167,14 @@ func newRootCmd() *cobra.Command {
 				if err := cliconfig.ConfigureCEIPOptIn(); err != nil {
 					return err
 				}
+			}
+			return nil
+		},
+		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
+			// Ensure mutual exclusion in current contexts just in case if any plugins with old
+			// plugin-runtime sets k8s context as current when tae context is already set as current
+			if err := utils.EnsureMutualExclusiveCurrentContexts(); err != nil {
+				return err
 			}
 			return nil
 		},
--- a/pkg/utils/common_test.go
+++ b/pkg/utils/common_test.go
@@ -4,7 +4,15 @@
 package utils
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
+
+	"github.com/otiai10/copy"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 )
 
 // TestContainsString tests the ContainsString function.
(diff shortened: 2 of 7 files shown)


pkg/pluginmanager/manager.go
```diff
@@ -975,6 +975,13 @@ func doDeletePluginFromCatalog(pluginName string, target configtypes.Target, cat
 // If the central-repo is disabled, all discovered plugins will be installed.
 func SyncPlugins() error {
 	log.Info("Checking for required plugins...")
+	// ensure the k8s and tae current contexts are mutually exclusive.
```
prkalle, line 978: May be we can avoid this check in `SyncPlugins()` method by adding a check for ensuring mutual exclusion in root command's `PersistentPreRunE` hook.

prkalle: After discussion, removed this check and added the check in `PersistentPreRunE` hook.

pkg/utils/common_test.go
```diff
@@ -87,3 +95,85 @@ func TestGenerateKey(t *testing.T) {
 		})
 	}
 }
+
+func setupMultiConcurrentContexts(t *testing.T) func() {
+	tkgConfigFile, err := os.CreateTemp("", "config")
```
marckhouzam, line 100: Nit: let's stop calling these files with a `tkg` prefix.  We can just use `configFile` and `configFileNG`

pkg/utils/common_test.go
```diff
@@ -87,3 +95,85 @@ func TestGenerateKey(t *testing.T) {
...
+	return cleanup
+}
+
+// TestGenerateKey tests the GenerateKey function.
+func TestEnsureMutualExclusiveCurrentContexts(t *testing.T) {
+	cleanup := setupMultiConcurrentContexts(t)
+
+	defer func() {
+		cleanup()
+	}()
```
marckhouzam, line 132: No need to wrap `cleanup()` in another `func`.  You can just do `defer cleanup()`

marckhouzam (commented):
Elegant.  Thanks @prkalle 
One important thing to fix (which explains CI failing) and two nits.

pkg/utils/common_test.go
prkalle: Agreed. Updated.

pkg/utils/common_test.go
prkalle: Make sense as there is only one method for defer. Fixed.

pkg/command/root.go
```diff
@@ -134,6 +135,13 @@ func newRootCmd() *cobra.Command {
 		// Flag parsing must be deactivated because the root plugin won't know about all flags.
 		DisableFlagParsing: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			// Ensure mutual exclusion in current contexts just in case if any plugins with old
+			// plugin-runtime sets k8s context as current when tae context is already set as current
+			if err := utils.EnsureMutualExclusiveCurrentContexts(); err != nil {
+				return err
+			}
+			return nil
```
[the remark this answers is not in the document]
prkalle: Thanks for catching this @marckhouzam . Fixed.

vuil (approved):
lgtm. thanks for the changes and fixes!

marckhouzam (approved):
Once the go.mod/go.sum file are updated, this LGTM.
But technically, I don't even thing the go.mod file needs to point to a new plugin-runtime, right?  This change actually does the synchronization on its own.

prkalle:
> Once the go.mod/go.sum file are updated, this LGTM. But technically, I don't even thing the go.mod file needs to point to a new plugin-runtime, right? This change actually does the synchronization on its own.

We won't need for this change (where we ensure the mutual exclusion is maintained for plugins using older plugin runtime) but the CLI should be updated to use latest plugin-runtime to ensure mutual exclusion among the current contexts(except TMC) if user uses the command `tanzu context use <context-name>`

marckhouzam:
> We won't need for this change (where we ensure the mutual exclusion is maintained for plugins using older plugin runtime), but the CLI should be updated to use latest plugin-runtime to ensure mutual exclusion among the current contexts(except TMC) if user uses the command `tanzu context use <context-name>`

I agree with the update to the new plugin-runtime.
But to make sure we are on the same page: with the `PersistentPreRun/PostRun` of this PR, won't we have mutual exclusion all the time, even without the new plugin runtime?

marckhouzam (approved):
LGTM @prkalle 
Thanks

prkalle:
> > We won't need for this change (where we ensure the mutual exclusion is maintained for plugins using older plugin runtime), but the CLI should be updated to use latest plugin-runtime to ensure mutual exclusion among the current contexts(except TMC) if user uses the command `tanzu context use <context-name>`
> 
> I agree with the update to the new plugin-runtime. But to make sure we are on the same page: with the `PersistentPreRun/PostRun` of this PR, won't we have mutual exclusion all the time, even without the new plugin runtime?

Yes, that's correct. (Though `PersistentPreRun/PostRun` would ensure the mutual exclusion, it should be just a safe net and is intended to handle the case where old plugins using old plugin-runtime and plugin-runtime API should actually handle the mutual exclusion while setting the current context)

PR #333, opened 2023-06-01
UX: Plugin commands help message updated

chandrareddyp:
### What this PR does / why we need it

This PR fixes the CLI UX help message for the plugin sub-commands. Before this fix, the mandatory plugin name was referred to as [name], which typically indicates an option input parameter. As part of this fix, the help message for the mandatory plugin name has been updated to PLUGIN_NAME (a qualified string in capital letters, without square brackets). 

This fix applies to the following commands:
`tanzu plugin describe`
`tanzu plugin install`
`tanzu plugin upgrade`
`tanzu plugin delete`

### Which issue(s) this PR fixes
1. E2E test cases have been added as part of this PR to validate the plugin commands' help message.
2. Manual testing has been conducted, and the following is the output:

**Before the fix, help message for the plugin install/describe/upgrade/delete commands:**
```
❯ to plugin describe -h
Displays detailed information for a plugin

Usage:
tanzu plugin describe [name] [flags]

Flags:
  -h, --help            help for describe
  -o, --output string   Output format (yaml|json|table)
  -t, --target string   target of the plugin (kubernetes[k8s]/mission-control[tmc]/global)
❯ to plugin install -h
Install a specific plugin by name or specify all to install all plugins of a group

Usage:
tanzu plugin install [name] [flags]

Examples:

    # Install all plugins of the vmware-tkg/default plugin group version v2.1.0
    tanzu plugin install --group vmware-tkg/default:v2.1.0

    # Install all plugins of the latest version of the vmware-tkg/default plugin group
    tanzu plugin install --group vmware-tkg/default

    # Install the latest version of plugin "myPlugin"
	# If the plugin exists for more than one target, an error will be thrown
    tanzu plugin install myPlugin

    # Install the latest version of plugin "myPlugin" for target kubernetes
    tanzu plugin install myPlugin --target k8s

    # Install version v1.0.0 of plugin "myPlugin"
    tanzu plugin install myPlugin --version v1.0.0

Flags:
      --group string     install the plugins specified by a plugin-group version
  -h, --help             help for install
  -t, --target string    target of the plugin (kubernetes[k8s]/mission-control[tmc]/global)
  -v, --version string   version of the plugin (default "latest")
❯ to plugin upgrade -h
Installs the latest version available for the specified plugin

Usage:
tanzu plugin upgrade [name] [flags]

Flags:
  -h, --help            help for upgrade
  -t, --target string   target of the plugin (kubernetes[k8s]/mission-control[tmc]/global)
[...]

PR #544, opened 2023-10-14
Update `tanzu config set/unset` help message

chandrareddyp:
As part of this pull request, the tanzu config set and tanzu config unset commands' --help messages have been updated to be consistent with the help messages of all other Tanzu Core commands.

Fixes #

- E2E tests are added
Before Fix
```
❯ tanzu config set --help
Set config values at the given path. path values: [unstable-versions, cli.edition, features.global.<feature>, features.<plugin>.<feature>, env.<variable>]

Usage:
tanzu config set <path> <value> [flags]

Flags:
  -h, --help   help for set
❯ tanzu config unset --help
Unset config values at the given path. path values: [features.global.<feature>, features.<plugin>.<feature>, env.global.<variable>, env.<plugin>.<variable>]

Usage:
tanzu config unset <path> [flags]

Flags:
  -h, --help   help for unset
❯ 
```
After the fix
```
❯ t config set --help  
Set config values at the given PATH. Supported PATH values: [features.global.<feature>, features.<plugin>.<feature>, env.<variable>]

Usage:
tanzu config set PATH <value> [flags]

Flags:
  -h, --help   help for set
❯ t config unset --help
Unset config values at the given PATH. Supported PATH values: [features.global.<feature>, features.<plugin>.<feature>, env.<variable>]

Usage:
tanzu config unset PATH [flags]

Flags:
  -h, --help   help for unset
❯ 
```

### Release note

```release-note
`tanzu config set` and `tanzu config unset` commands `--help` messages are updated.
```

### DIFF
--- a/pkg/command/config.go
+++ b/pkg/command/config.go
@@ -83,16 +83,16 @@ var getConfigCmd = &cobra.Command{
 }
 
 var setConfigCmd = &cobra.Command{
-	Use:               "set <path> <value>",
-	Short:             "Set config values at the given path",
-	Long:              "Set config values at the given path. path values: [unstable-versions, cli.edition, features.global.<feature>, features.<plugin>.<feature>, env.<variable>]",
+	Use:               "set PATH <value>",
+	Short:             "Set config values at the given PATH",
+	Long:              "Set config values at the given PATH. Supported PATH values: [features.global.<feature>, features.<plugin>.<feature>, env.<variable>]",
 	ValidArgsFunction: completeSetConfig,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if len(args) < 2 {
-			return errors.Errorf("both path and value are required")
+			return errors.Errorf("both PATH and <value> are required")
 		}
 		if len(args) > 2 {
-			return errors.Errorf("only path and value are allowed")
+			return errors.Errorf("only PATH and <value> are allowed")
 		}
 
 		// Acquire tanzu config lock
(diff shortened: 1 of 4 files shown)


pkg/command/plugin.go
```diff
@@ -202,7 +202,7 @@ func newDescribePluginCmd() *cobra.Command {
 
 func newInstallPluginCmd() *cobra.Command { //nolint:funlen
 	var installCmd = &cobra.Command{
-		Use:   "install [" + pluginNameCaps + "]",
+		Use:   "install " + pluginNameCaps,
```
marckhouzam, line 205: This is a bit tricky because the `PLUGIN_NAME` is optional when we use `--group`.
This was mentioned in a previous PR: https://github.com/vmware-tanzu/tanzu-cli/pull/333#discussion_r1213685505

Note that `plugin install all` is no longer supported.
I would leave this line as is.

marckhouzam:
You may want to consider changing the usage string for `config set` and `config unset` to use the uppercase format

pkg/command/plugin.go
marckhouzam: Maybe we should put a comment to say that the brackets are on purpose just so we don’t try to change it a third time

chandrareddyp: > This is a bit tricky because the PLUGIN_NAME is optional when we use --group.
The current help message is below:
---------------------------------------------------- 
❯ tanzu plugin install -h
Install a specific plugin by name or specify all to install all plugins of a group
Usage:
tanzu plugin install [PLUGIN_NAME] [flags]

---------------------------------------------------- 
The PLUGIN_NAME is not optional for the `tanzu plugin install` command. Therefore, we should maintain consistency in the help message by keeping PLUGIN_NAME without square brackets. We should update the help message to include the '--group' flag.
eg:
---------------------------------------------------- 
❯ tanzu plugin install -h
Install a specific plugin by name or specify all to install all plugins of a group
Usage:
tanzu plugin install --group PLUGIN_GROUP_NAME
---------------------------------------------------- 
We can discuss with the team, and take decision.

marckhouzam: Actually the plugin name is optional of you use `--group` although you’re right, the help message doesn’t say that.

pkg/command/config.go
```diff
@@ -83,9 +83,9 @@ var getConfigCmd = &cobra.Command{
 }
 
 var setConfigCmd = &cobra.Command{
-	Use:               "set <path> <value>",
+	Use:               "set PATH <value>",
 	Short:             "Set config values at the given path",
-	Long:              "Set config values at the given path. path values: [unstable-versions, cli.edition, features.global.<feature>, features.<plugin>.<feature>, env.<variable>]",
+	Long:              "Set config values at the given path. Supported PATH values: [features.global.<feature>, features.<plugin>.<feature>, env.<variable>]",
```
marckhouzam, line 88: Can you do the same two changes for the `unsetConfigCmd` please

marckhouzam (approved):
LGTM
Thanks

PR #526, opened 2023-10-08
Clarify the output of "plugin group get --all" and add unit tests

marckhouzam:
### What this PR does / why we need it

This PR enhances the output of `tanzu plugin group get --all` to make it clear which plugins are standalone and which are context-scoped.

Please refer to the section "testing done" for the new output.

The top-most commit of this PR provides general unit tests for the `tanzu plugin group` sub-commands.
The testing coverage for `plugin_group.go` has gone up from 0% to 93%. 

### Which issue(s) this PR fixes

Fixes #525

### Describe testing done for PR

```
# Same output as before without the --all flag
$ tz plugin group get vmware-tkg/default
Plugins in Group:  vmware-tkg/default:v2.4.0
  NAME                TARGET      VERSION
  isolated-cluster    global      v0.31.0
  management-cluster  kubernetes  v0.31.0
  package             kubernetes  v0.31.0
  pinniped-auth       global      v0.31.0
  secret              kubernetes  v0.31.0
  telemetry           kubernetes  v0.31.0

# Enhanced output with the --all flag
# Notice the information line that mentions the `tanzu plugin install -group`
# command mentions the plugin group that the user used
$ tz plugin group get vmware-tkg/default --all
Plugins in Group:  vmware-tkg/default:v2.4.0

Standalone Plugins
  NAME                TARGET      VERSION
  isolated-cluster    global      v0.31.0
  management-cluster  kubernetes  v0.31.0
  package             kubernetes  v0.31.0
  pinniped-auth       global      v0.31.0
  secret              kubernetes  v0.31.0
  telemetry           kubernetes  v0.31.0

[i] The standalone plugins in this plugin group are installed when the 'tanzu plugin install --group vmware-tkg/default' command is invoked.

Contextual Plugins
  NAME                TARGET      VERSION
  cluster             kubernetes  v0.31.0
  feature             kubernetes  v0.31.0
  kubernetes-release  kubernetes  v0.31.0

[i] The contextual plugins in this plugin group are automatically installed, and only available for use, when a Tanzu context which supports them is created or activated/used.

# Notice the information line that mentions the `tanzu plugin install -group`
# command mentions the plugin group that the user used
$ tz plugin group get vmware-tkg/default:v2.4.0 --all
Plugins in Group:  vmware-tkg/default:v2.4.0

Standalone Plugins
  NAME                TARGET      VERSION
  isolated-cluster    global      v0.31.0
  management-cluster  kubernetes  v0.31.0
  package             kubernetes  v0.31.0
[...]

PR #559, opened 2023-10-20
Unhide --all flag for plugin group get

marckhouzam:

Following the improvements to the output of `tanzu plugin group get --all` from #526 , this PR makes the `--all` flag un-hidden.

Fixes # N/A

```
# Notice the --all flag in the help
$ tz plugin group get  -h
Get the content of the specified plugin-group.  A plugin-group provides a list of plugin name/version combinations which can be installed in one step.  This command allows to see the list of plugins included in the specified group.

Usage:
tanzu plugin group get GROUP_NAME [flags]

Flags:
      --all             include the contextual plugins
  -h, --help            help for get
  -o, --output string   output format (yaml|json|table)

# Notice that shell completion now suggest the --all flag since it no longer hidden 
$ tz plugin group get vmware-tkg/default -<TAB>
--all         -- include the contextual plugins
--help    -h  -- help for get
--output  -o  -- output format (yaml|json|table)
```
### Release note

```release-note
Make the `--all` flag of `tanzu plugin group get` not hidden.
```

### DIFF
--- a/pkg/command/plugin_group.go
+++ b/pkg/command/plugin_group.go
@@ -145,8 +145,7 @@ func newGetCmd() *cobra.Command {
 	f.StringVarP(&outputFormat, "output", "o", "", "output format (yaml|json|table)")
 	utils.PanicOnErr(getCmd.RegisterFlagCompletionFunc("output", completionGetOutputFormats))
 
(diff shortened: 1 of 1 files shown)


vuil (approved):
lgtm, thanks

PR #561, opened 2023-10-20
Show a core command's short text in ActiveHelp when completing arguments

marckhouzam:

Setup a generic ValidArgsFunction to add a command's 'short' string as activeHelp when doing arg completion for any command.

Doing something like "tanzu context <TAB>" will show normal shell completions but will now also print the short help for "tanzu context".

Note that activeHelp only gets shown for `bash` and `zsh`.

```
$ tanzu context <TAB>
Command help: Configure and manage contexts for the Tanzu CLI
--
create  -- Create a Tanzu CLI context
delete  -- Delete a context from the config
get     -- Display a context from the config
list    -- List contexts
unset   -- Unset the active context so that it is not used by default.
use     -- Set the context to be used by default
```

The prefix added before the help text is to deal with situation like the following:

```
$ tz context create tmc --endpoint unstable.tmc-dev.cloud.vmware.com <TAB>
Command help: Create a Tanzu CLI context
```

Without the prefix the text "Create a Tanzu CLI context" could be interpreted as an action the user must do, instead of the description of the command.

Because this change would affect most of the shell completion unit tests, each batch of such test was updated to disable this feature using TANZU_ACTIVE_HELP=no_short_help.

A unit test specific to this feature was added in root_test.go.

Fixes # N/A

Extensive unit test for shell completion show that nothing was broken.

New feature:
```
$ tz plugin <TAB>
Command help: Manage CLI plugins
--
clean            -- Clean the plugins
describe         -- Describe a plugin
download-bundle  -- Download plugin bundle to the local system
group            -- Manage plugin-groups
install          -- Install a plugin
list             -- List installed plugins
search           -- Search for available plugins
source           -- Manage plugin discovery sources
sync             -- Installs all plugins recommended by the active contexts
uninstall        -- Uninstall a plugin
upgrade          -- Upgrade a plugin
upload-bundle    -- Upload plugin bundle to a repository

$ tz plugin group <TAB>
Command help: Manage plugin-groups
--
get     -- Get the content of the specified plugin-group
search  -- Search for available plugin-groups

$ tz version
Command help: Version information
This command does not take any more arguments (but may accept flags).
```

### Release note

```release-note
Show a core command's short text in ActiveHelp when completing arguments
```

Although this feature was added as part of the PersistentPreRun so that it can affect all commands, it only applies to _core_ commands.  To apply to plugins, we need to make a similar change in tanzu-plugin-runtime.  And the change would only affect plugins that have been re-compiled with this new tanzu-plugin-runtime.

Although I have actually coded the required change, I feel the change is too risky so late in the cycle.  This is because it mucks around replacing `PersistentPreRuns` and such.

If it is felt that the change for plugins must be available before merging this PR than we should delay this PR.

### DIFF
--- a/pkg/command/ceip_participation_test.go
+++ b/pkg/command/ceip_participation_test.go
@@ -118,6 +118,10 @@ var _ = Describe("ceip-participation command tests", func() {
 })
 
 func TestCompletionCeip(t *testing.T) {
+	// This is global logic and needs not be tested for each
+	// command.  Let's deactivate it.
+	os.Setenv("TANZU_ACTIVE_HELP", "no_short_help")
+
 	tests := []struct {
 		test     string
 		args     []string
@@ -167,4 +171,6 @@ func TestCompletionCeip(t *testing.T) {
 			assert.Equal(spec.expected, out.String())
 		})
 	}
+
+	os.Unsetenv("TANZU_ACTIVE_HELP")
 }
--- a/pkg/command/cert_test.go
+++ b/pkg/command/cert_test.go
@@ -252,6 +252,10 @@ func TestCompletionCert(t *testing.T) {
 	assert.Nil(t, err)
 	os.Setenv("TANZU_CONFIG_NEXT_GEN", configFileNG.Name())
 
+	// This is global logic and needs not be tested for each
+	// command.  Let's deactivate it.
+	os.Setenv("TANZU_ACTIVE_HELP", "no_short_help")
+
 	// Create some test certs
 	cert1 := configtypes.Cert{
 		Host:           "example.com",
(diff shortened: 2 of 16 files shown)


pkg/command/ceip_participation_test.go
```diff
@@ -118,6 +118,10 @@ var _ = Describe("ceip-participation command tests", func() {
 })
 
 func TestCompletionCeip(t *testing.T) {
+	// This is global logic and needs not be tested for eac
```
anujc25, line 121: nit: Another places as well.
```suggestion
	// This is global logic and needs not be tested for each
```

anujc25 (approved):
LGTM. Adding a comment on why this environment variable `TANZU_ACTIVE_HELP` was used in the Tests to reactive active help would be nice.

marckhouzam:
> LGTM. Adding a comment on why this environment variable `TANZU_ACTIVE_HELP` was used in the Tests to reactive active help would be nice.

In the end you made me realize that it is very obscure that the variable `TANZU_ACTIVE_HELP` comes from Cobra and that its value is returned by `cobra.GetActiveHelpConfig(cmd)`. 

So I replaced the call do `cobra.GetActiveHelpConfig(cmd)` with the more explicit `os.Getenv("TANZU_ACTIVE_HELP")`

PR #579, opened 2023-10-27
Rename --ca-certificate option to --ca-cert in "tanzu config cert add" and "tanzu config cert update" commands

prkalle:
This PR renames the command option `--ca-certificate ` to `ca-cert` in "tanzu config cert add" and "tanzu config cert update" commands

Fixes #

Validated the help and also the functionality for `tanzu config cert add` command with renamed `--ca-cert` options
```
❯ ./bin/tanzu config cert add -h
Add a certificate configuration for a host

Usage:
tanzu config cert add [flags]

Examples:

    # Add CA certificate for a host
    tanzu config cert add --host test.vmware.com --ca-cert path/to/ca/ert

    # Add CA certificate for a host:port
    tanzu config cert add --host test.vmware.com:8443 --ca-cert path/to/ca/ert

    # Set to skip verifying the certificate while interacting with host
    tanzu config cert add --host test.vmware.com  --skip-cert-verify true

    # Set to allow insecure (http) connection while interacting with host
    tanzu config cert add --host test.vmware.com  --insecure true

Flags:
      --ca-cert string            path to the public certificate
  -h, --help                      help for add
      --host string               host or host:port
      --insecure string           allow the use of http when interacting with the host (default "false")
      --skip-cert-verify string   skip server's TLS certificate verification (default "false”)



#### tried shell completion and it shows --ca-cert option instead of --ca-certificate
./bin/tanzu config cert add --host test2.host.com --ca-cert                                                         
--ca-cert           -- path to the public certificate
--help              -- help for add
--insecure          -- allow the use of http when interacting with the host
--skip-cert-verify  -- skip server's TLS certificate verification

#### successful with --ca-cert option
❯ ./bin/tanzu config cert add --host test2.host.com --ca-cert ~/openssl/server.crt
[ok] successfully added certificate data for host test2.host.com
❯ ./bin/tanzu config cert list
  HOST            CA-CERTIFICATE  SKIP-CERT-VERIFICATION  INSECURE
  test1.host.com  Not configured  false                   true
  localhost:9876  <REDACTED>      false                   false
  test2.host.com  <REDACTED>      false                   false

## Also validated the deprecration --ca-certificate option still works with deprecation message
❯  ./bin/tanzu config cert delete test2.host.com 
[ok] deleted certificate data for host test2.host.com
❯ ./bin/tanzu config cert add --host test2.host.com --ca-certificate ~/openssl/server.crt
Flag --ca-certificate has been deprecated, this was done in the v1.1.0 release, it will be removed following the deprecation policy (6 months). Use the --ca-cert flag instead.

[ok] successfully added certificate data for host test2.host.com
```

Validated the help and also the functionality for `tanzu config cert update` command with renamed `--ca-cert` options

```
❯ ./bin/tanzu config cert update -h
Update certificate configuration for a host

Usage:
tanzu config cert update HOST [flags]

Examples:

    # Update CA certificate for a host,
    tanzu config cert update test.vmware.com --ca-cert path/to/ca/ert

    # Update CA certificate for a host:port,
    tanzu config cert update test.vmware.com:5443 --ca-cert path/to/ca/ert

    # Update whether to skip verifying the certificate while interacting with host
    tanzu config cert update test.vmware.com  --skip-cert-verify true

    # Update whether to allow insecure (http) connection while interacting with host
    tanzu config cert update test.vmware.com  --insecure true

Flags:
      --ca-cert string            path to the public certificate
  -h, --help                      help for update
      --insecure string           allow the use of http when interacting with the host (true|false)
      --skip-cert-verify string   skip server's TLS certificate verification (true|false)


./bin/tanzu config cert update test2.host.com --ca-cert                                                           
--ca-cert           -- path to the public certificate
--help              -- help for update
--insecure          -- allow the use of http when interacting with the host (true|false)
--skip-cert-verify  -- skip server's TLS certificate verification (true|false)

## validated the new command option --ca-cert works
❯ ./bin/tanzu config cert update test2.host.com --ca-cert  ~/openssl/server.crt
[ok] updated certificate data for host test2.host.com
❯ ./bin/tanzu config cert list
  HOST            CA-CERTIFICATE  SKIP-CERT-VERIFICATION  INSECURE
  test1.host.com  Not configured  false                   true
  test.host.com   <REDACTED>      false                   false
  test2.host.com  <REDACTED>      false                   false

### validated the deprecated command line option --ca-certificate still works with deprecation message
❯ ./bin/tanzu config cert update test2.host.com --ca-certificate  ~/openssl/server.crt
Flag --ca-certificate has been deprecated, this was done in the v1.1.0 release, it will be removed following the deprecation policy (6 months). Use the --ca-cert flag instead.

[ok] updated certificate data for host test2.host.com
```

### Release note

```release-note
Rename --ca-certificate option to --ca-cert in "tanzu config cert add" and "tanzu config cert update" commands. The --ca-certificate option is still supported but deprecated in both the commands.
```

### DIFF
--- a/pkg/command/cert.go
+++ b/pkg/command/cert.go
@@ -59,8 +59,13 @@ func newCertCmd() *cobra.Command {
 		return cobra.AppendActiveHelp(nil, "Please provide 'host' or 'host:port'"), cobra.ShellCompDirectiveNoFileComp
 	}))
 
-	// The completion for this flag is simple file completion, which is configured by default
+	// --ca-certificate is renamed to --ca-cert
 	addCertCmd.Flags().StringVarP(&caCertPathForAdd, "ca-certificate", "", "", "path to the public certificate")
+	caCertificateDeprecationMsg := "this was done in the v1.1.0 release, it will be removed following the deprecation policy (6 months). Use the --ca-cert flag instead.\n"
+	utils.PanicOnErr(addCertCmd.Flags().MarkDeprecated("ca-certificate", caCertificateDeprecationMsg))
+	// The completion for this flag is simple file completion, which is configured by default
+	addCertCmd.Flags().StringVarP(&caCertPathForAdd, "ca-cert", "", "", "path to the public certificate")
+
 	addCertCmd.Flags().StringVarP(&skipCertVerifyForAdd, "skip-cert-verify", "", "false", "skip server's TLS certificate verification")
 	utils.PanicOnErr(addCertCmd.RegisterFlagCompletionFunc("skip-cert-verify", compSkipFlag))
 
@@ -69,8 +74,12 @@ func newCertCmd() *cobra.Command {
 
 	utils.PanicOnErr(cobra.MarkFlagRequired(addCertCmd.Flags(), "host"))
 
-	// The completion for this flag is simple file completion, which is configured by default
+	// --ca-certificate is renamed to --ca-cert
 	updateCertCmd.Flags().StringVarP(&caCertPathForUpdate, "ca-certificate", "", "", "path to the public certificate")
+	utils.PanicOnErr(updateCertCmd.Flags().MarkDeprecated("ca-certificate", caCertificateDeprecationMsg))
+	// The completion for this flag is simple file completion, which is configured by default
+	updateCertCmd.Flags().StringVarP(&caCertPathForUpdate, "ca-cert", "", "", "path to the public certificate")
+
 	updateCertCmd.Flags().StringVarP(&skipCertVerifyForUpdate, "skip-cert-verify", "", "", "skip server's TLS certificate verification (true|false)")
 	utils.PanicOnErr(updateCertCmd.RegisterFlagCompletionFunc("skip-cert-verify", compSkipFlag))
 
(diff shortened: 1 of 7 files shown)


pkg/command/cert.go
```diff
@@ -59,8 +59,13 @@ func newCertCmd() *cobra.Command {
 		return cobra.AppendActiveHelp(nil, "Please provide 'host' or 'host:port'"), cobra.ShellCompDirectiveNoFileComp
 	}))
 
-	// The completion for this flag is simple file completion, which is configured by default
+	// --ca-certificate is renamed to --ca-cert
```
marckhouzam, line 62: From what I understand, this rename is to align with the terminology of the command which is `cert` and not `certificate`.  So this change is great.

What do you think however of using `--ca-cert-path`?
I find it would be clearer of what is expected as a flag value.
I compared with `imgpkg` and its similar flag is `--registry-ca-cert-path`; I don't think we need the `registry` prefix, but the `path` suffix may be useful.

marckhouzam (commented):
Would you mind replacing the `--ca-certificate` usage in the tests and docs?

prkalle:
> Would you mind replacing the `--ca-certificate` usage in the tests and docs?

Thanks for catching this. Updated the PR.

pkg/command/cert.go
prkalle: > What do you think however of using `--ca-cert-path`? I find it would be clearer of what is expected as a flag value. I compared with `imgpkg` and its similar flag is `--registry-ca-cert-path`; I don't think we need the `registry` prefix, but the `path` suffix may be useful.

I agree `-ca-cert-path` would be clear and helpful. However we have followed similar approach for `--*ca-certificate` for other command options for example [here](https://github.com/vmware-tanzu/tanzu-cli/blob/main/pkg/command/context.go#L210).

marckhouzam: Good point!  We need to consistent without our own tool.
Thanks for pointing this out to me.
Ok, so `--ca-cert` it is.

marckhouzam (approved):
LGTM
Thanks

PR #135, opened 2023-03-26
Fix generate-all-docs

vuil:
### What this PR does / why we need it

The core cli commands doc generation did not recurse down the root command tree. This change fixes that.
Did some cleanup and added missing unit test.

### Which issue(s) this PR fixes
Fixes #95

### Describe testing done for PR

```
go test ./pkg/command/... -run TestGenDocs
ok      github.com/vmware-tanzu/tanzu-cli/pkg/command   2.324s

> make build-all
> mkdir docs/cli/commands
# generate docs in docs/cli/commands for core commands and plugins installed
> ./bin/tanzu generate-all-docs
> find docs/cli/commands/ | sort
docs/cli/commands/
docs/cli/commands//README.md
docs/cli/commands//tanzu.md
docs/cli/commands//tanzu_builder.md
docs/cli/commands//tanzu_builder_cli.md
docs/cli/commands//tanzu_builder_cli_add-plugin.md
docs/cli/commands//tanzu_builder_cli_compile.md
docs/cli/commands//tanzu_builder_init.md
docs/cli/commands//tanzu_builder_inventory.md
docs/cli/commands//tanzu_builder_inventory_init.md
docs/cli/commands//tanzu_builder_inventory_plugin-group.md
docs/cli/commands//tanzu_builder_inventory_plugin-group_activate.md
docs/cli/commands//tanzu_builder_inventory_plugin-group_add.md
docs/cli/commands//tanzu_builder_inventory_plugin-group_deactivate.md
docs/cli/commands//tanzu_builder_inventory_plugin.md
docs/cli/commands//tanzu_builder_inventory_plugin_activate.md
docs/cli/commands//tanzu_builder_inventory_plugin_add.md
docs/cli/commands//tanzu_builder_inventory_plugin_deactivate.md
docs/cli/commands//tanzu_builder_plugin.md
docs/cli/commands//tanzu_builder_plugin_build-package.md
docs/cli/commands//tanzu_builder_plugin_build.md
docs/cli/commands//tanzu_builder_plugin_publish-package.md
docs/cli/commands//tanzu_builder_publish.md
docs/cli/commands//tanzu_completion.md
docs/cli/commands//tanzu_config.md
docs/cli/commands//tanzu_config_get.md
docs/cli/commands//tanzu_config_init.md
docs/cli/commands//tanzu_config_server.md
docs/cli/commands//tanzu_config_server_delete.md
docs/cli/commands//tanzu_config_server_list.md
docs/cli/commands//tanzu_config_set.md
[...]

PR #580, opened 2023-10-30
Fix unit tests

marckhouzam:

Fix the gen_docs unit tests by setting a temporary environment.

After the merge of #135 the unit tests have started failing because the prompts were not being handled.
Furthermore, by not having a set temporary configuration files, the test would re-use the local environment which is problematic when running it on our own machines because we have different plugins installed.

Fixes # N/A

Ran the test locally

### Release note

```release-note

```


prkalle (approved):
LGTM. Thanks for the changes!

PR #604, opened 2023-12-02
Refresh the plugin DB on `plugin source init`

marckhouzam:

This PR forces a refresh of the plugin inventory DB on a `tanzu plugin source init`.

Not refreshing the plugin inventory DB on `plugin source init` can cause confusion for some situations.  For example, shell completion, which only uses the inventory cache, would continue using the old cache after a `plugin source init`; similarly, the installation of the essential plugins would not get updated until the DB was refreshed through another command.

Furthermore, this becomes even more important as we plan to improve the DB cache to not check the digest for a certain period of time; in such a case, after running a `plugin source init`, the DB could be stale for all commands.

Fixes # N/A

```
# Set the plugin source to something else than the central repo
[...]
$ tz plugin source update default --uri localhost:9876/tanzu-cli/plugins/central:small

# Notice that shell completion does not find any "cost" plugin since there is none in this repo
# We do this so that we can check later that shell completion does find it.
$ tz __complete plugin install ''|grep cost

# Check the digest of this plugin source
$ ls -l /Users/kmarc/.cache/tanzu/plugin_inventory/default/digest.*
-rw-r--r--  1 kmarc  staff     0B  2 Dec 15:04 /Users/kmarc/.cache/tanzu/plugin_inventory/default/digest.793fbad3e13912b6d1432f3383d7bd2de6ebdbf793706e4dc3dd45bb35df5239

# Notice that the new DB is being read right way
$ tz plugin source init
[ok] successfully initialized discovery source

# Notice the new digest
$ ls -l /Users/kmarc/.cache/tanzu/plugin_inventory/default/digest.*
-rw-r--r--  1 kmarc  staff     0B  2 Dec 15:05 /Users/kmarc/.cache/tanzu/plugin_inventory/default/digest.3bda1cf622a73873e51804cefdc6a246ad955de079ccfb7c49016418fe65be12

# Notice that shell completion immediately know about the new content of the DB
$ tz __complete plugin install ''|grep cost
cost	A cost object for cluster and clustergroup
```

### Release note

```release-note
Update the plugin inventory DB on a `tanzu plugin source init`
```

### DIFF
--- a/pkg/command/discovery_source.go
+++ b/pkg/command/discovery_source.go
@@ -165,6 +165,14 @@ func newInitDiscoverySourceCmd() *cobra.Command {
 				return err
 			}
 
+			// Refresh the inventory DB
+			if discoverySource, err := configlib.GetCLIDiscoverySource(config.DefaultStandaloneDiscoveryName); err == nil {
+				// Ignore any failures since the real operation
+				// the user is trying to do is set the config
+				// to the central repo, which was done above
+				_ = checkDiscoverySource(*discoverySource)
+			}
+
 			log.Successf("successfully initialized discovery source")
 			return nil
 		},
--- a/pkg/command/discovery_source_test.go
+++ b/pkg/command/discovery_source_test.go
@@ -7,11 +7,13 @@ import (
 	"bytes"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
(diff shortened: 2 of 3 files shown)


anujc25 (approved):
LGTM. Thanks

Issue #357, opened 2023-06-15
Support arm64 for core and plugins


scothis:
**Describe the feature request**

AMR CPUs are increasing in popularity for servers and desktop use, especially with the introduction of Apple Silicon. The Tanzu CLI not supporting this popular architecture creates unnecessary friction for users of newer Macs.

**Describe alternative(s) you've considered**

The Tanzu CLI can run under rosetta, but locally compiled plugins are built for the native host architecture by default `darwin_arm64`, the `tanzu plugin install` command insists on installing `darwin_amd64` binaries which are not built so the install fails.

Compiling the Tanzu CLI directly for `darwin_arm64` also works, but installing pre-built plugins will fail unless they are compiled for `darwin_arm64`, which they are not.

**Additional context**

Macs are a bit special in that they can execute either `darwin_arm64` (native) or `darwin_amd64` (emulated). When installing a plugin for a `darwin_arm64` if the plugin is not available as `darwin_arm64`, it should automatically fall back to `darwin_amd64`.

Previously https://github.com/vmware-tanzu/tanzu-framework/issues/440

PR #617, opened 2023-12-18
Build a Chocolatey package that supports ARM64

marckhouzam:

This PR updates the Chocolatey package to be able to install a Windows ARM64 CLI build when running on an ARM64 Windows.

To do this, the build script now handles two binaries, the one for AMD64 and the one for ARM64, each one with its own SHA.  Both SHAs are injected in the `chocolateyInstall.ps1.tmpl` template file which is then used as the installation script.  At installation time, this script checks if its running on an ARM64 machine or an AMD64 machine and then downloads the appropriate CLI Windows binary.

Part of #357 

This is hard to test directly because there is no Windows ARM64 version currently published on this repo's Github release page.  So what I did was pretend that the `v1.0.0` AMD64 was the ARM64 version, and that the `v1.1.0` AMD64 version was the AMD64 version.  I built the new Chocolatey package with those "hacks" and then ran it on:
1. Windows 11 AMD64 and confirmed the AMD64 CLI was installed (so `v1.1.0`)
2. Windows 11 ARM64 and confirmed the ARM64 CLI was installed (so `v1.0.0`)
3. Windows 10 AMD64 and confirmed the AMD64 CLI was installed (so `v1.1.0`)
 
I also tested building the new Chocolatey package using VMware's internal build system and confirmed the proper installation script was generated using the expected SHAs for AMD64 and ARM64.

### Release note

```release-note
Add support to install the Windows ARM64 CLI through Chocolatey.
```

The script assumes that the Github binary for Windows ARM64 is named `tanzu-cli-windows-arm64-windows11.zip`.
This may need a bit of discussion.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,7 +207,12 @@ choco-package: ## Build a Chocolatey package
 	fi
 	@# The nuspec file uses a variable but variables don't seem to work anymore
 	@# with chocolatey 2.0.0 so we continue using version 1.4.0
-	docker run --rm -e VERSION=$(BUILD_VERSION) -e SHA_FOR_CHOCO=$(SHA_FOR_CHOCO) -v $(ROOT_DIR):$(ROOT_DIR) $(CHOCO_IMAGE) $(ROOT_DIR)/hack/choco/build_package.sh
+	docker run --rm \
+		-e VERSION=$(BUILD_VERSION) \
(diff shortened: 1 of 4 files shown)


anujc25 (approved):
LGTM.

PR #637, opened 2024-01-08
Implement refresh of database cache based on digest timestamp

mpanchajanya:

- Implement RefreshDatabase function that verifies the discovery source digest file last timestamp and refreshes the database if digest timestamp is past 24 hours
- The default threshold of 24 hours can be overridden using a env variable `TANZU_CLI_PLUGIN_DB_CACHE_REFRESH_THRESHOLD`

Fixes #

1. tanzu plugin clean and next tanzu command refreshes the db

```
> tz plugin clean
> tz plugin list
[i] Refreshing plugin inventory cache for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.
[i] The tanzu cli essential plugins have not been installed and are being installed now. The install may take a few seconds.
[i] Installing plugins from plugin group 'vmware-tanzucli/essentials:v1.0.0'
[i] Installing plugin 'telemetry:v1.1.0' with target 'global' 

Standalone Plugins

```

### Release note

```release-note
Implement automatic refresh of database repository cache
```

### DIFF
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -187,15 +187,15 @@ func newRootCmd() *cobra.Command {
 				}
 			}
 
-			// Install or update essential plugins
-			InstallEssentialPlugins(cmd)
-
 			// Prompt for CEIP agreement
 			if !shouldSkipPrompts(cmd) {
 				if err := cliconfig.ConfigureCEIPOptIn(); err != nil {
 					return err
 				}
 			}
+
+			// Install or update essential plugins
+			InstallEssentialPlugins(cmd)
 
 			setupActiveHelp(cmd, args)
 
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -20,6 +20,7 @@ import (
 	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
 	cliconfig "github.com/vmware-tanzu/tanzu-cli/pkg/config"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/constants"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/discovery"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/pluginmanager"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/pluginsupplier"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/telemetry"
@@ -237,6 +238,9 @@ func InstallEssentialPlugins(cmd *cobra.Command) {
 		}
 	}
 	if !skipEssentials {
+		// Refresh the database
+		_ = discovery.RefreshDatabase()
+
 		// Check if essential plugins are installed and up to date if not Install or Upgrade the Essential plugins
 		_, _ = pluginmanager.InstallPluginsFromEssentialPluginGroup()
 	}
--- a/pkg/constants/defaults.go
+++ b/pkg/constants/defaults.go
@@ -3,6 +3,10 @@
 
 package constants
 
+import (
+	"time"
+)
+
 const (
 	// TanzuCLISystemNamespace  is the namespace for tanzu cli resources
 	TanzuCLISystemNamespace = "tanzu-cli-system"
@@ -22,4 +26,7 @@ const (
 
 	// DefaultCLIEssentialsPluginGroupName  name of the essentials plugin group which is used to install essential plugins
 	DefaultCLIEssentialsPluginGroupName = "vmware-tanzucli/essentials"
+
+	// DefaultDigestExpirationThreshold is the default value for digest expiration threshold
+	DefaultDigestExpirationThreshold = 24 * time.Hour
 )
--- a/pkg/constants/defaults.go
+++ b/pkg/constants/defaults.go
@@ -27,6 +27,6 @@ const (
 	// DefaultCLIEssentialsPluginGroupName  name of the essentials plugin group which is used to install essential plugins
 	DefaultCLIEssentialsPluginGroupName = "vmware-tanzucli/essentials"
 
-	// DefaultDigestExpirationThreshold is the default value for digest expiration threshold
-	DefaultDigestExpirationThreshold = 24 * time.Hour
+	// DefaultPluginDBCacheRefreshThreshold is the default value for db cache refresh
+	DefaultPluginDBCacheRefreshThreshold = 24 * time.Hour
 )
--- a/pkg/constants/env_variables.go
+++ b/pkg/constants/env_variables.go
@@ -53,4 +53,5 @@ const (
 
 	// TanzuCLIOAuthLocalListenerPort is the port to be used by local listener for OAuth authorization flow
 	TanzuCLIOAuthLocalListenerPort = "TANZU_CLI_OAUTH_LOCAL_LISTENER_PORT"
+	TanzuCLIDBRefreshThreshold = "TANZU_CLI_DATABASE_REFRESH_THRESHOLD"
 )
--- a/pkg/constants/env_variables.go
+++ b/pkg/constants/env_variables.go
@@ -47,7 +47,11 @@ const (
 	ConfigVariablePluginDBCacheTTL = "TANZU_CLI_PLUGIN_DB_CACHE_TTL_SECONDS"
 
 	// ConfigVariablePluginDBCacheRefreshThreshold Change the default value of db cache refresh threshold
-	ConfigVariablePluginDBCacheRefreshThreshold = "TANZU_CLI_PLUGIN_DB_CACHE_REFRESH_THRESHOLD_HOURS"
+	// A duration string is a possibly signed sequence of
+	// decimal numbers, each with optional fraction and a unit suffix,
+	// such as "300ms", "-1.5h" or "2h45m".
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+	ConfigVariablePluginDBCacheRefreshThreshold = "TANZU_CLI_PLUGIN_DB_CACHE_REFRESH_THRESHOLD"
 
 	// CSPLoginOrgID overrides the CSP default OrgID to which the user logs into, using CLI interactive login flow
 	// Note: More information regarding the CSP organizations can be found at
--- a/pkg/constants/env_variables.go
+++ b/pkg/constants/env_variables.go
@@ -46,6 +46,9 @@ const (
 	// Change the default value of the plugin inventory cache TTL
 	ConfigVariablePluginDBCacheTTL = "TANZU_CLI_PLUGIN_DB_CACHE_TTL_SECONDS"
 
+	// ConfigVariablePluginDBCacheRefreshThreshold Change the default value of db cache refresh threshold
+	ConfigVariablePluginDBCacheRefreshThreshold = "TANZU_CLI_PLUGIN_DB_CACHE_REFRESH_THRESHOLD_HOURS"
+
 	// CSPLoginOrgID overrides the CSP default OrgID to which the user logs into, using CLI interactive login flow
 	// Note: More information regarding the CSP organizations can be found at
 	// https://docs.vmware.com/en/VMware-Cloud-services/services/Using-VMware-Cloud-Services/GUID-CF9E9318-B811-48CF-8499-9419997DC1F8.html
@@ -53,5 +56,4 @@ const (
 
 	// TanzuCLIOAuthLocalListenerPort is the port to be used by local listener for OAuth authorization flow
 	TanzuCLIOAuthLocalListenerPort = "TANZU_CLI_OAUTH_LOCAL_LISTENER_PORT"
-	TanzuCLIDBRefreshThreshold = "TANZU_CLI_DATABASE_REFRESH_THRESHOLD"
 )
--- a/pkg/discovery/utils.go
+++ b/pkg/discovery/utils.go
@@ -4,7 +4,17 @@
 package discovery
 
 import (
+	"bufio"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/constants"
+
 	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
 	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 )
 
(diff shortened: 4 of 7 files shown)


marckhouzam (commented):
Some comment to start

pkg/command/root.go
```diff
@@ -187,6 +190,52 @@ func newRootCmd() *cobra.Command {
...
 
+func RefreshDatabaseCache(cmd *cobra.Command) {
+	skipCommandsForDatabaseRefresh := []string{
+		// The shell completion setup is not interactive, so it should not trigger a prompt
+		"tanzu __complete",
+		"tanzu completion",
+		// Common first command to run,
+		"tanzu version",
+		// It would be a chicken and egg issue if user tries to set CEIP configuration
+		// using "tanzu config set env.TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER yes"
```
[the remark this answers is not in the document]
mpanchajanya: refactored the implementation

pkg/command/root.go
```diff
@@ -164,6 +164,9 @@ func newRootCmd() *cobra.Command {
 			// Install or update essential plugins
 			InstallEssentialPlugins(cmd)
 
+			// Verify and Perform Scheduled Refresh of the database cache
+			RefreshDatabaseCache(cmd)
```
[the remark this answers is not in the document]
mpanchajanya: Move the RefreshDatabase inside InstallEssentialPlugins

go.mod
```diff
@@ -6,6 +6,8 @@ replace cloud.google.com/go => cloud.google.com/go v0.102.1
 
 replace github.com/vmware-tanzu/tanzu-cli/test/e2e/framework => ./test/e2e/framework
 
+replace github.com/vmware-tanzu/tanzu-plugin-runtime => ../tanzu-plugin-runtime
```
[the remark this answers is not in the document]
mpanchajanya: removed no longer needed

pkg/constants/defaults.go
```diff
@@ -22,4 +26,7 @@ const (
 
 	// DefaultCLIEssentialsPluginGroupName  name of the essentials plugin group which is used to install essential plugins
 	DefaultCLIEssentialsPluginGroupName = "vmware-tanzucli/essentials"
+
+	// DefaultDigestExpirationThreshold is the default value for digest expiration threshold
+	DefaultDigestExpirationThreshold = 24 * time.Hour
```
marckhouzam, line 31: We need a more specific name I believe.  The digest expires after the 30 minutes TTL but the CLI won't notice until a command that needs the DB is run.  This 24 hours will force a digest refresh on any (almost) command of the CLI.  I like the naming you used for the env var below, so how about `DefaultPluginDBCacheRefreshThreshold`?

pkg/constants/env_variables.go
```diff
@@ -49,5 +49,6 @@ const (
 	// CSPLoginOrgID overrides the CSP default OrgID to which the user logs into, using CLI interactive login flow
 	// Note: More information regarding the CSP organizations can be found at
 	// https://docs.vmware.com/en/VMware-Cloud-services/services/Using-VMware-Cloud-Services/GUID-CF9E9318-B811-48CF-8499-9419997DC1F8.html
-	CSPLoginOrgID = "TANZU_CLI_CLOUD_SERVICES_ORGANIZATION_ID"
+	CSPLoginOrgID              = "TANZU_CLI_CLOUD_SERVICES_ORGANIZATION_ID"
+	TanzuCLIDBRefreshThreshold = "TANZU_CLI_DATABASE_REFRESH_THRESHOLD"
```
marckhouzam, line 53: Can you move this right after `ConfigVariablePluginDBCacheTTL` and use a similar naming such as `ConfigVariablePluginDBCacheRefreshThreshold = "TANZU_CLI_PLUGIN_DB_CACHE_REFRESH_THRESHOLD_HOURS"`

pkg/discovery/utils.go
```diff
@@ -70,3 +80,96 @@ func compareRESTDiscoverySources(ds1, ds2 configtypes.PluginDiscovery) bool {
...
+		pluginDataDir := filepath.Join(common.DefaultCacheDir, common.PluginInventoryDirName, name)
+		matches, _ := filepath.Glob(filepath.Join(pluginDataDir, "digest.*"))
+
+		if len(matches) == 1 {
+			file, err := os.Open(matches[0])
+			if err != nil {
+				return err
+			}
+
+			scanner := bufio.NewScanner(file)
```
marckhouzam, line 142: I don't think you need to check the URI at all in this case.
Whenever we change a discovery source URI (`tz plugin source update`) we immediately refresh the cache, and therefore the URI will always match with the cache.
The time it is useful to check the URI is for test discoveries found in `TANZU_CLI_ADDITIONAL_PLUGIN_DISCOVERY_IMAGES_TEST_ONLY`, however those are not part of the config file and therefore this loop will not find them.

So there is no value in checking the URI.  You can simply check the timestamp of the `digest.*` for each discovery.

As for the test discoveries, I think this feature is not required for test discoveries.

pkg/discovery/utils.go
```diff
@@ -70,3 +80,96 @@ func compareRESTDiscoverySources(ds1, ds2 configtypes.PluginDiscovery) bool {
 		ds1.REST.BasePath == ds2.REST.BasePath &&
 		ds1.REST.Endpoint == ds2.REST.Endpoint
 }
+
+func getDiscoverySourceNameAndURL(source configtypes.PluginDiscovery) (string, string, error) {
+	var name string
+	var url string
+	switch {
+	case source.OCI != nil:
```
marckhouzam, line 88: We only support OCI, so you could actually use that without the others below

marckhouzam (commented):
This is much simpler than before 👏 

I put one comment that could simplify things further

1. The PR needs a manual rebase.
2. It would be nice to add an E2E test.  I suggest:
     - doing `tz plugin source init`
     - setting the `TANZU_CLI_PLUGIN_DB_CACHE_TTL_SECONDS` variable to something like 5 seconds
     - setting your new variable `TANZU_CLI_DATABASE_REFRESH_THRESHOLD` (using its new name) to something like 10 seconds
     - and running a `tz plugin list` every 2 seconds and looking for the DB refresh printout after the 5th command.  You can base yourself on the test case starting with `// use case: Plugin inventory DB digest check is only done once its TTL has expired` in `test/e2e/plugin_lifecycle/plugin_lifecycle_test.go`

pkg/constants/defaults.go
mpanchajanya: Thanks , Renamed the variable

pkg/constants/env_variables.go
mpanchajanya: Renamed and moved as suggested

pkg/discovery/utils.go
mpanchajanya: Updated the switch to only use oci and default

pkg/discovery/utils.go
mpanchajanya: Updated the function to only check for digest timestamp

test/e2e/plugin_lifecycle/plugin_lifecycle_test.go
```diff
@@ -618,4 +618,45 @@ var _ = framework.CLICoreDescribe("[Tests:E2E][Feature:Plugin-lifecycle]", func(
...
+			refreshingPluginInventoryCache = "Refreshing plugin inventory cache"
+		)
+
+		It("plugin list doesnt refresh db cache until threshold is passed", func() {
+
+			// Set the TTL to something small: 5 seconds
+			_ = os.Setenv(constants.ConfigVariablePluginDBCacheTTL, "5")
+			_ = os.Setenv(constants.ConfigVariablePluginDBCacheRefreshThreshold, "5s")
+
+			for i := 0; i < 5; i++ {
```
marckhouzam, line 641: I suggest looping 8 times, so the TTL is expired and that we have seen that at second 6 and 7 and 8, there was no refresh due to the threshold (assuming a 10 sec threshold set above)

test/e2e/plugin_lifecycle/plugin_lifecycle_test.go
```diff
@@ -618,4 +618,45 @@ var _ = framework.CLICoreDescribe("[Tests:E2E][Feature:Plugin-lifecycle]", func(
...
+				time.Sleep(time.Second * 1) // Sleep for 1 second
+				_, _, errStream, err := tf.PluginCmd.ListPlugins()
+				Expect(err).To(BeNil())
+				// No printouts on the error stream about refreshing the DB
+				Expect(errStream).ToNot(ContainSubstring(refreshingPluginInventoryCache))
+				Expect(errStream).ToNot(ContainSubstring(refreshingDBPrintout))
+			}
```
marckhouzam, line 648: Now 8 seconds have gone by, the TTL is expired but the threshold is not.  We should sleep 3 or 4 seconds to let the threshold expire

test/e2e/plugin_lifecycle/plugin_lifecycle_test.go
```diff
@@ -618,4 +618,45 @@ var _ = framework.CLICoreDescribe("[Tests:E2E][Feature:Plugin-lifecycle]", func(
...
+			_, _, errStream, err := tf.PluginCmd.ListPlugins()
+			Expect(err).To(BeNil())
+			// Now we expect printouts on the error stream about refreshing the DB
+			Expect(errStream).To(ContainSubstring(refreshingPluginInventoryCache))
+			Expect(errStream).To(ContainSubstring(refreshingDBPrintout))
```
marckhouzam, line 653: This second one won't happen because the actual digest value has not changed, so there is no need to download the DB.  for the test, that is fine, all we need to see is the first printout confirming the DB cache is checked (and the DB would be downloaded if it was needed)

marckhouzam (approved):
LGTM, thanks!

I put some comments to help you get the test working since I had to fight with my own test when I did something similar.

test/e2e/plugin_lifecycle/plugin_lifecycle_test.go
```diff
@@ -618,4 +618,45 @@ var _ = framework.CLICoreDescribe("[Tests:E2E][Feature:Plugin-lifecycle]", func(
 			}
 		})
 	})
+
+	// use case: Plugin inventory DB cache refresh check on every tanzu command
+	// and perform refresh only when digest timestamp is passed 24 hours or env variable TANZU_CLI_DATABASE_REFRESH_THRESHOLD is set
+	// a. run a "tanzu plugin source init" to do a digest check
+	// b. set the TTL to 5 seconds
+	// c. set the TANZU_CLI_DATABASE_REFRESH_THRESHOLD to 10 seconds
```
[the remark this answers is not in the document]
mpanchajanya: updated

test/e2e/cataloge2e/catalog_update_parallel_test.go
```diff
@@ -24,7 +24,7 @@ var _ = framework.CLICoreDescribe("[Tests:E2E][Feature:Catalog-Update]", func()
 
 			pluginsList, err := framework.GetPluginsList(tf, true)
 			Expect(err).To(BeNil(), "should not get any error for plugin list")
-			Expect(len(pluginsList)).Should(BeNumerically("==", 0), "plugins list should not return any plugins after plugin clean")
+			Expect(len(pluginsList)).Should(BeNumerically("==", 1), "plugins list should return only one essential plugin after plugin clean")
```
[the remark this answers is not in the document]
mpanchajanya: After this refresh change when we run
tanzu plugin clean
tanzu plugin list (This step will install essential plugins i.e. telemetry for now since the refresh of db runs if there is no digest file found)

test/e2e/plugin_lifecycle/plugin_lifecycle_test.go
mpanchajanya: Thanks. Updated the test case

test/e2e/plugin_lifecycle/plugin_lifecycle_test.go
mpanchajanya: Updated the test case

test/e2e/plugin_lifecycle/plugin_lifecycle_test.go
mpanchajanya: Thanks for the suggestion, updated the test accordingly.

test/e2e/plugin_lifecycle/plugin_lifecycle_test.go
```diff
@@ -618,4 +618,45 @@ var _ = framework.CLICoreDescribe("[Tests:E2E][Feature:Plugin-lifecycle]", func(
...
+			refreshingDBPrintout           = "Reading plugin inventory for"
+			refreshingPluginInventoryCache = "Refreshing plugin inventory cache"
+		)
+
+		It("plugin list doesnt refresh db cache until threshold is passed", func() {
+
+			// Set the TTL to something small: 5 seconds
+			_ = os.Setenv(constants.ConfigVariablePluginDBCacheTTL, "5")
+			_ = os.Setenv(constants.ConfigVariablePluginDBCacheRefreshThreshold, "5s")
+
```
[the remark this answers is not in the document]
mpanchajanya: This helps. Updated the test

test/e2e/Makefile
```diff
@@ -91,6 +91,7 @@ e2e-plugin-compatibility-tests:
 .PHONY: e2e-plugin-lifecycle-tests ## Execute CLI Core Plugin life cycle E2E test cases
 e2e-plugin-lifecycle-tests:
 	export TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL) ; \
+#	export TANZU_CLI_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_VERIFICATION_SKIP_LIST=$(TANZU_CLI_E2E_TEST_LOCAL_CENTRAL_REPO_URL) ; \
```
marckhouzam, line 94: We don't need this, right?

test/e2e/cataloge2e/catalog_update_parallel_test.go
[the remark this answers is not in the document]
marckhouzam: Oh right, that makes sense.  Thanks

marckhouzam (commented):
When trying out this change with a brand new installation, I noticed that the essential plugins get installed in between the two prompts (after the eula but before the telemetry).  I wonder if it would be better to have the two prompts follow each other and then deal with the essential plugins?

```
$ tz plugin list
? You must agree to the VMware General Terms in order to download, install, or
use software from this registry via Tanzu CLI. Acceptance of the VMware General
Terms covers all software installed via the Tanzu CLI during any Session.
“Session” means the period from acceptance until any of the following occurs:
(1) a change to VMware General Terms, (2) a new major release of the Tanzu CLI
is installed, (3) software is accessed in a separate software distribution
registry, or (4) re-acceptance of the General Terms is prompted by VMware.

To view the VMware General Terms, please see https://www.vmware.com/vmware-general-terms.html.

If you agree, the essential plugins (required by the tanzu cli) will be automatically installed.

Note: this prompt can be avoided by running "tanzu config eula accept".

 Yes

==
[i] Refreshing plugin inventory cache for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.
[i] The tanzu cli essential plugins have not been installed and are being installed now. The install may take a few seconds.
[i] Installing plugins from plugin group 'vmware-tanzucli/essentials:v1.0.0'
[i] Installing plugin 'telemetry:v1.1.0' with target 'global'

? VMware's Customer Experience Improvement Program ("CEIP") provides VMware with
information that enables VMware to improve its products and services and fix
problems. By choosing to participate in CEIP, you agree that VMware may collect
technical information about your use of VMware products and services on a
regular basis. This information does not personally identify you.

For more details about the program, please see https://www.vmware.com/trustvmware/ceip.html.

Note: this prompt can be avoided by setting the environment variable
TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER to "yes" or "no".

Do you agree to participate in the Customer Experience Improvement Program?
 Yes
```

mpanchajanya:
> When trying out this change with a brand new installation, I noticed that the essential plugins get installed in between the two prompts (after the eula but before the telemetry). I wonder if it would be better to have the two prompts follow each other and then deal with the essential plugins?
> 
> ```
> $ tz plugin list
> ? You must agree to the VMware General Terms in order to download, install, or
> use software from this registry via Tanzu CLI. Acceptance of the VMware General
> Terms covers all software installed via the Tanzu CLI during any Session.
> “Session” means the period from acceptance until any of the following occurs:
> (1) a change to VMware General Terms, (2) a new major release of the Tanzu CLI
> is installed, (3) software is accessed in a separate software distribution
> registry, or (4) re-acceptance of the General Terms is prompted by VMware.
> 
> To view the VMware General Terms, please see https://www.vmware.com/vmware-general-terms.html.
> 
> If you agree, the essential plugins (required by the tanzu cli) will be automatically installed.
> 
> Note: this prompt can be avoided by running "tanzu config eula accept".
> 
> Do you agree to the VMware General Terms?
>  Yes
> 
> ==
> [i] Refreshing plugin inventory cache for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.
> [i] Reading plugin inventory for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.
> [i] The tanzu cli essential plugins have not been installed and are being installed now. The install may take a few seconds.
> [i] Installing plugins from plugin group 'vmware-tanzucli/essentials:v1.0.0'
> [i] Installing plugin 'telemetry:v1.1.0' with target 'global'
> 
> ? VMware's Customer Experience Improvement Program ("CEIP") provides VMware with
> information that enables VMware to improve its products and services and fix
> problems. By choosing to participate in CEIP, you agree that VMware may collect
> technical information about your use of VMware products and services on a
> regular basis. This information does not personally identify you.
> 
> For more details about the program, please see https://www.vmware.com/trustvmware/ceip.html.
> 
> Note: this prompt can be avoided by setting the environment variable
> TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER to "yes" or "no".
> 
> Do you agree to participate in the Customer Experience Improvement Program?
>  Yes
> ```

Yes This makes sense. Moved the installation of essential plugins after both eula and ceip prompts

marckhouzam (approved):
LGTM
Thanks!

PR #649, opened 2024-01-17
Add show details reminder in group search output

vuil:

Shows a notes at the bottom of the group search tabular output about using the --show-details flag to see all versions of the groups.

Fixes #

Added unit tests.

Output changes are as follows:
```
> tanzu plugin group search
  GROUP                       DESCRIPTION                          LATEST
  vmware-tanzucli/essentials  Essential plugins for the Tanzu CLI  v1.0.0
  vmware-tap/default          Plugins for TAP                      v1.7.3
  vmware-tkg/default          Plugins for TKG                      v2.4.1
  vmware-tmc/default          Plugins for TMC                      v1.0.0

Note: to view all plugin group versions available, use 'tanzu plugin group search --show-details'
>
```

### Release note

```release-note
Added note in "tanzu plugin group search" about using --show-details to see all versions of groups 
```

### DIFF
--- a/pkg/command/plugin_group.go
+++ b/pkg/command/plugin_group.go
@@ -28,6 +28,8 @@ var (
 	showNonMandatory bool
 )
 
+const groupSearchShowDetailsMsg = "Note: To view all plugin group versions available, use 'tanzu plugin group search --show-details'."
+
 func newPluginGroupCmd() *cobra.Command {
 	var pluginGroupCmd = &cobra.Command{
 		Use:   "group",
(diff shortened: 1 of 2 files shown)


pkg/command/plugin_group.go
```diff
@@ -28,6 +28,8 @@ var (
 	showNonMandatory bool
 )
 
+const groupSearchShowDetailsMsg = "Note: to view all plugin group versions available, use 'tanzu plugin group search --show-details'"
```
marckhouzam, line 31: Nit: all other notes start their sentence with an uppercase letter.
(Except https://github.com/vmware-tanzu/tanzu-cli/blob/af5d92dce7321228bfd913e15976edcaedb2f0d8/pkg/config/config.go#L33 and https://github.com/vmware-tanzu/tanzu-cli/blob/af5d92dce7321228bfd913e15976edcaedb2f0d8/pkg/config/config.go#L51)

marckhouzam (approved):
One nit, and the release notes sentence seems either incomplete or has an extra `in` at the end.

Besides that LGTM

pkg/command/plugin_group.go
vuil: Thanks for noticing! Fixed.

marckhouzam (approved):

PR #660, opened 2024-01-23
Fix: Catch the interrupt signal when spinner is running

chandrareddyp:

This PR addresses a bug that occurs when capturing the interrupt signal (Ctrl+C) during plugin installation. It ensures that the spinner is terminated gracefully and logs a final message indicating that a specific plugin failed to install. Prior to this fix, when a user interrupted a signal while the spinner was active during plugin installation, the terminal would hang, preventing the previous command from exiting properly.

Fixes #

Manual testing has done:
1. Power-Shell: `tanzu context use` output:
image
2. Z Shell:  `tanzu context use` output:
image

### Release note

```release-note

```

### DIFF
--- a/pkg/pluginmanager/manager.go
+++ b/pkg/pluginmanager/manager.go
@@ -746,23 +746,25 @@ func installOrUpgradePlugin(p *discovery.Discovered, version string, installTest
 	// Log message based on different installation conditions
 	installingMsg, installedMsg, errMsg := getPluginInstallationMessage(p, version, plugin != nil, isPluginAlreadyInstalled)
 
-	var spinnerErr, pluginErr error
+	var spinnerErr error
 	var sow component.OutputWriterSpinner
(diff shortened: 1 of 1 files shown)


pkg/pluginmanager/manager.go
```diff
@@ -746,23 +746,26 @@ func installOrUpgradePlugin(p *discovery.Discovered, version string, installTest
 	// Log message based on different installation conditions
 	installingMsg, installedMsg, errMsg := getPluginInstallationMessage(p, version, plugin != nil, isPluginAlreadyInstalled)
 
-	var spinnerErr, pluginErr error
+	var spinnerErr error
```
marckhouzam, line 749: This variable and the `signalChannel` one can go in the loop below for simplicity.

marckhouzam (approved):
Nice!

PR #666, opened 2024-02-01
fix: return error if input file for --to-tar exists

chandrareddyp:
Previously, the `tanzu plugin download-bundle --to-tar` command would overwrite the output file if an existing file was specified. This PR implements a fix to throw an error if the target file already exists when attempting the bundle download. This change prevents accidentally overriding content in an existing file.

Fixes #

1)Manual Testing has done.
image

2)E2E tests are updated.

### Release note

```release-note
The `tanzu plugin download-bundle --to-tar` command now prevents overwriting an existing output file.
```

### DIFF
--- a/test/e2e/airgapped/airgapped_test.go
+++ b/test/e2e/airgapped/airgapped_test.go
@@ -19,6 +19,8 @@ import (
 )
 
 const InvalidPath = "invalid path for \"%s\""
+const fileExists = "the file '%s' already exists"
+const showThrowErr = "should throw error for incorrect input path"
 
 var _ = framework.CLICoreDescribe("[Tests:E2E][Feature:Airgapped-Plugin-DownloadBundle-UploadBundle-Lifecycle]", func() {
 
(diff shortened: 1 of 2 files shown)


pkg/command/plugin_bundle.go
```diff
@@ -51,6 +53,9 @@ to an internet-restricted environment. Please also see the "upload-bundle" comma
 			if !dpbo.dryRun && dpbo.tarFile == "" {
 				return errors.New("flag '--to-tar' is required")
 			}
+			if _, err := os.Stat(dpbo.tarFile); err == nil {
```
marckhouzam, line 56: I'm curious, why do the check here instead of in `verifyTarFile()` where we already do some validation?

marckhouzam (commented):
Could you just change the release note sentence to change "existing target files" to "an existing output file"?

pkg/command/plugin_bundle.go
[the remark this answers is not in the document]
chandrareddyp: Yes it causes issue in case of dry-run, i have updated the fix.

test/e2e/airgapped/airgapped_test.go
```diff
@@ -328,20 +330,32 @@ var _ = framework.CLICoreDescribe("[Tests:E2E][Feature:Airgapped-Plugin-Download
...
 			err := tf.PluginCmd.DownloadPluginBundle(e2eTestLocalCentralRepoImage, []string{}, "")
-			Expect(err).NotTo(BeNil(), "should throw error for incorrect input path")
+			Expect(err).NotTo(BeNil(), showThrowErr)
 			Expect(strings.Contains(err.Error(), "flag '--to-tar' is required")).To(BeTrue())
 		})
 		// Test case: (negative use case) directory name only for --to-tar
 		It("plugin download-bundle when to-tar path is a directory", func() {
-			err := tf.PluginCmd.DownloadPluginBundle(e2eTestLocalCentralRepoImage, []string{}, tempDir)
-			Expect(err).NotTo(BeNil(), "should throw error for incorrect input path")
-			Expect(strings.Contains(err.Error(), fmt.Sprintf(InvalidPath, tempDir))).To(BeTrue())
```
marckhouzam, line 338: For simplicity, why not use `tempDir` as the directory instead of creating a new one?

PR #671, opened 2024-02-07
Feature/e2e on windows

chandrareddyp:

Fixes #

### Release note

```release-note

```

### DIFF
--- a/.github/workflows/cli_core_e2e_windows_test_Cygwin.yaml
+++ b/.github/workflows/cli_core_e2e_windows_test_Cygwin.yaml
@@ -1,0 +1,56 @@
+name: Cygwin###Tanzu CLI Core E2E Tests on Windows
+
+on:
+  pull_request:
+    branches: [main, release-*]
+  push:
+    branches: [ main, test*, release-*]
+
+jobs:
+  build:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -232,30 +274,56 @@ test-with-summary-report: tools
...
+		-v "D:\a\tanzu-cli\tanzu-cli\hack\central-repo\registry-content:C:\registry" \
+		-v "D:\a\tanzu-cli\tanzu-cli\hack\central-repo\certs:C:\certs" \
+		stefanscherer/registry-windows:latest > /dev/null && \
+		echo "Started docker test central repo with images:" && \
+		$(ROOT_DIR)/hack/central-repo/upload-plugins.sh info
+	
+	@echo "Docker test central repo started at localhost:9876"
+
+.PHONY: start-test-central-repo-11
 start-test-central-repo: stop-test-central-repo setup-custom-cert-for-test-central-repo ## Starts up a test central repository locally with docker
```
marckhouzam, line 315: Careful here.  This target is named the same as the one at line 296 so it will overwrite the above one.
I'm not sure which one you are modifying, but this is the one that runs.

PR #679, opened 2024-02-20
Add new target `operations`/`ops` 

anujc25:

This PR adds a new target `operations`/`ops`

Fixes #

### Release note

```release-note
Add the new target `operations`
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,8 +6,6 @@ replace cloud.google.com/go => cloud.google.com/go v0.102.1
 
 replace github.com/vmware-tanzu/tanzu-cli/test/e2e/framework => ./test/e2e/framework
 
-replace github.com/vmware-tanzu/tanzu-plugin-runtime => github.com/anujc25/tanzu-plugin-runtime v0.0.0-20240214002508-cc83ffc09fda
-
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.6
 	github.com/Masterminds/semver v1.5.0
@@ -44,7 +42,7 @@ require (
 	github.com/vmware-tanzu/carvel-ytt v0.40.0
 	github.com/vmware-tanzu/tanzu-cli/test/e2e/framework v0.0.0-00010101000000-000000000000
 	github.com/vmware-tanzu/tanzu-framework/capabilities/client v0.0.0-20230523145612-1c6fbba34686
-	github.com/vmware-tanzu/tanzu-plugin-runtime v1.2.0-dev.0.20240206192120-055fd83cf63d
+	github.com/vmware-tanzu/tanzu-plugin-runtime v1.2.0-dev.0.20240222161642-59c7320668fa
 	go.pinniped.dev v0.20.0
 	golang.org/x/mod v0.12.0
 	golang.org/x/oauth2 v0.8.0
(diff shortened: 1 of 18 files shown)


pkg/command/root.go
```diff
@@ -297,6 +299,18 @@ var tmcCmd = &cobra.Command{
 	},
 }
 
+var opsCmd = &cobra.Command{
+	Use:     "operations",
+	Short:   "Tanzu CLI Operations related commands",
```
anujc25, line 304: @vuil any suggestion for the help for this target?

go.mod
```diff
@@ -6,6 +6,8 @@ replace cloud.google.com/go => cloud.google.com/go v0.102.1
 
 replace github.com/vmware-tanzu/tanzu-cli/test/e2e/framework => ./test/e2e/framework
 
+replace github.com/vmware-tanzu/tanzu-plugin-runtime => github.com/anujc25/tanzu-plugin-runtime v0.0.0-20240214002508-cc83ffc09fda
```
anujc25, line 9: Update once https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/163 is merged.

marckhouzam:
While I continue reviewing, I thought I'd mention that the shell completion logic was not updated for the new `operations/ops` target.  That change starts with `pkg/command/completion_helper.go`, but needs to be added to every completion function that suggests a target;  you should be able to find them quickly by looking for `compTMCTarget`.

The unit tests will also need to be updated to match... Sorry about that.

vuil (approved):
lgtm, a reminder to update go.mod before merging.

prkalle (approved):
LGTM. Thanks for the changes!

pkg/command/root.go
anujc25: Updated.

marckhouzam (commented):
We need to update:
1. `TargetList` in `pkg/common/constants.go`
2. `TargetList` in `test/e2e/framework/framework_constants.go`

What about documentation?  Should we add something in `docs/quickstart/quickstart.md` and/or `docs/full/README.md`?

When there are not `ops` plugins installed we get
```
$ tz ops                                                                                                                         <gke_dev-adhol>
Note: No plugins are currently installed for the "operations" target.
      Such plugins will be accessible when a "operations" context is created/activated or a "operations" plugin is explicitly installed.
...
```
should we change the note since we now have a target that cannot be a context?  Maybe just for this target?

As a follow-up PR, we may want to consider adding `operations` plugins to the test central repo.

anujc25:
> should we change the note since we now have a target that cannot be a context? Maybe just for this target?
> 
Can we remove this line completely? `Such plugins will be accessible when a "<target>" context is created/activated or a "<target>" plugin is explicitly installed.` 


> As a follow-up PR, we may want to consider adding `operations` plugins to the test central repo.
>
👍

anujc25:
> Such plugins will be accessible when a "<target>" context is created/activated or a "<target>" plugin is explicitly installed.

I have removed `Such plugins will be accessible when a "<target>" context is created/activated or a "<target>" plugin is explicitly installed.` help line for now. If we think we need to add it back, I can add it back as part of the followup PR.

PR #685, opened 2024-02-27
Add boolean validation for tanzu config set feature flags

mpanchajanya:
- Add validation for tanzu config set feature flag value to restrict to boolean true or false values.

Fixes #

- Added unit tests
- Manual testing

```
> tz config get | grep foo
> tz config set features.global.foo bar
[x] : invalid value provided only boolean true or false are accepted
> tz config set features.global.foo true
> tz config set features.global.foo false
> tz config set features.global.foo bar
[x] : invalid value provided only boolean true or false are accepted
> tz config set features.global.foo ""
[x] : invalid value provided only boolean true or false are accepted
> tz config set features.global.foo true
> tz config get | grep foo
            foo: "true"
> 

```

### Release note

```release-note
Tanzu CLI will only allow boolean true or false feature flag values
```

### DIFF
--- a/pkg/command/config.go
+++ b/pkg/command/config.go
@@ -145,6 +145,9 @@ func setConfiguration(pathParam, value string) error {
 		if len(paramArray) != 3 {
 			return errors.New("unable to parse config path parameter into three parts [" + strings.Join(paramArray, ".") + "]  (was expecting 'features.<plugin>.<feature>'")
 		}
+		if value != "true" && value != "false" { //nolint:goconst
+			return errors.New("invalid value provided only boolean true or false are accepted")
+		}
 		return configlib.SetFeature(paramArray[1], paramArray[2], value)
 	case ConfigLiteralEnv:
 		if len(paramArray) != 2 {
--- a/pkg/command/config_test.go
+++ b/pkg/command/config_test.go
@@ -69,6 +69,62 @@ func TestConfigEnv(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


marckhouzam (commented):
For completeness let's mention that this PR is technically not a backwards-compatible change. A script could currently run `tz config set features.cluster.myfeat TRUE` and everything would work but now it will not be able to.  Considering the scenarios, I think a small breaking change is ok but I'd like opinions.
@mpanchajanya @vuil @anujc25 

If we want to reduce the risk of breakage we could allow the widest, valid and consistent values.   Those values must work for both APIs, so they would be `true/false`, `TRUE/FALSE` and `True/False`.  That being said, I think `true/false` is enough but I'd also like opinions.

anujc25:
> For completeness let's mention that this PR is technically not a backwards-compatible change. A script could currently run `tz config set features.cluster.myfeat TRUE` and everything would work but now it will not be able to. Considering the scenarios, I think a small breaking change is ok but I'd like opinions. @mpanchajanya @vuil @anujc25
> 
> If we want to reduce the risk of breakage we could allow the widest, valid and consistent values. Those values must work for both APIs, so they would be `true/false`, `TRUE/FALSE` and `True/False`. That being said, I think `true/false` is enough but I'd also like opinions.

Thanks for the explanation. Yes, I think this should be fine to accept as the field was used to set a boolean so far.

anujc25 (approved):

marckhouzam (approved):

PR #687, opened 2024-03-01
Add plugin-level remapping support

vuil:

```
    Adds experimental functionality to remap in the CLI command tree
    commands at an entire plugin level.

    The mapping information is currently solely derived from the plugin's
    `info` command output it can extend potentially to mapping metadata 
    specified outside of the plugin binaries.

    Mapping to a nonexistent location should only be possible if the parent
    Command exists AND parent is not one associated another plugin's command
    tree root. (iow where the Command's Annotations indicate it is a wrapper for a
    plugin)

    Any command group that exists in the destination location of the mapping
    is removed from the command tree.

    Mapping is contingent on whether an active CLI context is of a type
    that is one of the supportedContextType list, if the latter is explicitly
    expressed by the plugin.

    An upcoming change will provide more contextual information to the
    plugin on how it's commands are invoked by the Tanzu CLI. Until then
    this remapping functionality is best limited to scenarios of relocating
    without command group renaming, or ones that only involve minor and
    non-surprising renames of the command group for the plugin's commands.
```

Fixes #

Update unit tests. Additionally

Assuming k8s-targetted plugin `apps` is installed

Build and install additional plugin with different PluginDescriptor ...
1)
```
        p, err := plugin.NewPlugin(&plugin.PluginDescriptor{
                Name:                 "appsv2",
                Description:          "Applicationz2 on Kubernetes",  << for ease of disambuigate of top level help
                SupportedContextType: []string{string(types.ContextTypeTanzu)},
                InvokedAs:            []string{"apps"},
...
        })
```
verified that the apps cmd node (invoked via tanzu apps ... ) is replaced with that for the new plugin. Similarly for `tanzu kubernetes app`, only when an active CLI context is of type tanzu

```
> tanzu

Usage:
  tanzu [command]

Available command groups:

  Admin
  Build
    apps                    Applicationz2 on Kubernetes

----

> tanzu kubernetes

Usage:
  tanzu kubernetes [command]

Aliases:
  kubernetes, k8s

Available command groups:

  Build
    apps                    Applicationz2 on Kubernetes
```

2)
```
        p, err := plugin.NewPlugin(&plugin.PluginDescriptor{
                Name:                 "appsv2",
                Description:          "Applicationz2 on Kubernetes",
                InvokedAs:            []string{"apps"},
...
        })
```
verified that the apps cmd node (invoked via tanzu apps ... ) is unconditionally replaced with that for the new plugin. Similarly for `tanzu kubernetes app`

3)
```
        p, err := plugin.NewPlugin(&plugin.PluginDescriptor{
                Name:                 "appsv2",
                Description:          "Applicationz2 on Kubernetes",
                InvokedAs:            []string{"appsalt"},
...
        })
```
verified that the appsalt cmd node (invoked via tanzu appsalt ... ) is introduced. 
This is not a recommended scenario as the presence of the original apps command group could lead to UX artifacts like duplicated aliases warning.

### Release note

```release-note
    Adds experimental functionality to remap in the CLI commands at an entire plugin level.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,7 +42,7 @@ require (
 	github.com/vmware-tanzu/carvel-ytt v0.40.0
 	github.com/vmware-tanzu/tanzu-cli/test/e2e/framework v0.0.0-00010101000000-000000000000
 	github.com/vmware-tanzu/tanzu-framework/capabilities/client v0.0.0-20230523145612-1c6fbba34686
-	github.com/vmware-tanzu/tanzu-plugin-runtime v1.2.0-dev.0.20240222161642-59c7320668fa
+	github.com/vmware-tanzu/tanzu-plugin-runtime v1.3.0-dev.0.20240302001423-9b6a526ce007
 	go.pinniped.dev v0.20.0
 	golang.org/x/mod v0.12.0
 	golang.org/x/oauth2 v0.8.0
--- a/pkg/cli/plugin_cmd.go
+++ b/pkg/cli/plugin_cmd.go
@@ -22,9 +22,9 @@ import (
 // the kubernetes group. Any remap location that points to one should also map
 // to the other to be consistent.
 // This function returns the alternate location, if applicable.
-func alternateRemapLocation(p *PluginInfo, location string) string {
+func alternateRemapLocation(p *PluginInfo, remapLocation string) string {
 	if p.Target == configtypes.TargetK8s {
-		cmdHierarchy := strings.Split(strings.TrimSpace(location), " ")
+		cmdHierarchy := strings.Split(remapLocation, " ")
 		if len(cmdHierarchy) == 1 {
 			return fmt.Sprintf("kubernetes %s", cmdHierarchy[0])
 		}
@@ -39,8 +39,9 @@ func alternateRemapLocation(p *PluginInfo, location string) string {
 func GetCommandMapForPlugin(p *PluginInfo) map[string]*cobra.Command {
 	cmdMap := map[string]*cobra.Command{}
 
-	for _, remapLocation := range p.InvokedAs {
-		cmdHierarchy := strings.Split(strings.TrimSpace(remapLocation), " ")
+	for _, invokedAsItem := range p.InvokedAs {
+		remapLocation := strings.TrimSpace(invokedAsItem)
+		cmdHierarchy := strings.Split(remapLocation, " ")
 
 		if len(cmdHierarchy) > 0 {
 			cmdMap[remapLocation] = getCmdForPluginEx(p, cmdHierarchy[len(cmdHierarchy)-1])
--- a/pkg/cli/plugin_cmd.go
+++ b/pkg/cli/plugin_cmd.go
@@ -5,20 +5,73 @@ package cli
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
 
 	"github.com/spf13/cobra"
 
-	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
+	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
 )
+
+// K8s-targeted plugins command group shows up as top level commands and under
+// the kubernetes group. Any remap location that points to one should also map
+// to the other to be consistent.
+// This function returns the alternate location, if applicable.
+func alternateRemapLocation(p *PluginInfo, location string) string {
+	if p.Target == configtypes.TargetK8s {
+		cmdHierarchy := strings.Split(strings.TrimSpace(location), " ")
+		if len(cmdHierarchy) == 1 {
+			return fmt.Sprintf("kubernetes %s", cmdHierarchy[0])
+		}
+		if len(cmdHierarchy) == 2 && cmdHierarchy[0] == "kubernetes" {
+			return cmdHierarchy[1]
+		}
+	}
+	return ""
+}
+
+// GetCommandMapForPlugin returns how the plugin's commands should be mapped
+func GetCommandMapForPlugin(p *PluginInfo) map[string]*cobra.Command {
+	cmdMap := map[string]*cobra.Command{}
+
+	for _, remapLocation := range p.InvokedAs {
+		cmdHierarchy := strings.Split(strings.TrimSpace(remapLocation), " ")
+
+		if len(cmdHierarchy) > 0 {
+			cmdMap[remapLocation] = getCmdForPluginEx(p, cmdHierarchy[len(cmdHierarchy)-1])
+		}
+		if alternateLocation := alternateRemapLocation(p, remapLocation); alternateLocation != "" {
+			cmdMap[alternateLocation] = cmdMap[remapLocation]
+		}
+	}
+
+	return cmdMap
+}
 
 // GetCmdForPlugin returns a cobra command for the plugin.
 func GetCmdForPlugin(p *PluginInfo) *cobra.Command {
+	return getCmdForPluginEx(p, p.Name)
+}
+
+// GetUnmappedCmdForPlugin returns a cobra command for the plugin unless there
+// are remapping directives in the plugin info, in which case it will return
+// nil instead.
+func GetUnmappedCmdForPlugin(p *PluginInfo) *cobra.Command {
+	if len(p.InvokedAs) > 0 {
+		return nil
+	}
+	return getCmdForPluginEx(p, p.Name)
+}
+
+// getCmdForPluginEx returns a cobra command for the plugin.
+func getCmdForPluginEx(p *PluginInfo, cmdGroupName string) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   p.Name,
+		Use:   cmdGroupName,
 		Short: p.Description,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			runner := NewRunner(p.Name, p.InstallationPath, args)
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -108,6 +108,7 @@ func NewRootCmd() (*cobra.Command, error) { //nolint: gocyclo,funlen
 				// We therefore replace the existing command with the new command by removing the old and
 				// adding the new one.
 				maskedPluginsWithPluginOverlap = append(maskedPluginsWithPluginOverlap, matchedCmd.Name())
+
 				rootCmd.RemoveCommand(matchedCmd)
 				rootCmd.AddCommand(cmd)
 			} else if plugins[i].Name != "login" {
@@ -124,6 +125,7 @@ func NewRootCmd() (*cobra.Command, error) { //nolint: gocyclo,funlen
 	}
 
 	remapCommandTree(rootCmd, plugins)
+	updateTargetCommandGroupVisibility()
 
 	if len(maskedPluginsWithPluginOverlap) > 0 {
 		catalog.DeleteIncorrectPluginEntriesFromCatalog()
@@ -144,7 +146,7 @@ func remapCommandTree(rootCmd *cobra.Command, plugins []cli.PluginInfo) {
 	for pathKey, cmd := range cmdMap {
 		matchedCmd, parentCmd := findSubCommandByPath(rootCmd, pathKey)
 
-		if parentCmd != nil && parentCmd.Annotations != nil && parentCmd.Annotations["type"] == common.CommandTypePlugin {
+		if parentCmd != nil && isPluginCommand(parentCmd) {
 			fmt.Fprintf(os.Stderr, "Remap of plugin into command tree (%s) associated with another plugin is not supported\n", parentCmd.Name())
 			continue
 		}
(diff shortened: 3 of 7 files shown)


go.mod
```diff
@@ -6,6 +6,8 @@ replace cloud.google.com/go => cloud.google.com/go v0.102.1
 
 replace github.com/vmware-tanzu/tanzu-cli/test/e2e/framework => ./test/e2e/framework
 
+replace github.com/vmware-tanzu/tanzu-plugin-runtime => github.com/vuil/tanzu-plugin-runtime a454868
```
vuil, line 9: to be updated once runtime side changes are merged

marckhouzam: I wasn't aware you could fix a sha when doing a replace.  Nice!

marckhouzam (commented):
Things seems to work nicely 👏 Very cool to see!

Some comments to clarify my understanding

pkg/cli/plugin_cmd.go
```diff
@@ -5,20 +5,73 @@ package cli
...
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
 )
 
+// K8s-targeted plugins command group shows up as top level commands and under
+// the kubernetes group. Any remap location that points to one should also map
+// to the other to be consistent.
+// This function returns the alternate location, if applicable.
+func alternateRemapLocation(p *PluginInfo, location string) string {
+	if p.Target == configtypes.TargetK8s {
```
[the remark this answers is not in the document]
vuil: The InvokedAs string expectation is intentional, and is meant to be decouplable from the target concept eventually. The target is not used in the remapping algorithm, other than in this specific function, which is meant to reproduce the default behavior making kubernetes-targetted plugin commands appear under kubernetes and under toplevel. So space-delimited values can convey that we can support a more flexible placement of the commands should we need to reorganize the CLI command tree.

pkg/cli/plugin_cmd.go
```diff
@@ -5,20 +5,73 @@ package cli
...
+		cmdHierarchy := strings.Split(strings.TrimSpace(location), " ")
+		if len(cmdHierarchy) == 1 {
+			return fmt.Sprintf("kubernetes %s", cmdHierarchy[0])
+		}
+		if len(cmdHierarchy) == 2 && cmdHierarchy[0] == "kubernetes" {
+			return strings.Join(cmdHierarchy[1:], " ")
+		}
```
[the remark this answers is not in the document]
vuil: Thanks for bring this up. As an experimental feature, the intention is to make available potentially useful mapping scenarios which includes:

- mapping a plugin to entirely replace that occupied by another
- placing the plugin's command tree under some other non-default "grouping" cmd 

But you noted in the plugin-runtime comment a case that I did not intend to support with this change, which is the ability to embed a plugin's command tree within that of another plugin's command tree. (think "tanzu space _builder cli add-plugin_", as an nonsensical example)
 
It is not because the mapping does not work, but it is more due to the "upper" plugin having no clue that this mapping has occurred (it's Usage will be incorrect) and that there are no known use cases that requires this.

So, I added as a topmost commit additional checks to prevent the remap if the parent Command is that which wraps a plugin.

vuil (commented):
Thanks @marckhouzam for the great comments and questions. 
I added a change to further constrain the remap algorithm to be more selective of the parent Command it can use to attach the plugin's command tree

pkg/pluginsupplier/plugin_supplier.go
```diff
@@ -48,6 +49,31 @@ func GetInstalledServerPlugins() ([]cli.PluginInfo, error) {
 	return serverPlugins, nil
 }
 
+// FilterPluginsByActiveContextType will exclude any plugin with an explicit
+// setting of supportedContextType that does not match the type of any active CLI context
+// Separating this conditional check so GetInstalledStandalonePlugins can
+// continue to return all standalone plugins regardless of supportedContextType
+func FilterPluginsByActiveContextType(plugins []cli.PluginInfo) (result []cli.PluginInfo, err error) {
```
vuil, line 56: I didn't want to the potential filtering via active contexts factor into existing plugin-listing functions in this package, hence introducing it as another function. Curious if others have suggestions to alternatives? 
cc @marckhouzam @anujc25

pkg/cli/plugin_cmd.go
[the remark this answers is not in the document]
marckhouzam: To are correct and I hadn’t thought through the implications of adding a plugin under another one.  Such a deeper remapping would only work (I think) under a core command’s command tree since the cli is aware of all of those. A plugin could for example add `tanzu context createtmc`. I’m only mentioning it as a possible future potential but I like that you blocked all that for now

pkg/pluginsupplier/plugin_supplier.go
anujc25: I think this seems a reasonable thing to start with. We can enhance and update this once we finalize the USX around listing plugins with the `tanzu plugin list` command.

pkg/command/root_test.go
```diff
@@ -87,6 +87,80 @@ func (s *TestPluginSupplier) GetInstalledPlugins() ([]*cli.PluginInfo, error) {
...
+			"TANZU_CONFIG",
+			"TANZU_CONFIG_NEXT_GEN",
+			"TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER",
+			"TANZU_CLI_EULA_PROMPT_ANSWER",
+		},
+	}
+}
+
+// helper to clean up CLI environment
+func tearDownTestCLIEnvironment(env testCLIEnvironment) {
```
marckhouzam, line 154: Nice consolidation.

pkg/command/root_test.go
```diff
@@ -823,13 +825,6 @@ func TestEnvVarsSet(t *testing.T) {
 	assert.Nil(err)
 	// Make sure the variable is now set during the call to the CLI
 	assert.Equal(envVarValue, os.Getenv(envVarName))
-
-	// Cleanup
-	os.Unsetenv("TANZU_CONFIG")
-	os.Unsetenv("TANZU_CONFIG_NEXT_GEN")
-	os.Unsetenv("TANZU_CLI_CEIP_OPT_IN_PROMPT_ANSWER")
-	os.Unsetenv("TANZU_CLI_EULA_PROMPT_ANSWER")
-	os.Unsetenv(envVarName)
```
marckhouzam, line 832: Nit: the last Unsetenv should be kept (although it surely has no impact to remove it)

pkg/command/root_test.go
```diff
@@ -908,3 +903,311 @@ func TestCompletionShortHelpInActiveHelp(t *testing.T) {
...
+			stdout := os.Stdout
+			stderr := os.Stderr
+			defer func() {
+				os.Stdout = stdout
+				os.Stderr = stderr
+			}()
+			os.Stdout = w
+			os.Stderr = w
+
+			assert.Nil(err)
```
marckhouzam, line 1175: Stray assert?

pkg/command/root_test.go
```diff
@@ -908,3 +903,311 @@ func TestCompletionShortHelpInActiveHelp(t *testing.T) {
...
+		args              []string
+		expected          []string
+		unexpected        []string
+		expectedFailure   bool
+		activeContextType configtypes.ContextType
+	}{
+		{
+			test: "one unmapped k8s plugin",
+			pluginVariants: []fakePluginRemapAttributes{
+				fakePluginRemapAttributes{
```
marckhouzam, line 966: The `fakePluginRemapAttributes` part of this line and the similar ones below give me a warning that it is a redundant declaration.  It seems that since the array type is specified, the element type need not be.

pkg/command/root_test.go
```diff
@@ -908,3 +903,311 @@ func TestCompletionShortHelpInActiveHelp(t *testing.T) {
...
+					name:      "dummy2",
+					target:    configtypes.TargetK8s,
+					invokedAs: []string{"dummy"},
+					aliases:   []string{"dum"},
+				},
+			},
+			args:       []string{"dummy", "say", "hello"},
+			expected:   []string{"hello"},
+			unexpected: []string{"duplciated across plugins"},
```
marckhouzam, line 1033: duplciated -> duplicated

pkg/command/root_test.go
```diff
@@ -908,3 +903,311 @@ func TestCompletionShortHelpInActiveHelp(t *testing.T) {
...
+				},
+				fakePluginRemapAttributes{
+					name:      "dummy3",
+					target:    configtypes.TargetK8s,
+					invokedAs: []string{"dummy"},
+					aliases:   []string{"dum"},
+				},
+			},
+			args:     []string{"dummy", "say", "hello"},
+			expected: []string{"hello", "Warning, multiple command groups are being remapped to the same command names : \"dummy, dummy\""},
```
marckhouzam, line 1088: Showing "dummy" twice seems not very useful.  In the code (not the test) should we instead print the plugin name so we can figure out which plugins are problematic?

pkg/command/root_test.go
```diff
@@ -908,3 +903,311 @@ func TestCompletionShortHelpInActiveHelp(t *testing.T) {
...
+			expected:   []string{"hello"},
+			unexpected: []string{"duplciated across plugins"},
+		},
+		{
+			test: "map to deeper subcommand is valid as long as parent exists",
+			pluginVariants: []fakePluginRemapAttributes{
+				fakePluginRemapAttributes{
+					name:      "dummy2",
+					target:    configtypes.TargetK8s,
+					invokedAs: []string{"dummy", "kubernetes dummy"},
```
marckhouzam, line 1041: Could we add a test that only specifies `dummy` in `invokedAs` but that checks that we can still call it as `[]string{"kubernetes", "dummy", "say", "hello"}`?

And the reverse as another test (specify only `kubernetes dummy` but still call as  `[]string{"dummy", "say", "hello"}`?

marckhouzam (commented):
Really excellent unit tests!  Gives strong confidence in the feature.
A couple of suggestions for the test code.

marckhouzam:
I assume that the explanation:
"An upcoming change will provide more contextual information to the plugin on how it's commands are invoked by the Tanzu CLI. Until then this remapping functionality is best limited to scenarios of relocating without command group renaming, or ones that only involve minor and non-surprising renames of the command group for the plugin's commands."

means that we have to accept that the help text shows the plugin name and not the `invokeAs` name?

pkg/command/root.go
```diff
@@ -144,8 +201,11 @@ func setupTargetPlugins() error {
 
 	// Insert the plugin commands under the appropriate target command
 	for i := range installedPlugins {
-		if cmd, exists := mapTargetToCmd[installedPlugins[i].Target]; exists {
-			cmd.AddCommand(cli.GetCmdForPlugin(&installedPlugins[i]))
+		if targetCmd, exists := mapTargetToCmd[installedPlugins[i].Target]; exists {
+			cmd := cli.GetUnmappedCmdForPlugin(&installedPlugins[i])
+			if cmd != nil {
+				targetCmd.AddCommand(cmd)
+			}
```
marckhouzam, line 208: Right below this line we look if a target should be hidden because it has no plugins.
This logic needs to be moved until the mapped plugins have been installed.

You can try it now by removing all plugins and then installing a single mapped plugin under the kubernetes target.  You will notice that the `kubernetes` target is hidden from the help and shell completion, even though the mapped plugin is invokable under that target.

marckhouzam: Also, this function uses `pluginsupplier.GetInstalledPlugins()` and therefore does not respect the new `SupportedContextType` field.  This means that if a plugin specifies `SupportedContextType: "tanzu"` without an `InvokedAs` remap, the CLI will remove that plugin if there is no "tanzu" context active, BUT it will not remove the plugin from under targets.

marckhouzam (approved):
I'm pretty much finished.
I like this a lot, nice job.
I let you decide what should be fixed for alpha.1 and what can be fixed after it.

pkg/command/root_test.go
vuil: good idea, thanks!

vuil:
> I assume that the explanation: "An upcoming change will provide more contextual information to the plugin on how it's commands are invoked by the Tanzu CLI. Until then this remapping functionality is best limited to scenarios of relocating without command group renaming, or ones that only involve minor and non-surprising renames of the command group for the plugin's commands."
> 
> means that we have to accept that the help text shows the plugin name and not the `invokeAs` name?

It is a rough edge, but I am inclined keep this and address improvement as a followup. Partly because changes will likely involve additiona runtime changes, and when made should hopefully be thorough enough for plugin code (not just the help text) can also account for the mapping.

pkg/command/root_test.go
vuil: I have already added a //Todo in the code that generated the warning to  provide a better message in a followup. Just need to think through what best to show. (Note: the "dummy" twice is not reporting that two plugins is mapping to the same dummy command node, but more that two nodes name "dummy" ("dummy" and "kubernetes dummy") are both multiply mapped.

pkg/command/root.go
vuil: Nice catch! Updated to filter plugins via ActiveContext as well, and added unit test case

pkg/command/root_test.go
marckhouzam: Turns out this stray assert was actually a copy/paste from line 755 which has the same "problem" and that git blames me for 😆 

Would you mind also removing the one at line 755 since you have your hands in this file?

marckhouzam (approved):
Some final nits in the tests but LGTM

marckhouzam (approved):

PR #688, opened 2024-03-06
Fix shell interpolation issues in get-new-contributors step

vuil:
For more info see:
[discussion on SO](https://stackoverflow.com/questions/75024695/github-actions\-how-to-handle-a-in-json-when-passing-it-across-steps)

Fixes #

### Release note

```release-note

```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -116,9 +116,11 @@ jobs:
 
     - name: 'Identify New Contributors Section'
       id: get-new-contributors
+      env:
+        body: "${{ fromJson(steps.get-github-release-notes.outputs.data).body }}"
       run: |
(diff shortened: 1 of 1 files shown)


anujc25 (approved):

.github/workflows/release.yaml
```diff
@@ -116,9 +116,11 @@ jobs:
 
     - name: 'Identify New Contributors Section'
       id: get-new-contributors
+      env:
+        body: "${{ fromJson(steps.get-github-release-notes.outputs.data).body }}"
```
marckhouzam, line 120: Does this set the “$body” variable? Because below we use a new variable called “$data” instead

vuil: Ouch. Thanks for catching. Updated.

marckhouzam (approved):
LGTM

PR #691, opened 2024-03-07
Add global --verbose / -v flag

mpanchajanya:

Fixes #

### Release note

```release-note

```


pkg/command/root.go
```diff
@@ -291,13 +331,22 @@ func newRootCmd() *cobra.Command {
...
-		if err == nil {
-			log.SetVerbosity(int32(logValue))
+func setLoggerVerbosity(verbosity int) {
+	// If verbose global flag is passed set the log level verbosity  if not then check if env variable TANZU_CLI_LOG_LEVEL is set
+	if verbosity > 0 {
+		// Set the log level verbosity with the verbosity value
+		log.SetVerbosity(int32(verbosity))
+
+		// Set the TANZU_CLI_LOG_LEVEL env with the verbosity value
+		_ = os.Setenv(log.EnvTanzuCLILogLevel, strconv.Itoa(verbosity))
```
marckhouzam, line 341: For plugins to consume transparently. Nice!  

In the unit test you write, can you check that this variable is set as expected?

pkg/command/root.go
```diff
@@ -291,13 +331,22 @@ func newRootCmd() *cobra.Command {
...
-func setLoggerVerbosity() {
-	// Configure the log level if env variable TANZU_CLI_LOG_LEVEL is set
-	logLevel := os.Getenv(log.EnvTanzuCLILogLevel)
-	if logLevel != "" {
-		logValue, err := strconv.ParseInt(logLevel, 10, 32)
-		if err == nil {
-			log.SetVerbosity(int32(logValue))
+func setLoggerVerbosity(verbosity int) {
+	// If verbose global flag is passed set the log level verbosity  if not then check if env variable TANZU_CLI_LOG_LEVEL is set
+	if verbosity > 0 {
```
marckhouzam, line 336: I think this will always be true, since the default value for the flag is > 0.
Notice how the variable is still ignored:
```
$ TANZU_CLI_LOG_LEVEL=3 tz version
[i] I am level default
[i] I am level 1
version: v1.3.0-dev
buildDate: 2024-03-12
sha: f61583163-dirty
arch: arm64
```
Can you write a test for this?

Fixing this is a bit tricky.
If the user sets the `--verbose` flag then the flag value should be respected (ignoring `TANZU_CLI_LOG_LEVEL`); if the flag is not set, then `TANZU_CLI_LOG_LEVEL` should be honoured.
To know if the flag was set on the command-line we can:
1. set a default value that a user would not use like -1.  The problem is that the help text of the flag will show that default value.  That's not good.
2. don't set a default value (technically) by setting it to 0.  Cobra will not put it in the help text in that case.  Then if we see a value of 0, we assume the flag was not used.  BUT, do we want to allow the user to set `--verbose 0`?  It could make sense if they want to disable all logs maybe?  or else they cannot turn off level 1 logs.  So that approach is also not good.
3. ask Cobra if the flag was actually found on the command line.  We do this in `completeDeletePlugin()`. You can do it with `if !cmd.Flags().Changed("verbose") {verbose = -1}`

pkg/command/root.go
```diff
@@ -240,8 +247,41 @@ func newRootCmd() *cobra.Command {
...
+						// Skip the next argument
+						i++
+					} else {
+						log.Errorf("Missing value for verbose flag")
+					}
+				}
+			}
+
+			// Sets the verbosity of the logger
+			setLoggerVerbosity(verbose)
```
marckhouzam, line 275: Should we fail execution if the user sets an invalid value for `--verbose`?
Notice the difference:
```
$ tz version --verbose allo
[x] : invalid argument "allo" for "--verbose" flag: strconv.ParseInt: parsing "allo": invalid syntax

$ tz version --verbose -10
[i] I am level default
[i] I am level 1
[i] I am level 2
[i] I am level 3
...
```

pkg/command/root.go
```diff
@@ -240,8 +247,41 @@ func newRootCmd() *cobra.Command {
 		// silencing usage for now as we are getting double usage from plugins on errors
 		SilenceUsage: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			// Sets the verbosity of the logger if TANZU_CLI_LOG_LEVEL is set
-			setLoggerVerbosity()
+			// Manually parse the flags
+			for i := 0; i < len(args); i++ {
+				arg := args[i]
+				if arg == "-v" || arg == "--verbose" {
```
marckhouzam, line 253: Cobra automatically verifies that there is a value set for the flag.
In fact, I don't believe this if statement will ever be true because by the time `PersistentPreRunE` is called, all flags have been parsed and removed from the list of args.

You can remove this whole loop

pkg/command/root.go
```diff
@@ -240,8 +247,41 @@ func newRootCmd() *cobra.Command {
...
+						}
+						// Skip the next argument
+						i++
+					} else {
+						log.Errorf("Missing value for verbose flag")
+					}
+				}
+			}
+
+			// Sets the verbosity of the logger
```
marckhouzam, line 274: We still need to check if a plugin has also declared a `--verbose` flag, and in that case pass it to the plugin.
I now realize this is a little complicated because if the user does:
```
tz telemetry --verbose true
```
the command will fail because Cobra will try to parse `true` as an integer.

We can do three things:
1. make the `--verbose` flag a string so that it never fails to parse and handle the value checking ourselves.  But that won't work if the plugin set the flag as a boolean since it does not take a value while a string flag does.  We may be able to get around that by setting the flags `NoOptDefVal` field.  I haven't tried.
2. _Before_ calling `Execute()` which means before Cobra parses the flags, manually parse `os.Args` to find `--verbose` and remove it from the args if the plugin has declared it.  That means also figuring out which plugin is being called.  This gets pretty tricky but can be done.  I would prefer to avoid it though.
3. Discuss if we can work with plugin authors to figure out if we really need to worry about plugins having declared `--verbose` as something else than `int` or `string`.

marckhouzam (commented):
Thanks for this.  This may seem like a small change, but it is quite tricky.

Nit: can you fix the commit message?  One title line, then an empty line, then a description.

I noticed that the [style guide](docs/plugindev/style_guide.md) recommends plugins support a `--verbose` flag.  Could you update that section of the doc?

pkg/command/root.go
```diff
@@ -33,6 +33,9 @@ import (
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/plugin"
 )
 
+// verbose flag to set the log lovel
```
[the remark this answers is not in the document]
mpanchajanya: Updated

pkg/command/root.go
```diff
@@ -42,6 +45,9 @@ func NewRootCmd() (*cobra.Command, error) { //nolint: gocyclo,funlen
 	// Configure defined environment variables found in the config file
 	cliconfig.ConfigureEnvVariables()
 
+	// Initialize the global verbose flag to set the log level verbosity (acceptable values 1 - 9) (default value 1)
+	rootCmd.PersistentFlags().IntVar(&verbose, "verbose", 1, "number for the log level verbosity (default value 1) (acceptable values 1 - 9)")
```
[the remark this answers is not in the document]
mpanchajanya: Removed additional default help text

pkg/command/root.go
mpanchajanya: So I have added this loop to parse args when plugin binaries are triggered.
1. For CLI commands; Yes args will be empty and this loop will not run
2. For Plugin commands; Since CLI doesn't parse args and sends all args to plugin this loop will run to verify if verbose flag is set by the user and then passes to plugin.

pkg/command/root.go
mpanchajanya: Added validation to verbose flag
`> tz version --verbose -10
Error: Invalid value for verbose flag. It should be equal or greater than 0
> tz telemetry --verbose -10
[x] Missing value for verbose flag
`

pkg/command/root.go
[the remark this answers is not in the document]
mpanchajanya: Updated the default value to 0

pkg/command/root.go
mpanchajanya: Updated the if statement to `if verbosity >= 0 ` since we set the verbose to -1 if user doesn't set the flag

pkg/command/root.go
mpanchajanya: For Plugin commands - CLI doesnt parse the args to its flags but sends all the args to plugin as arguments. we only verify if verbose flag is passed by user and then set the verbosity accordingly.  
Even if verbose flag is bool/string for plugins, CLI would pass those args.  As part of verbosity we need to determine what should we set the verbosity if verbose is true or false.

pkg/command/root.go
```diff
@@ -240,8 +247,74 @@ func newRootCmd() *cobra.Command {
 		// silencing usage for now as we are getting double usage from plugins on errors
 		SilenceUsage: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			// Sets the verbosity of the logger if TANZU_CLI_LOG_LEVEL is set
-			setLoggerVerbosity()
+
+			fmt.Println(args, "args", verbose)
+
+			// Validate the verbose flag
+			if verbose < 0 {
```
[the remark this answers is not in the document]
mpanchajanya: Added the check

pkg/command/root.go
```diff
@@ -240,8 +247,74 @@ func newRootCmd() *cobra.Command {
...
+			if !cmd.Flags().Changed("verbose") {
+				verbose = -1
+			}
+
+			/**
+			Manually parse the flags when plugin command is triggered
+			1. For CLI commands; all flags have been parsed and removed from the list of args and this loop will not run
+			2. For Plugin commands; Since CLI doesn't parse args and sends all args to plugin this loop will run to verify if verbose flag is set by the user and then passes to plugin.
+			*/
+			for i := 0; i < len(args); i++ {
```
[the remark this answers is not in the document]
mpanchajanya: Yes, moved to a sep func

pkg/command/root.go
```diff
@@ -240,8 +247,74 @@ func newRootCmd() *cobra.Command {
...
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			// Sets the verbosity of the logger if TANZU_CLI_LOG_LEVEL is set
-			setLoggerVerbosity()
+
+			fmt.Println(args, "args", verbose)
+
+			// Validate the verbose flag
+			if verbose < 0 {
+				log.Errorf("Invalid value for verbose flag. It should be 0 or greater")
+				os.Exit(1)
```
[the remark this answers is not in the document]
mpanchajanya: updated to return error

pkg/command/root.go
```diff
@@ -42,6 +45,9 @@ func NewRootCmd() (*cobra.Command, error) { //nolint: gocyclo,funlen
 	// Configure defined environment variables found in the config file
 	cliconfig.ConfigureEnvVariables()
 
+	// Initialize the global verbose flag to set the log level verbosity (acceptable values 1 - 9)
+	rootCmd.PersistentFlags().IntVar(&verbose, "verbose", 0, "number for the log level verbosity(acceptable values 1 - 9)")
```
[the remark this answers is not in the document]
mpanchajanya: updated help text

pkg/command/root.go
mpanchajanya: Thanks added the check

docs/plugindev/style_guide.md
```diff
@@ -208,9 +208,6 @@ Available input components:
 
 ### Feedback principles
 
-Useful for everyone, critical to the usability of screen-readers and automation
-Support verbosity flags (`-v`, `--verbose`).
-
```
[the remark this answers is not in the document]
mpanchajanya: Updated the styleguide

pkg/command/root.go
```diff
@@ -240,8 +247,74 @@ func newRootCmd() *cobra.Command {
...
+			}
+
+			/**
+			Manually parse the flags when plugin command is triggered
+			1. For CLI commands; all flags have been parsed and removed from the list of args and this loop will not run
+			2. For Plugin commands; Since CLI doesn't parse args and sends all args to plugin this loop will run to verify if verbose flag is set by the user and then passes to plugin.
+			*/
+			for i := 0; i < len(args); i++ {
+				arg := args[i]
+				if arg == "--verbose" {
```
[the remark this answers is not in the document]
mpanchajanya: Thanks @marckhouzam  Added logic to verify the plugin flags and filter out verbose flag if plugin doesnt support

pkg/command/root.go
```diff
@@ -290,14 +329,49 @@ func newRootCmd() *cobra.Command {
...
+				nextArg := args[i+1]
+				// Check if the next argument is another flag
+				if strings.HasPrefix(nextArg, "-") {
+					log.Errorf("Missing value for verbose flag")
+					os.Exit(1)
+				}
+				// Try to convert the next argument to an integer
+				// If it's not an integer, CLI doesn't parse the verbose flag
+				if v, err := strconv.Atoi(nextArg); err == nil {
+					verbose = v
```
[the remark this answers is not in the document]
mpanchajanya: Current Behaviour

Use Case 1 ->  **Plugin with --verbose int**  - Tanzu CLI only processes the --verbose flag of type int and then passes it to plugin if it supports --verbose flag.

Use Case 2 -> **Plugin with --verbose bool/string** - Tanzu CLI will not process the flag and passes the flag arg directly to plugin

pkg/command/root.go
```diff
@@ -290,14 +331,59 @@ func newRootCmd() *cobra.Command {
 	return rootCmd
 }
 
+// determineVerbosityFromPluginCommandArgs processes the command line arguments for plugin commands.
+// It specifically looks for the '--verbose' flag and validates its value.
+// The function returns an error if the verbose flag is missing its value or the value is not an integer between 0 and 9.
+func determineVerbosityFromPluginCommandArgs(args []string) error {
```
marckhouzam, line 337: This logic is similar to what we are doing in `plugin_cmd.go`.
I wonder if we could consolidate it in `plugin_cmd.go` and call Cobra's `ParseFlags` on the `--verbose` flag args like helm does.

That way Cobra would fill the `verbose` variable instead of us having to do it here.
And we'll be prepared for any other future global flag we may add later on.

pkg/command/discovery_source_test.go
```diff
@@ -145,6 +145,8 @@ func Test_createAndListDiscoverySources(t *testing.T) {
 	testSource2 := "localhost:9876/tanzu-cli/plugins/sandbox1:small"
 	os.Setenv(constants.ConfigVariableAdditionalDiscoveryForTesting, testSource1+","+testSource2)
 
+	rootCmd, err = NewRootCmd()
+	assert.Nil(err)
```
marckhouzam, line 149: What broke that we need this?

marckhouzam (changes requested):
we should not merge this until we figure out what to do for existing plugins that may already have such a flag

PR #705, opened 2024-03-15
fix: Do not prompt for user confirmation to delete context if not exists

warango4:
When user wants to delete a non-existing context, they would first be prompted about deleting the context and then they'd see the error pointing the context does not exist. 

This MR intends to improve the user experience and not be prompted to any survey if the context does not exist in the firs place. 

N/A

- Updated unit testing
- Built and executed locally and tried to delete non-existing context.

image


prkalle (approved):
LGTM. Thanks!

anujc25 (approved):
LGTM.

PR #700, opened 2024-03-13
Update test central repo with some test Central Config

marckhouzam:
### What this PR does / why we need it

This PR starts things off for the implementation of the Central Configuration feature of #699.

The first commit of the PR updates the test central repository images to contain a new `central_config.yaml` file.
This will make it simpler to work in the implementation of the Central Configuration feature.

The second commit updates the CLI to include the new `central_config.yaml` when setting up the plugin source cache.

With this PR, whenever the CLI uses any of the test central repo images a new file `central_config.yaml` will be part of the `$HOME/.cache/tanzu/plugin_inventory/default` directory.

### Which issue(s) this PR fixes

Part of #699

### Describe testing done for PR

Make sure things still work with the real central repo.
```
$ tz plugin source init
[i] Refreshing plugin inventory cache for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.
[i] Reading plugin inventory for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.
[ok] successfully initialized discovery source
$ ls -l ~/.cache/tanzu/plugin_inventory/default
total 1440
-rw-r--r--  1 kmarc  staff      70 13 Mar 11:03 digest.b6c4a21c4da1764327406642b1bf2591638ae05a020305bd1395de2f323a9ab9
-rw-r--r--  1 kmarc  staff       0 13 Mar 11:03 metadata.digest.none
-rw-------  1 kmarc  staff  733184 13 Mar 11:03 plugin_inventory.db
$ tz plugin search --name builder
[i] The tanzu cli essential plugins have not been installed and are being installed now. The install may take a few seconds.
[i] Installing plugins from plugin group 'vmware-tanzucli/essentials:v1.0.0'
[i] Installed plugin 'telemetry:v1.1.0' with target 'global' (from cache)

  NAME     DESCRIPTION             TARGET  LATEST
  builder  Build Tanzu components  global  v1.2.0
```

Now try with the new test central:small image
```
$ tz config set env.TANZU_CLI_PLUGIN_DISCOVERY_IMAGE_SIGNATURE_VERIFICATION_SKIP_LIST localhost:9876/tanzu-cli/plugins/central:small

$ tz plugin source update default -u localhost:9876/tanzu-cli/plugins/central:small
[i] Refreshing plugin inventory cache for "localhost:9876/tanzu-cli/plugins/central:small", this will take a few seconds.
[...]

PR #707, opened 2024-03-18
Notify user of a new recommended CLI version

### LINKED ISSUE #706: CLI detects when a new version is available

Today, users have to go out-of-band to discover if/when there's a new version of the Tanzu CLI available and they may go for months without upgrading.

We'd like to encourage users to upgrade regularly so we can mitigate risks associated with users running older versions (which don't have the latest features e.g. `v1.1.0`'s support for setting `tanzu plugin install --group vmware-tkg/default:vMAJOR` , or `vMAJOR.MINOR`, etc...)

It would be nice if the Tanzu CLI periodically alerted the user to a different recommended CLI version.

The Central Configuration concept of #699 can be used to update the list of recommended versions as soon as a new release of the CLI is available and the CLI can notify the user when such a new release is detected.

**Additional context**

marckhouzam:

**This PR may seem very big but notice it has 1750 lines of new unit/e2e tests.**

This PR teaches the CLI to detect a new release of the CLI itself and notify the user they should upgrade.
The notification is printed at most once every 24 hours by default.

The PR also implements a yaml data-store that the CLI can use to store different information that does not belong in the configuration file.  This new file is located in `$HOME/.config/.data-store.yaml`.  For this feature, the CLI stores the timestamp of the last time the user was notified to upgrade.

The PR also implements an API to read the Central Configuration obtained from the Central Repository of plugins, as introduced in #700. For this feature, the Central Configuration is read to detect any new version of the CLI having been released.  The PR also updates the test central repo to improve the format of the central config.

Things to know:
1. The version notification includes the recommended version for ~Major,~ Minor and Patch (see the testing section below for an example)
    - ~In practice, there should never be a line for Major since we expect all users to be using some `v1` version.  If a user is still using the `v0.90` version, then the line for the MAJOR version will be shown~.
    - In practice, we mostly release minor versions of the CLI (the only patch up to date has been `v0.90.1`) so in most cases there will be a single line showing the new minor
2. Pre-releases are _only_ recommended if the current CLI version is also a pre-release
5. The notification can be triggered by any CLI command, including plugin commands, except for `tanzu __complete`, `tanzu completion` and `tanzu version`
6. The notification is printed _after_ the execution of the current command.
8. The version notification output is sent to `stderr` so that it does not affect potential parsing of the real command output
9. The variable `TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS` can be used to change the period of the notification (default 1 day) or to disable the feature by setting it to `0`.  If this variable is set to a negative number then the delay will be in seconds instead of days; this allows for testing.
10. If any of the recommended versions (Minor or Patch) is lower than the current version, it is ignored.
11.  The format of the value for `cli.core.cli_recommended_versions` in the central config is changed from a comma-separated list to a yaml array of structs.  With the latest central config API this is easier to parse, it is also easier for a human to read, and it even allows to add comments.  The use of structs makes it future-proof as new fields can be added in the future without breaking existing CLIs.

#### When to update the central config

Whenever there is a new release of the CLI, be it major, minor, patch or even pre-release, we should add the new version to the production Central Config in the existing array of key `cli.core.cli_recommended_versions` and publish a new image of the production central repo.  There is no rush to do it but no message about a new version will be printed until it is published.  We probably only want to publish once the version is GA (except for pre-releases). 

#### Real scenarios

**When v1.3.0 is released with this feature**

No notification will be printed since existing CLIs (pre `v1.3.0`) don't have this feature.
A notification of a new recommended version will only happen once we release a new minor or patch _after_ the `v1.3.0` version.

The central config should still be updated with the list of recommended version:
```
cli.core.cli_recommended_versions:
- version: v1.3.0
- version: v1.2.0
- version: v1.1.0
- version: v1.0.0
- version: v0.90.1
```

**A new minor release of the CLI is published**

As soon as the release is available,  users can install it.  However, no new message will be shown until we update the Central Configuration with this new version.

If a user installs the new version, no message will be printed either although the current central config recommends an older version, it is for this use case that we don't print a message to downgrade to an older version, as it would not be valid in this case.

When GA happens we then add the new minor version to the production Central Config in the existing array of key `cli.core.cli_recommended_versions` and publish a new image of the production central repo.  At this time a running older CLI will notice the new version BUT only when it refreshes its central repo data, which will happen at most within 24 hours of usage (thanks to the essential plugins threshold of #637); once the CLI refreshes its central repo it will print the notification of the new version, at most once every 24 hours.

**A new patch release of the CLI is published**

Same as for a minor except that in the central config, the existing patch version of the same minor should be replaced.  For example if the central config recommends `v1.3.0` and we release `v1.3.1`, then the entry `v1.3.0` should be replaced by `v1.3.1`.

**A new major release of the CLI is published**

Same as for a minor.
Note that although the CLI won't recommend a new major, we do want to include the new major in the central config so that people that start using that major can eventually get notifications of a new minor/patch for that major.

**A new pre-release of the CLI is published**

Depending on if this pre-release is for a new minor or not, the pre-release should be added (new minor) or updated (existing minor) into the central config.

**If the pre-release is for a new patch version, then the recommended version of the previous patch should NOT be removed, and the pre-release should be added as a new recommended entry in the central config**.  For example, if `v1.3.4` is a recommended version and a `v1.3.5-alpha.1` is released, then the `v1.3.4` entry should not be removed until the final `v1.3.5` is available; the `v1.3.5-alpha.1` version should be added as a new entry in the central config.

No need to wait for GA for a pre-release.

Fixes #706 

Added unit tests. and E2E tests.

#### Manual testing:

Setup.  Notice the recommended versions that are part of the test central repo (at the end of the setup):
```
[...]
[i] Refreshing plugin inventory cache for "localhost:9876/tanzu-cli/plugins/central:small", this will take a few seconds.

$ grep recommended_versions ~/.cache/tanzu/plugin_inventory/default/central_config.yaml -A10
cli.core.cli_recommended_versions:
- version: v2.1.0-alpha.2
- version: v2.0.2
- version: v1.5.0-beta.0
- version: v1.4.4
- version: v1.3.3
- version: v1.2.2
- version: v1.1.1
- version: v0.90.0
cli.core.some-string: "the meaning of life, the universe, and everything"
cli.core.some-int: 42
```

Run a CLI with a version that is not recommended:
```
$ make build BUILD_VERSION=v1.4.3

# Any command triggers the check for recommended versions.
# This is the very first check, so it will happen and then the 24 hours delay will start for the next check
$ tz plugin list
[i] The tanzu cli essential plugins are outdated and are being updated now. The update may take a few seconds.
[i] Installing plugins from plugin group 'vmware-tanzucli/essentials:v9.9.9'
[i] Installed plugin 'telemetry:v9.9.9' with target 'global' (from cache)

Standalone Plugins
  NAME         DESCRIPTION                                             TARGET      VERSION        STATUS
  appsv2       Applications on Kubernetes for TAP (SaaS distribution)  kubernetes  v0.1.1-beta.2  installed
  supplychain  supplychain management                                  kubernetes  v0.1.0-beta.2  installed
  telemetry    telemetry functionality                                 global      v9.9.9         installed
  workload     create, update, view and list Tanzu Workloads.          kubernetes  v0.1.0-beta.2  installed

==
Note: A new version of the Tanzu CLI is available. You are at version: v1.4.3.
To benefit from the latest security and features, please update to a recommended version:
  - v1.4.4

This message will print at most once per 24 hours until you update the CLI.
Set TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS to adjust this period (0 to disable).

# Run the command again (or any other) and notice there is no notification since the
# last notification was less than 24 hours ago
$ tz plugin list
Standalone Plugins
  NAME         DESCRIPTION                                             TARGET      VERSION        STATUS
  appsv2       Applications on Kubernetes for TAP (SaaS distribution)  kubernetes  v0.1.1-beta.2  installed
  supplychain  supplychain management                                  kubernetes  v0.1.0-beta.2  installed
  telemetry    telemetry functionality                                 global      v9.9.9         installed
  workload     create, update, view and list Tanzu Workloads.          kubernetes  v0.1.0-beta.2  installed
```

Run the CLI with a version that is the latest minor but not the latest patch, and notice only the newer patch is recommended:
```
$ make build BUILD_VERSION=v1.4.2

$ TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS=-5 tanzu plugin source list
  NAME     IMAGE
  default  localhost:9876/tanzu-cli/plugins/central:small

==
Note: A new version of the Tanzu CLI is available. You are at version: v1.4.2.
To benefit from the latest security and features, please update to a recommended version:
  - v1.4.4

This message will print at most once per 5 seconds until you update the CLI.
Set TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS to adjust this period (0 to disable).
```

Test that nothing is recommended if the recommended version is older
```
$ make build BUILD_VERSION=v1.4.5

$ TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS=-5 tanzu plugin source list
  NAME     IMAGE
  default  localhost:9876/tanzu-cli/plugins/central:small
```

Test the configuration of the notification delay
```
$ make build BUILD_VERSION=v1.3.0

# Remove the data-store.yaml file to pretend this is the first version check
$ rm ~/.config/tanzu/.data-store.yaml; \
tz plugin group search; \
sleep 7; \
TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS=-8 tz plugin search --name cluster -t k8s; \
sleep 1; \
TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS=-8 tz plugin search --name cluster -t k8s

# The first prints the notification
  GROUP                       DESCRIPTION                                 LATEST
  vmware-tanzucli/essentials  Desc for vmware-tanzucli/essentials:v9.9.9  v9.9.9
  vmware-tkg/default          Desc for vmware-tkg/default:v9.9.9          v9.9.9
  vmware-tkg/shortversion     Desc for vmware-tkg/shortversion:v9.9.9     v9.9.9
  vmware-tmc/tmc-user         Desc for vmware-tmc/tmc-user:v9.9.9         v9.9.9

Note: To view all plugin group versions available, use 'tanzu plugin group search --show-details'.

==
Note: A new version of the Tanzu CLI is available. You are at version: v1.3.0.
To benefit from the latest security and features, please update to a recommended version:
  - v1.4.4
  - v1.3.3

This message will print at most once per 24 hours until you update the CLI.
Set TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS to adjust this period (0 to disable).

# The next command is run 7 seconds later but we set TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS
# to 8 seconds, so there is no notification
  NAME     DESCRIPTION            TARGET      LATEST
  cluster  cluster functionality  kubernetes  v9.9.9

# One second later, when we again set TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS
# to 8 seconds, we get the notification as expected
  NAME     DESCRIPTION            TARGET      LATEST
  cluster  cluster functionality  kubernetes  v9.9.9

Note: A new version of the Tanzu CLI is available. You are at version: v1.3.0.
To benefit from the latest security and features, please update to a recommended version:
  - v1.4.4
  - v1.3.3

This message will print at most once per 8 seconds until you update the CLI.
Set TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS to adjust this period (0 to disable).
```

Run a CLI with a pre-release version and notice that pre-release recommendations are now shown:
```
$ make build BUILD_VERSION=v1.3.0-alpha.1

$ TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS=-5 tanzu context list
  NAME  ISACTIVE  TYPE

==
Note: A new version of the Tanzu CLI is available. You are at version: v1.3.0-alpha.1.
To benefit from the latest security and features, please update to a recommended version:
  - v1.5.0-beta.0
  - v1.3.3

This message will print at most once per 5 seconds until you update the CLI.
Set TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS to adjust this period (0 to disable).
```

Check the 3 commands that don't trigger the check and notice there is no notification:
```
$ TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS=-1 tanzu version
version: v1.3.0-alpha.1
buildDate: 2024-03-18
sha: cd9962ab5
arch: arm64

$ TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS=-1 tanzu completion bash
# bash completion V2 for tanzu                                -*- shell-script -*-
[...]
if [[ $(type -t compopt) = "builtin" ]]; then
    complete -o default -F __start_tanzu tanzu
else
    complete -o default -o nospace -F __start_tanzu tanzu
fi

# ex: ts=4 sw=4 et filetype=sh

$ TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS=-1 tanzu __complete c
completion	Output shell completion code
config	Configuration for the CLI
context	Configure and manage contexts for the Tanzu CLI
:4
```

Check that the notification is on `stderr`:
```
$ TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS=-1 tanzu plugin list > /dev/null

==
Note: A new version of the Tanzu CLI is available. You are at version: v1.3.0-alpha.1.
To benefit from the latest security and features, please update to a recommended version:
  - v1.5.0-beta.0
  - v1.3.3

This message will print at most once per 1 seconds until you update the CLI.
Set TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS to adjust this period (0 to disable).
```

### Release note

```release-note
On execution, the CLI will notify the user of an available new version of the CLI itself, at most once every 24 hours.
```

If you want to test this, you will need to use the latest test central repo.
If you haven't yet, don't forget to cleanup your test central repo cache to get the latest one:
```
$ cd tanzu-cli
$ \rm -rf hack/central-repo/registry-content
```


pkg/recommendedversion/recommended_version.go
```diff
@@ -0,0 +1,263 @@
...
+	} else {
+		fmt.Fprintf(writer, "Note: A new version of the Tanzu CLI is available. You are at version: %s.\n", currentVersion)
+		fmt.Fprintln(writer, "To benefit from the latest security and features, please update to a recommended version:")
+	}
+
+	if major != "" {
+		if utils.IsNewVersion(major, currentVersion) {
+			fmt.Fprintf(writer, "  - %s (new features but breaking changes)\n", major)
+		} else {
+			fmt.Fprintf(writer, "  - %s ([!] you should downgrade to a previous major version)\n", major)
```
anujc25, line 238: If I understand this correctly and if we are showing this when the recommended version is lower than currently installed, I am a bit concerned with this warning about downgrading in general. 

1. Why would we want to suggest something like this?
2. What if we released a newer version of the CLI but the central configuration is still recommending an older version? This can easily happen in air-gapped scenarios.

_(Same applies to minor and patch version downgrades as well)_

pkg/recommendedversion/recommended_version.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+	}
+	if minor != "" {
+		if utils.IsNewVersion(minor, currentVersion) {
+			fmt.Fprintf(writer, "  - %s (new features with backwards-compatibility)\n", minor)
```
anujc25, line 243: This feels too verbose to me. Anyone knowing the Tanzu CLI version policy should understand this and we can get away with just specifying `(new features)` for minor upgrades. WDYT?
```suggestion
			fmt.Fprintf(writer, "  - %s (new features)\n", minor)
```

anujc25 (commented):
I am still reviewing the PR but adding some early comments.

pkg/recommendedversion/recommended_version.go
```diff
@@ -0,0 +1,263 @@
...
+
+	if utils.IsNewVersion(currentVersion, major) || utils.IsNewVersion(currentVersion, minor) || utils.IsNewVersion(currentVersion, patch) {
+		fmt.Fprintf(writer, "WARNING: Due to a problem it is recommended not to use the current version: %s.\n", currentVersion)
+		fmt.Fprintln(writer, "Please use a recommended version:")
+	} else {
+		fmt.Fprintf(writer, "Note: A new version of the Tanzu CLI is available. You are at version: %s.\n", currentVersion)
+		fmt.Fprintln(writer, "To benefit from the latest security and features, please update to a recommended version:")
+	}
+
+	if major != "" {
```
anujc25, line 234: **Thought:** Should we be showing this notification when we come out with a newer major version? The major version is a big change and with the breaking changes, it is unlikely that the user will be upgrading the CLI to the new Major version. In that case, the user will need to specify `TANZU_CLI_RECOMMEND_VERSION_DELAY_SECONDS` to a very high value to avoid this additional message.

Considering that, I am thinking whether we should just be showing this warning for `minor` and `patch` version upgrades?

pkg/recommendedversion/recommended_version.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/vmware-tanzu/tanzu-cli/pkg/datastore"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/utils"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
+)
+
+// dataStoreLastVersionCheckKey is the data store key used to store the last
+// time the version check was done
+const (
+	centralConfigRecommendedVersionsKey = "cli.core.cli_recommended_versions"
```
anujc25, line 32: After thinking a bit more and the way the API is designed to get the Key/Value, I feel this might cause potential confusion to the user. (I should have raised this as part of #700 but will put my thoughts here as the other one is merged).

I see that we are using key names as `cli.core.cli_recommended_versions` in the central configuration. Considering this is a yaml file that can grow in the future, should we be using a yaml way of defining this instead of just plain `key:value`?

For example:
```
cli:
  core:
    cli_recommended_versions: 
      - v2.1.0-alpha.2
      - v2.0.2
      - v1.5.0-beta.0
      - v1.4.4
      - v1.3.3
      - v1.2.2
      - v1.1.1
      - v0.90.0
```

So when retrieving the `cli_recommended_version` information we can invoke the API as `GetCentralConfigEntry("cli.core.cli_recommended_versions")`? 
This can be done using https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured package in a generalized way as the package allows different APIs to retrieve a specific nested resource from `map[string]interface{}`.

_Just adding my thoughts here so that we can think which approach is better and discuss further if needed._

pkg/recommendedversion/recommended_version.go
marckhouzam: Those are excellent point.  Maybe we should remove the "downgrade" aspect from the feature.

My idea was that let's say we release `v1.3.0` and a couple days after we find it has a serious bug but it will take a couple more days before we can fix it in `v1.3.1`, then we could just remove `v1.3.0` from the list of recommended versions and the user would be told not to use that version.

But your points are too important.  So, if we want to support a "downgrade" aspect, we may need to design more for it.  For example, we could have a different entry in the central config that lists "version not to be used", and only if we add a version that that second list would the WARNING be shown.

But for now, I think I should remove the WARNING.

Thanks catching this problem!

pkg/recommendedversion/recommended_version.go
marckhouzam: Yeah, that's good for me.

The UX altogether may be too verbose, so don't hesitate to point out what you don't like, even if you don't have a better suggestion to give.

pkg/recommendedversion/recommended_version.go
marckhouzam: Another good point.

I've seen many tool print things like "mytool v4.0 is coming soon! look for the release", so I thought showing the MAJOR version would be nice.  But you are right that if I don't want the MAJOR Version I will be annoyed that the CLI keeps mentioning it.

I think MAJORs are so rare that we can remove the printout.

pkg/recommendedversion/recommended_version.go
```diff
@@ -0,0 +1,263 @@
...
+	// Put a delimiter before this notification so the user
+	// can see it is not part of the command output
+	fmt.Fprintln(writer, "\n==")
+
+	if utils.IsNewVersion(currentVersion, major) || utils.IsNewVersion(currentVersion, minor) || utils.IsNewVersion(currentVersion, patch) {
+		fmt.Fprintf(writer, "WARNING: Due to a problem it is recommended not to use the current version: %s.\n", currentVersion)
+		fmt.Fprintln(writer, "Please use a recommended version:")
+	} else {
+		fmt.Fprintf(writer, "Note: A new version of the Tanzu CLI is available. You are at version: %s.\n", currentVersion)
+		fmt.Fprintln(writer, "To benefit from the latest security and features, please update to a recommended version:")
```
vuil, line 231: nit: "latest security and features" doesn't read right. You mean "latest security improvements and features", or "latest security and feature enhancements", perhaps?

That said, is that statement always true with every update?

pkg/recommendedversion/recommended_version.go
marckhouzam: My thought was that each feature using the central config would have the knowledge of the data it needs.
So, the API allows key:value only as `value` is very feature-specific and it is the feature code that will know how to decode it.  I didn't feel the central config API should be into too much flexibility but instead leave the feature deal with that.  For example, if a feature wants to save a yaml text as the `value` then it can fetch the whole `value` using this API and then use `yaml.Unmarshal`.  

That way, the feature (or later on plugins) could use json, csv, or any format they want, not only yaml.

But I'll admit I'm having trouble seeing if the API would be annoying to use.
If you have examples in mind that would help me better understand.

pkg/recommendedversion/recommended_version.go
vuil: I share similar question about major version downgrade message.

I am not sure how the "we do publish plugins earlier" in point 2 is relevant here.

Finally, under what situation can this possibly happen? If the airgap situation is the only realistic situation maybe a different message is warranted ( to the effect of "your airgap central repo state is wrong, contact your administrator if you want to use this version, ...)

For non airgap situations is a scenario where the new CLI major version is picked up but the cached central config state still needs up to default 30m to show up on the CLI host. How do we plan to address this? D we have a means to invalidate the cached data on a new CLI install ?

vuil:
Based on a few comments, it might be good to understand procedurally how time-sensitive it is with regards to when to publishing the CLI and when to publish the recommendation data in the Central Configuration. Maybe update the PR description with this info?

pkg/recommendedversion/recommended_version.go
```diff
@@ -0,0 +1,263 @@
...
+		return
+	}
+	recommendedVersions, err := SortRecommendedVersionsDescending(value)
+	if err != nil {
+		log.V(7).Error(err, "failed to sort recommended versions")
+		return
+	}
+
+	currentVersion := buildinfo.Version
+	includePreReleases := utils.IsPreRelease(currentVersion)
```
vuil, line 71: This is a reasonable heuristic (consider prerelease recommendations only if current cli is itself a prerelease) to use a default, but it would be nice if one can configure if prereleases should always be or always not be considered as well. 

maybe a helper called utils.IncludePreReleaseRecommendations(currentVersion) can utilize isPreRelease(currentVersion) _and_ some config or envvar setting?

pkg/recommendedversion/recommended_version.go
anujc25: > I am not sure how the "we do publish plugins earlier" in point 2 is relevant here, however. @anujc25

Sorry for the confusion. I mean publish CLI earlier. What I meant is Core CLI gets published to package managers much earlier (1 week) before we announce GA.

pkg/recommendedversion/recommended_version.go
anujc25: The main concern I have is that, if the configuration grows in the future, it might be hard to maintain and might not be easily readable.

>So, the API allows key:value only as value is very feature-specific and it is the feature code that will know how to decode it. I didn't feel the central config API should be into too much flexibility but instead leave the feature deal with that. For example, if a feature wants to save a yaml text as the value then it can fetch the whole value using this API and then use yaml.Unmarshal.

I was thinking that implementing a fine-grained (specific) API to get the required information might be sufficient to address this. Meaning instead of providing the `GetCentralConfigEntry` API as part of the `CentralConfig` interface, we can implement `GetRecommenededVersions` directly. 

I will spend some time rethinking if there are any other limitations as I am not able to think of anything besides the readability of this (possibly a huge) file. But as the user does not need to read that file, we can get away with the way it is defined now.

anujc25: Thanks for explaining this. I got confused earlier but the way it is implemented should be fine. So, this type of key in only for a root level key, and we can define anything below that key.

In that case, is it safe to say we can define the above configuration as:
```
cli.core.cli_recommended_versions: 
      - v2.1.0-alpha.2
      - v2.0.2
      - v1.5.0-beta.0 
      - v1.4.4
      - v1.3.3  # Some comment
      - v1.2.2
      - v1.1.1.  # This structure will allow us to add any specific comment to a version
      - v0.90.0
```

Basically, instead of using `string` for the version, can we use an array?

pkg/centralconfig/central_config.go
```diff
@@ -0,0 +1,39 @@
...
+
+// CentralConfigEntryKey represents the key of a central configuration entry.
+type CentralConfigEntryKey struct {
+	Key string
+}
+
+// CentralConfigEntryValue represents the value of a central configuration entry.
+type CentralConfigEntryValue struct {
+	Value interface{}
+}
```
anujc25, line 25: nit:
I am trying to understand the reasoning behind creating a wrapper here instead of just accepting `key string` as an input and returning `value interface{}` directly. 

Are we thinking about adding more metadata with the `CentralConfigEntryKey` and `CentralConfigEntryValue`?

pkg/recommendedversion/recommended_version.go
marckhouzam: Good idea. I’ll read the yank-array value with the API then unmarshal it to get a real array. 

I’ll update the PR

marckhouzam: yank-array -> yaml-array

no joke, auto-correct changed it on me not once but three times!

pkg/centralconfig/central_config.go
marckhouzam: I did this because I was thinking that if we want to provide this API to plugins, then having a type, as you say, allows to add more fields in a backwards-compatible way.

That being said, I do find it heavy and it may stay in the CLI forever, in which case we don't care about backwards-compatibility.

I'm ok to simplify it and not use a wrapper.  Let me know if you agree.

pkg/datastore/data_store.go
```diff
@@ -0,0 +1,160 @@
...
+// DeleteDataStoreValue deletes the key and value from the data store.
+// The previous value of the key is returned.
+func DeleteDataStoreValue(key DataStoreKey) (DataStoreValue, error) {
+	content, err := getDataStoreContent(true)
+	if err != nil {
+		return nil, err
+	}
+
+	deletedValue := content[key]
+	delete(content, key)
```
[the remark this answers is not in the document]
marckhouzam: I'll go through the code to see if that can bring more flexibility.

pkg/recommendedversion/recommended_version.go
marckhouzam: > My idea was that let's say we release `v1.3.0` and a couple days after we find it has a serious bug but it will take a couple more days before we can fix it in `v1.3.1`, then we could immediately remove `v1.3.0` from the list of recommended versions and the user would be told not to use that version.

Let's consider this another feature that would need more design if we ever choose to add it (I don't think we will 😉 )

pkg/recommendedversion/recommended_version.go
marckhouzam: Now that we won't recommend a new version, there will only be two lines at most.
I think @anujc25, users will be able to understand the difference, so I'll remove the explanations.

pkg/recommendedversion/recommended_version.go
marckhouzam: Now that we won't recommend a new major version, there will only be two lines at most.
I think @anujc25 you are right, users will be able to understand the difference, so I'll remove the explanations.

pkg/datastore/data_store.go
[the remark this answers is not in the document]
marckhouzam: Done

pkg/recommendedversion/recommended_version.go
marckhouzam: Working with @anujc25 we improved the API.  `GetCentralConfigEntry(key string, out interface{})` now also takes an `out` pointer and the method will unmarshall the value into this `out` variable respecting its type.  This allows the caller to get a proper Go type that match the value being read from the central config.

marckhouzam:
> Based on a few comments, it might be good to understand procedurally how time-sensitive it is with regards to when to publish a CLI version and when to publish the recommendation data in the Central Configuration. Maybe update the PR description with this info?

I have added a `Real scenarios` and a `When to update the central config` section to the PR description.

One scenario that may not be obvious is for pre-releases:

If the pre-release is for a new patch version, then the recommended version of the previous patch should NOT be removed, and the pre-release should be added as a new recommended entry in the central config. For example, if `v1.3.4` is a recommended version and a `v1.3.5-alpha.1` is released, then the `v1.3.4` entry should not be removed until the final `v1.3.5` is available; the `v1.3.5-alpha.1` version should instead be added as a new entry in the central config.

marckhouzam:
I've updated the API for the new data store to use the same approach than the Central Configuration API (marshall/unmarshall the value stored) as it is a nicer approach.

I have also added documentation about:
- the central configuration
- the CLI data store
- the detection of new CLI releases

pkg/recommendedversion/recommended_version.go
```diff
@@ -0,0 +1,268 @@
...
+		(patch == "" || !utils.IsNewVersion(patch, currentVersion)) {
+		// The current version is the best recommended version
+		return
+	}
+
+	// Put a delimiter before this notification so the user
+	// can see it is not part of the command output
+	fmt.Fprintln(writer, "\n==")
+	fmt.Fprintf(writer, "Note: A new version of the Tanzu CLI is available. You are at version: %s.\n", currentVersion)
+	fmt.Fprintln(writer, "To benefit from the latest security and features, please update to a recommended version:")
```
vuil, line 238: (my previous comment was gone, re-adding in case you missed it)

nit: "latest security and features" doesn't read right. You mean "latest security improvements and features", or "latest security and feature enhancements", perhaps?

That said, is that statement always true with every update?
I mean, is it possible a patch does not contain security-related enhancements at all?
...
which leads me to thinking (as a future feature enhancement) if the Central Config for recommended versions should also contain an optional one-liner or two describing the main purpose of the update. 
That way, the information can be used here to augment an otherwise generic message about the recommended version.

marckhouzam: Now that we won't recommend a new major version, there will only be two lines at most.
I think @anujc25 is right, users will be able to understand the difference, so I have removed the explanations since they were contentious.

But I do like your suggestion of adding an "optional one-liner or two describing the main purpose of the update".
This makes me worry about the future-proofing of the current recommendedVersions format.  If we want to add a description later, we'll need to use a new key since the current key/value is read as an array of strings, which is not cleanly extensible.

We could instead read it as a struct containing a `[]string`, so that we could add a new field later.

What do you think?

pkg/recommendedversion/recommended_version.go
marckhouzam: We have agreed to do this as a follow-up

pkg/recommendedversion/recommended_version.go
marckhouzam: After discussion we agreed to try to do this.
It turned out to be easy enough that I added it as a top-commit on this PR.

docs/quickstart/install.md
```diff
@@ -503,11 +503,23 @@ private plugin discovery source. Verify that plugins are discoverable by
...
+
+Trick: use the very small `vmware-tanzucli/essentials` plugin group if you only want to update the Central Configuration,
+i.e, `tanzu plugin download-bundle --group vmware-tanzucli/essentials --to-tar update_config.tar`
+
 ### Interacting with a central repository hosted on a registry with self-signed CA or with expired CA
 
 If a user has configured a central repository on a custom registry (e.g. air-gaped environment) with a self-signed CA or
-if the
-registry CA
+if the registry CA
```
vuil, line 522: supernit: the wrapping is still weird after the change.

vuil (approved):
Looks good to me. This is a really nice set of changes, @marckhouzam !

A few things:
I assume you plan to do some rebase/squashing. If you do, please retain a handful of logical commits so it will be easier for future reference.

Also, something to consider as a followup:
Over time, it will become more and more important that a CLI has up-to-date information about the central configuration, else might start to exhibit different/possibly unwanted behavior. 
The situation that is really important to get right (though it is likely not the only one) is when the user picks up a Central-Config aware CLI for the first time.
I suggest that any caching behavior we have today with regards to the fetching and updating of the data from the centrl OCI image be examined. 
The conservative thing to do would be to invalidate the cache on any new CLI version.
Few options if we are to do this:
track in the data store file any CLI version(s) that the CLI host has seen, and perform any cleanup/invalidation (and update of the data store file with the new version, of course) if we encounter a new CLI version.
Alternatively, store in the cache the version of the CLI responsible for populating it and treat any new cli version as an indication that the cache should be invalidate. 

There could be more uncommon situations where the cache data needs to be invalidated as well. One could be when a discovery source is updated. It is less likely to happen in production, but might be worth verifying as well.

marckhouzam:
> Looks good to me. This is a really nice set of changes, @marckhouzam !

Thank you @vuil and @anujc25 for working with me to reach this final iteration!

> A few things: I assume you plan to do some rebase/squashing. If you do, please retain a handful of logical commits so it will be easier for future reference.

Good point I have squashed on my machine and kept the following three commits (most recent at the top of the list):
- Have the CLI notify the user of a new version
- Provide an API to read the Central Configuration
- Implement a yaml data store

> Also, something to consider as a followup: Over time, it will become more and more important that a CLI has up-to-date information about the central configuration, else might start to exhibit different/possibly unwanted behavior. The situation that is really important to get right (though it is likely not the only one) is when the user picks up a Central-Config aware CLI for the first time. I suggest that any caching behavior we have today with regards to the fetching and updating of the data from the centrl OCI image be examined. 

Excellent point!
I will track this as a follow-up.  Thanks!

PR #711, opened 2024-03-21
Update the tanzu context kubeconfig to use projectID instead of projectName if the projectID is not empty string

prkalle:
This PR updates the tanzu context kubeconfig generation logic to use projectID instead of projectName if the projectID is not empty string

Fixes #

Tested updating the context with only the project name and it is sucess
```
❯ ./bin/tanzu context update tanzu-active-resource TAP_SAAS_PRD --project e2e

# kubeconfigs cluster.server URL has been updated with project name 
❯ kubectl config view --minify
apiVersion: v1
clusters:
- cluster:
    server: https://api.tanzu.cloud.vmware.com/org/bc6f01a5-8618-4fed-a457-161d3a138052/project/e2e
  name: tanzu-cli-TAP_SAAS_PRD/current
contexts:
- context:
    cluster: tanzu-cli-TAP_SAAS_PRD/current
    user: tanzu-cli-TAP_SAAS_PRD-user
  name: tanzu-cli-TAP_SAAS_PRD
current-context: tanzu-cli-TAP_SAAS_PRD
kind: Config
preferences: {}
users:
- name: tanzu-cli-TAP_SAAS_PRD-user
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1
      args:
      - context
      - get-token
      - TAP_SAAS_PRD
      command: tanzu
      env: []
      interactiveMode: IfAvailable
      provideClusterInfo: false

❯ kubectl get spaces
NAME          AGE
amit-space    11d
wfd-prode2e   11d


# CLI context additionalMetadata updated with projectName
    additionalMetadata:
        tanzuClusterGroupName: ""
        tanzuMissionControlEndpoint: https://tmc.tanzu.cloud.vmware.com
        tanzuOrgID: bc6f01a5-8618-4fed-a457-161d3a138052
        tanzuOrgName: TAP SAAS PRD
        tanzuProjectID: ""
        tanzuProjectName: e2e
        tanzuSpaceName: ""
```

Tested updating the context with both project-id and name and verified the kubeconfig genreated has cluster.server URL is using `projectID` specified instead of project name
```
❯ ./bin/tanzu context update tanzu-active-resource prem-ucp-intg-lat-stg --project beta2-workflow-demo --project-id 8aeb4bd3-3e14-4763-a3d7-1c6eb1da2b8e

# kubeconfig showing the cluster.server URL has been updated to use project-id instead of project name if project-id is provided 
❯ kubectl config view --minify
apiVersion: v1
clusters:
- cluster:
    server: https://api.tanzu-dev.cloud.vmware.com/org/b1d48027-bb69-4a56-a5b8-e941ef29fa4b/project/8aeb4bd3-3e14-4763-a3d7-1c6eb1da2b8e
  name: tanzu-cli-prem-ucp-intg-lat-stg/current
contexts:
- context:
    cluster: tanzu-cli-prem-ucp-intg-lat-stg/current
    user: tanzu-cli-prem-ucp-intg-lat-stg-user
  name: tanzu-cli-prem-ucp-intg-lat-stg
current-context: tanzu-cli-prem-ucp-intg-lat-stg
kind: Config
preferences: {}
users:
- name: tanzu-cli-prem-ucp-intg-lat-stg-user
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1
      args:
      - context
      - get-token
      - prem-ucp-intg-lat-stg
      command: tanzu
      env: []
      interactiveMode: IfAvailable
      provideClusterInfo: false

```

Tested with the test environment where the project UCP resource is named with project id and ucp resource has annotation for project displayName
```
❯ ./bin/tanzu context create testprojectsyncer --type tanzu --staging --endpoint https://test-project-syncer.stacks.bluesky.tmc-dev.cloud.vmware.com/
Log in by visiting this link:

    https://console-stg.cloud.vmware.com/csp/gateway/discovery?client_id=tanzu-cli-client-id&code_challenge=HbKyNt8N0ShSN_bz7pPXHwQDKv6dFTqFvmvVBqZDdqg&code_challenge_method=S256&redirect_uri=http%3A%2F%2F127.0.0.1%3A51618%2Fcallback&response_type=code&state=5853660f83c71e77b9baead92b8fc8a2

[ok] Successfully logged into 'TestProjectSyncer' organization and created a tanzu context
[i] Checking for required plugins for context 'testprojectsyncer'...
❯ ./bin/tanzu context list
  NAME                                ISACTIVE  TYPE             PROJECT      SPACE
  TAP_SAAS_PRD                        false     tanzu            sre-project  test
  TestProjectSyncer-staging-3d4f75b3  false     tanzu
  mytmc-ctx                           true      mission-control  n/a          n/a
  prem-ucp-prod-test                  false     tanzu            sre-project  test
  testprojectsyncer                   true      tanzu
  tkg-mgmt-vc                         false     kubernetes       n/a          n/a
  tt-test-selfmg                      false     mission-control  n/a          n/a

❯ kubectl get projects
NAME                                   DISPLAYNAME
9ba6ba75-9782-48a2-ae3b-eaa11c6e5316   newpRojje
3351bd40-5150-46f7-96c5-d368b5f99932   newProjectTestne
fd72b29b-bd90-40a6-a684-3e78de397028   Project4
default
```
Changed the tanzu context active resource to project  and tested the kubeconfig works
```
❯ ./bin/tanzu context update tanzu-active-resource testprojectsyncer --project newpRojje --project-id 9ba6ba75-9782-48a2-ae3b-eaa11c6e5316
❯ ./bin/tanzu project list
Listing projects from dcd7e146-57e6-4a1b-905c-0d4079dc9f38 org, current context is newpRojje project
NAME                                   ACTIVE   AGE
9ba6ba75-9782-48a2-ae3b-eaa11c6e5316   false    36h
3351bd40-5150-46f7-96c5-d368b5f99932   false    30h
fd72b29b-bd90-40a6-a684-3e78de397028   false    29h
default                                false    13h

❯ kubectl get spaces
No resources found

❯ kubectl config view --minify
apiVersion: v1
clusters:
- cluster:
    server: https://test-project-syncer.stacks.bluesky.tmc-dev.cloud.vmware.com//org/dcd7e146-57e6-4a1b-905c-0d4079dc9f38/project/9ba6ba75-9782-48a2-ae3b-eaa11c6e5316
  name: tanzu-cli-testprojectsyncer/current
contexts:
- context:
    cluster: tanzu-cli-testprojectsyncer/current
    user: tanzu-cli-testprojectsyncer-user
  name: tanzu-cli-testprojectsyncer
current-context: tanzu-cli-testprojectsyncer
kind: Config
preferences: {}
users:
- name: tanzu-cli-testprojectsyncer-user
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1
      args:
      - context
      - get-token
      - testprojectsyncer
      command: tanzu
      env: []
      interactiveMode: IfAvailable
      provideClusterInfo: false
```

### Release note

```release-note

```

### DIFF
--- a/pkg/command/context.go
+++ b/pkg/command/context.go
@@ -1532,15 +1532,8 @@ var tanzuActiveResourceCmd = &cobra.Command{
 func setTanzuCtxActiveResource(_ *cobra.Command, args []string) error {
 	name := args[0]
 
-	if spaceStr != "" && clustergroupStr != "" {
-		return errors.Errorf("either space or clustergroup can be set as active resource. Please provide either --space or --clustergroup option")
-	}
-	// TODO(prkalle): Add condition to check projectID is also set if project name set
-	if projectStr == "" && spaceStr != "" {
-		return errors.Errorf("space cannot be set without project name. Please provide project name also using --project option")
-	}
-	if projectStr == "" && clustergroupStr != "" {
-		return errors.Errorf("clustergroup cannot be set without project name. Please provide project name also using --project option")
+	if err := validateActiveResourceOptions(); err != nil {
+		return err
 	}
 
 	ctx, err := config.GetContext(name)
@@ -1561,7 +1554,13 @@ func setTanzuCtxActiveResource(_ *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.Wrap(err, "failed updating the context %q with the active tanzu resource")
 	}
-	err = updateTanzuContextKubeconfig(ctx, projectStr, spaceStr, clustergroupStr)
+	// TODO (prkalle): Adding this fallback logic to support the backward compatibility. This should be updated to use projectID for official release
+	// If the projectIDStr is set it would be used for kubeconfig generation, else use the project name
+	projectVal := projectStr
+	if projectIDStr != "" {
+		projectVal = projectIDStr
+	}
+	err = updateTanzuContextKubeconfig(ctx, projectVal, spaceStr, clustergroupStr)
 	if err != nil {
 		return errors.Wrap(err, "failed to update the tanzu context kubeconfig")
 	}
@@ -1569,7 +1568,26 @@ func setTanzuCtxActiveResource(_ *cobra.Command, args []string) error {
 	return nil
 }
 
-func updateTanzuContextKubeconfig(cliContext *configtypes.Context, projectName, spaceName, clustergroupName string) error {
+func validateActiveResourceOptions() error {
+	if spaceStr != "" && clustergroupStr != "" {
+		return errors.Errorf("either space or clustergroup can be set as active resource. Please provide either --space or --clustergroup option")
+	}
+
+	// TODO(prkalle): Need to update the checks to make project ID and project Name mandatory for official release
+	if (projectStr == "" && projectIDStr == "") && spaceStr != "" {
+		// TODO(prkalle): update the error message later for official release to use --project and --project-id options to set the project
+		return errors.Errorf("space cannot be set without project. Please set the project")
+	}
+
+	if (projectStr == "" && projectIDStr == "") && clustergroupStr != "" {
+		// TODO(prkalle): update the error message later for official release to use --project and --project-id options to set the project
+		return errors.Errorf("clustergroup cannot be set without project. Please set the project")
+	}
+
+	return nil
+}
+
+func updateTanzuContextKubeconfig(cliContext *configtypes.Context, project, spaceName, clustergroupName string) error {
 	kcfg, err := clientcmd.LoadFromFile(cliContext.ClusterOpts.Path)
 	if err != nil {
 		return errors.Wrap(err, "unable to load kubeconfig")
@@ -1580,7 +1598,7 @@ func updateTanzuContextKubeconfig(cliContext *configtypes.Context, projectName, 
 		return errors.Errorf("kubecontext %q doesn't exist", cliContext.ClusterOpts.Context)
 	}
 	cluster := kcfg.Clusters[kubeContext.Cluster]
-	cluster.Server = prepareClusterServerURL(cliContext, projectName, spaceName, clustergroupName)
+	cluster.Server = prepareClusterServerURL(cliContext, project, spaceName, clustergroupName)
 	err = clientcmd.WriteToFile(*kcfg, cliContext.ClusterOpts.Path)
 	if err != nil {
 		return errors.Wrap(err, "failed to update the context kubeconfig file")
@@ -1588,12 +1606,12 @@ func updateTanzuContextKubeconfig(cliContext *configtypes.Context, projectName, 
 	return nil
 }
 
-func prepareClusterServerURL(context *configtypes.Context, projectName, spaceName, clustergroupName string) string {
+func prepareClusterServerURL(context *configtypes.Context, project, spaceName, clustergroupName string) string {
 	serverURL := context.ClusterOpts.Endpoint
-	if projectName == "" {
+	if project == "" {
 		return serverURL
 	}
-	serverURL = serverURL + "/project/" + projectName
+	serverURL = serverURL + "/project/" + project
 
 	if spaceName != "" {
 		return serverURL + "/space/" + spaceName
--- a/pkg/telemetry/metrics_helper.go
+++ b/pkg/telemetry/metrics_helper.go
@@ -58,10 +58,16 @@ func hashString(str string) string {
 }
 
 func computeEndpointSHAForTanzuContext(ctx *configtypes.Context) string {
-	// returns SHA256 of concatenated string of Endpoint and orgId/ProjectName/SpaceName
+	// use projectID if not empty, else use projectName
+	projectVal := stringValue(ctx.AdditionalMetadata[configlib.ProjectNameKey])
+	projectID := stringValue(ctx.AdditionalMetadata[configlib.ProjectIDKey])
+	if projectID != "" {
+		projectVal = projectID
+	}
+	// returns SHA256 of concatenated string of Endpoint and orgId/Project/SpaceName
 	return hashString(ctx.GlobalOpts.Endpoint +
 		stringValue(ctx.AdditionalMetadata[configlib.OrgIDKey]) +
-		stringValue(ctx.AdditionalMetadata[configlib.ProjectNameKey]) +
+		projectVal +
 		stringValue(ctx.AdditionalMetadata[configlib.SpaceNameKey]) +
 		stringValue(ctx.AdditionalMetadata[configlib.ClusterGroupNameKey]))
 }
--- a/go.mod
+++ b/go.mod
@@ -42,7 +42,7 @@ require (
 	github.com/vmware-tanzu/carvel-ytt v0.40.0
 	github.com/vmware-tanzu/tanzu-cli/test/e2e/framework v0.0.0-00010101000000-000000000000
 	github.com/vmware-tanzu/tanzu-framework/capabilities/client v0.0.0-20230523145612-1c6fbba34686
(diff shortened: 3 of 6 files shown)


pkg/command/context.go
```diff
@@ -1535,7 +1535,9 @@ func setTanzuCtxActiveResource(_ *cobra.Command, args []string) error {
 	if spaceStr != "" && clustergroupStr != "" {
 		return errors.Errorf("either space or clustergroup can be set as active resource. Please provide either --space or --clustergroup option")
 	}
-	// TODO(prkalle): Add condition to check projectID is also set if project name set
+	if (projectStr != "" && projectIDStr == "") || (projectStr == "" && projectIDStr != "") {
+		return errors.Errorf("both project name and project ID are required to set a project as active. Please provide them using --project and --project-id options")
```
marckhouzam, line 1539: Older plugins that only specify the project name will not work with this version of the CLI.
So this breaks backwards-compatibility.
I assume we can accept that because the plugins are still beta?

marckhouzam (commented):
The changes look good but I have a couple of questions.

The function `computeEndpointSHAForTanzuContext()` still uses the project name.  Is that ok?

As for `tz context use` and `tz context list` they print the project name; I assume that is ok, there is no need to include the project id?

pkg/command/context.go
prkalle: Yes, that's correct. This is a breaking change and plugins should update to use new plugin-runtime and new CLI version to work. Yes,It should be acceptable as the plugins are still in beta.

marckhouzam:
Two points I just thought of that are unclear to me:
1. if a plugin only specifies the project name or only the project id when calling `SetTanzuContextActiveResource`, does the plugin get a visible error?  I'm unable to test now so I'm putting it here so we can confirm
2. why require both project name and project id?  Would it be easier for plugins to only provide project id since it is unique?

vuil (commented):
As discussed, at least for the time being, it would be good to make project id parameter optional, so I will wait for an update.

prkalle:
> Two points I just thought of that are unclear to me:
> 
> 1. if a plugin only specifies the project name or only the project id when calling `SetTanzuContextActiveResource`, does the plugin get a visible error?  I'm unable to test now so I'm putting it here so we can confirm

Yes, plugin should get this error . Example as shown below
```
[x] : space cannot be set without project. Please set the project

```
> 2. why require both project name and project id?  Would it be easier for plugins to only provide project id since it is unique?
Since we use project name in the context list to show current `PROJECT`, it would confuse users if this is empty because user may think the project is not set.  
However, after our discussion,since we decided to support the backward compatibility,I removed this condition for now, but for official release, planning to keep this check.

vuil (approved):
lgtm.

Thanks for putting in the comment about the fallback logic being temporary.
The testing done section will need some update now that we have made the --project-id optional for `update tanzu-active-resource`.

So with this change, it is possible to 
1. have an empty string tanzuProjectID field  when additionalMetadata:.tanzuProjectName is set to something valid
2. have a context with nonempty projectName and projectID, then have "update tanzu-active-resource" clears the projectID to ""

Is 1. intentional?
Should 2. be prevented?

These are not blockers and addressable in a followup, if necessary.

prkalle:
> lgtm.
> 
> Thanks for putting in the comment about the fallback logic being temporary. The testing done section will need some update now that we have made the --project-id optional for `update tanzu-active-resource`.

Thank you! I will update the PR description with test performed with new tests

> So with this change, it is possible to
> 
> 1. have an empty string tanzuProjectID field  when additionalMetadata:.tanzuProjectName is set to something valid
> 2. have a context with nonempty projectName and projectID, then have "update tanzu-active-resource" clears the projectID to ""
> 
> Is 1. intentional? 
Yes, was trying to prevent the case where  user having a context with nonempty projectName and projectID, then user tried "update tanzu-active-resource" and just updated the tanzuProjectName which would leave the previously set tanzuProjectID in the additionalMetadata section in the context, which is inconsisten data(projectID is of previous project and project name is of different project). This scenario should be a case of user error though.

> Should 2. be prevented? (I can't think of a scenario where this is not some kind of operator error)
> 
As mentioned above for point 1, I thought setting the projectID to "" is preferred approach than leaving the stale projectID in the additionalMetadata.

pkg/telemetry/metrics_helper.go
```diff
@@ -61,7 +61,7 @@ func computeEndpointSHAForTanzuContext(ctx *configtypes.Context) string {
 	// returns SHA256 of concatenated string of Endpoint and orgId/ProjectName/SpaceName
 	return hashString(ctx.GlobalOpts.Endpoint +
 		stringValue(ctx.AdditionalMetadata[configlib.OrgIDKey]) +
-		stringValue(ctx.AdditionalMetadata[configlib.ProjectNameKey]) +
+		stringValue(ctx.AdditionalMetadata[configlib.ProjectIDKey]) +
```
marckhouzam, line 64: With backwards-compatibility support now, the `ctx.AdditionalMetadata[configlib.ProjectIDKey]` could be empty, right?  So we should pick the one that is filled between the project name or id I think

marckhouzam:
> > Two points I just thought of that are unclear to me:
> > 
> > 1. if a plugin only specifies the project name or only the project id when calling `SetTanzuContextActiveResource`, does the plugin get a visible error?  I'm unable to test now so I'm putting it here so we can confirm
> 
> Yes, plugin should get this error . Example as shown below
> 
> ```
> [x] : space cannot be set without project. Please set the project
> ```

I was able to test it with the version of the PR _before_ you added backwards-compatibility, and since the backwards-compatibility may be temporary, let's continue this discussion to be prepared for the final release.
When running an old project plugins this is what I get:
```
$ tz project use default
[x] : both project name and project ID are required to set a project as active. Please provide them using --project and --project-id options
Error: [x] : both project name and project ID are required to set a project as active. Please provide them using --project and --project-id options

$ tz project version
v0.1.0-build.4
```

If this can happen to an end-user then the error message does not help at all to know what they can do to fix things.  I think we should figure out what scenarios could happen and discuss how best to handle such cases.

pkg/telemetry/metrics_helper.go
prkalle: Yes, I missed this after the backward compatibility changes. I will update it. Thanks

prkalle:
> > Yes, plugin should get this error . Example as shown below
> > ```
> > [x] : space cannot be set without project. Please set the project
> > ```
> 
> I was able to test it with the version of the PR _before_ you added backwards-compatibility, and since the backwards-compatibility may be temporary, let's continue this discussion to be prepared for the final release. When running an old project plugins this is what I get:
> 
> ```
> $ tz project use default
> [x] : both project name and project ID are required to set a project as active. Please provide them using --project and --project-id options
> Error: [x] : both project name and project ID are required to set a project as active. Please provide them using --project and --project-id options
> 
> $ tz project version
> v0.1.0-build.4
> ```
> 
> If this can happen to an end-user then the error message does not help at all to know what they can do to fix things. I think we should figure out what scenarios could happen and discuss how best to handle such cases.

First thing this should not happen as the project and space plugin would pre-check before calling the SetTanzuActiveContext  API from plugin runtime which inturn would call this command. I tested this and below is the output from the space plugin
```
❯ tanzu space use e2e-space
Error: No project found in active context of type tanzu. Use the "tanzu project use" command to select one.
```

This would only be visible to users who try this hidden command directly. Since  CLI is backward compatible if users uses this command in older environments, user should set only project name using `--project` option, and for newer environments user should atleast set projectID (but recommendation is to set projectName as well) using `--project-id`. To avoid this confusion for different environment users, I made it more generic ,and I did add a `TODO` comment to update this error for final release which would asking users to set both `--project` and `--project-id`.  I am open to suggestions to modify this error message that can fit for both environments.

marckhouzam (approved):
LGTM
Thanks @prkalle

PR #714, opened 2024-03-26
Document the user-facing CLI environment variables

marckhouzam:
### What this PR does / why we need it

Document the user-facing environment variables.

Environment variables used only in testing will be documented in a different PR.

You can read the new doc in markdown at https://github.com/marckhouzam/tanzu-cli/tree/feat/docEnvVars/docs/full#environment-variables-affecting-the-cli

### Which issue(s) this PR fixes

Fixes # N/A

### Describe testing done for PR

Visual check at https://github.com/marckhouzam/tanzu-cli/tree/feat/docEnvVars/docs/full#environment-variables-affecting-the-cli

Let CI run

### Release note

```release-note
Document the user-facing CLI environment variables.
```

PR #716, opened 2024-03-27
Document the CLI environment variables used for testing

marckhouzam:

Document the CLI environment variables used for testing
Environment variables that are user-facing are documented in #714 

You can read the new doc in markdown at https://github.com/marckhouzam/tanzu-cli/blob/feat/docTestEnvVars/docs/dev/README.md#environment-variables-for-testing-the-cli

Fixes # N/A

Visual check at https://github.com/marckhouzam/tanzu-cli/blob/feat/docTestEnvVars/docs/dev/README.md#environment-variables-for-testing-the-cli
Let CI run

### Release note

```release-note
Document the CLI environment variables used for testing
```

### DIFF
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -67,6 +67,46 @@ To run e2e tests for the repository:
 make e2e-cli-core
 ```
 
+### Environment variables for testing the CLI
+
+Some test options affecting the CLI are only available through the use of environment
+variables.  These variables are only meant to be used during testing and are not subject
+to the CLI's backwards-compatibility policy, i.e., they could be changed or removed at any time.
+These variables can be set in the shell or using `tanzu config set env.<VAR> <VALUE>`
+(see [the documentation about this command](../full/README.md#cli-configuration)).
+Below is the list of these variables:
+
+| Environment Variable | Description | Value |
+| -------------------- | ----------- | ----- |
+| `SQL_STATEMENTS_LOG_FILE` | Specifies a log file where SQL commands will be logged when _modifying_ the plugin inventory database.  This is done when publishing plugins using the `builder` plugin. | A file name with its path |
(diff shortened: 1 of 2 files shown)


docs/dev/README.md
```diff
@@ -55,6 +55,42 @@ To run e2e tests for the repository:
 make e2e-cli-core
 ```
 
+### Environment variables for testing the CLI
+
+Some test options affecting the CLI are only available through the use of environment
+variables.  Below is the list of such test variables:
```
anujc25, line 61: Should we add additional line stating that these variables are only for testing and can be updated or removed without any prior notice (or something like that)?

anujc25 (approved):
LGTM.

vuil:
lgtm, this section (as well as the one updated in #714) seems an appropriate place to reference the doc about using `tanzu config set env` as a means to persist the env vars. 
I think that detail was explained in a row in https://github.com/marckhouzam/tanzu-cli/blob/feat/docTestEnvVars/docs/full/README.md#cli-configuration.

vuil (approved):

docs/dev/README.md
marckhouzam: How about "These variables are only meant to be used during testing and are not subject to the CLI's backwards-compatibility policy, i.e., the could be changed or removed at any time."?

marckhouzam:
> lgtm, this section (as well as the one updated in #714) seems an appropriate place to reference the doc about using `tanzu config set env` as a means to persist the env vars. I think that detail was explained in a row in https://github.com/marckhouzam/tanzu-cli/blob/feat/docTestEnvVars/docs/full/README.md#cli-configuration.

Good idea.
How about: 
"These variables can be set in the shell or using `tanzu config set env.<VAR> <VALUE>`
(see [the documentation about this command](../full/README.md#cli-configuration))."

marckhouzam:
I've updated the PR with both @anujc25's and @vuil's suggestions

PR #718, opened 2024-03-28
Enable overriding a command if its alias matches the mapping candidate.

vuil:
This is temporary heuristic to ensure a command mapping candidate command replaces one whose alias matches the candidate's name.

An upcoming change would enable the plugin to be more explicit about what commands it intends to override.

Fixes #

ran unit tests

also tested by building an installing a plugin whose invokedAs only matches an alias of an existing plugin (e.g. apps plugin's alias, "app")

observe when running `tanzu` that only the invokedAs name appears in the list of top-level invocable commands

### Release note

```release-note

```

### DIFF
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -360,6 +360,18 @@ var opsCmd = &cobra.Command{
 	},
 }
 
+func matchOnCommandNameAndAliases(cmd *cobra.Command, value string) bool {
+	if cmd.Name() == value {
+		return true
+	}
+	for _, alias := range cmd.Aliases {
+		if alias == value {
+			return true
+		}
+	}
+	return false
+}
+
 func findSubCommand(rootCmd, subCmd *cobra.Command) *cobra.Command {
 	arrSubCmd := rootCmd.Commands()
 	for i := range arrSubCmd {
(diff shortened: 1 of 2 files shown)


prkalle (approved):
LGTM. Thanks!

pkg/command/root.go
```diff
@@ -370,14 +382,17 @@ func findSubCommand(rootCmd, subCmd *cobra.Command) *cobra.Command {
...
 }
 
-func findSubCommandByHierarchy(cmd *cobra.Command, hierarchy []string) (*cobra.Command, *cobra.Command) {
+func findSubCommandByHierarchy(cmd *cobra.Command, hierarchy []string, matcher func(*cobra.Command, string) bool) (*cobra.Command, *cobra.Command) {
 	childCmds := cmd.Commands()
 	for i := range childCmds {
-		if childCmds[i].Name() == hierarchy[0] {
-			if len(hierarchy) == 1 {
+		if len(hierarchy) == 1 {
+			if matcher(childCmds[i], hierarchy[0]) {
```
marckhouzam, line 389: Nice use of a matching mechanism!

marckhouzam (approved):
LGTM
Nice!

PR #723, opened 2024-04-03
Setup global initializer for missing central config file

### LINKED ISSUE #722: A new CLI installation should refresh the plugin discovery cache

ref: https://github.com/vmware-tanzu/tanzu-cli/pull/707#pullrequestreview-1963722024

CLI v1.3 will introduce a new `central_config.yaml` file in the plugin inventory OCI image (see #707). This file is copied to the cache at `~/.cache/tanzu/plugin_inventory/default/central_config.yaml`.

The problem is that older CLIs don't copy that file, even if it is in the OCI image. So, we have this problematic scenario:

1. we release CLI v1.3
2. we update the OCI image to have a `central_config.yaml` file
3. a user of CLI v1.2 uses, say, `tanzu plugin search` and downloads the latest OCI image it their cache, but _without_ `central_config.yaml` since its CLI v1.2
4. that user upgrades to CLI v1.3
=> they will not download the `central_config.yaml` as long as the OCI image content is not updated again and after they have upgraded to CLI 1.3

With the current feature of new CLI version recommendation, which needs the central config, the consequences are not bad, but if a different feature is added needing the central config in v1.3 or even later, we run the risk of having problems if the central config is not available on a user's machine.

The proposal is that whenever a version of the CLI is installed for the first time, the plugin inventory cache would be forcibly refreshed (not just the check of the OCI digest but the actual content of the OCI image).

We can store each version of the CLI that has been run into the new data store (also introduced in #707). When a CLI runs, it will check that datastore entry and if that CLI version wasn't run before it clears the cache. This opens the door to doing other things in the future upon a new CLI version installation.

**Additional context**

marckhouzam:

This PR is provided in two commits to make the review easier.

With the introduction of the Central Configuration in #707, there is now a `central_config.yaml` file part of the plugin inventory OCI image; however, CLIs < 1.3.0 do not copy this file into the cache, so this PR forces a cache invalidation and a re-download of the OCI image as soon as a CLI >= 1.3.0 notices the missing file.  This guarantees that the `central_config.yaml` file will be properly installed.  See #722 for more details.

To achieve this, the first commit of the PR implements a small framework allowing a "global initialization" of the CLI based on pre-defined triggers.  Different features can register their own trigger function and a corresponding initialization function which allow for any require global initialization.  The CLI will verify the triggers on any command execution and if the trigger function returns true, the corresponding initialization will be executed.

The first commit implements the global initialization framework.
The second commit uses the framework to initialize the plugin inventory cache if the `central_config.yaml` file is missing.

Fixes #722

#### What the new feature looks like

```
# Build a new version of the CLI
$ make build BUILD_VERSION=v1.3.1
build darwin-arm64 CLI with version: v1.3.1

# Run this new version on an empty plugin cache
rm ~/.cache/tanzu/plugin_inventory

# Note that the `tz version` command does NOT trigger the initialization
# (same for `completion` and `__complete`)
$ tz version
version: v1.3.1
buildDate: 2024-04-03
sha: b711f5630
arch: arm64

# any other command will trigger the global initializers
$ tz plugin list
Some initialization of the CLI is required.
Let's set things up for you.  This will just take a few seconds.

[i] Refreshing plugin inventory cache for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.

Initialization done!
==
Standalone Plugins
```

Run the command again to see there is no initialization anymore as it is only run once:
```
$ tz plugin list
Standalone Plugins
```

Now remove the `central_config.yaml` file to see that the CLI notices:
```
$ rm ~/.cache/tanzu/plugin_inventory/default/central_config.yaml

# Run any command
$ tz context list
Some initialization of the CLI is required.
Let's set things up for you.  This will just take a few seconds.

[i] Refreshing plugin inventory cache for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.

Initialization done!
==
  NAME                                  ISACTIVE  TYPE             PROJECT           SPACE  CLUSTERGROUP
  TAP_pre-integration-staging           false     tanzu

```

#### Main scenario motivating this PR

Prepare the test :
```
# We have an old CLI
$ ~/bin/tanzu.v1.2.0 version
version: v1.2.0
buildDate: 2024-02-07
sha: f3abe62e
arch: arm64

# And the new one built with the PR
$ tz version
version: v1.3.0-dev
buildDate: 2024-04-03
sha: b711f5630
arch: arm64

# Use a central repo that has a `central_config.yaml` file

[...]

# Start with no cache and no data store
$ rm ~/.cache/tanzu/plugin_inventory
```

Use an old CLI to first setup the cache.  This is similar to what a user of CLI v1.2 would have done.
Notice the `central_config.yaml` file is not copied to the cache
```
$ ~/bin/tanzu.v1.2.0 plugin source update default -u localhost:9876/tanzu-cli/plugins/central:small
[i] Refreshing plugin inventory cache for "localhost:9876/tanzu-cli/plugins/central:small", this will take a few seconds.

[i] Refreshing plugin inventory cache for "localhost:9876/tanzu-cli/plugins/central:small", this will take a few seconds.

# Notice that the central_config.yaml file is not in the cache
# because CLIs < 1.3.0 don't copy it over
$ ls ~/.cache/tanzu/plugin_inventory/default
digest.ca212530af87bcf8c41b4b89c48cf453c86621e5e8f9fb501d219a582878f336 plugin_inventory.db
metadata.digest.none
```

Now test the actual scenario driving this change.
Normally, when running v1.3.0-alpha.2 with the test repo we should see a notification saying v1.3.3 is available for upgrade. However, notice this does not happen because the `central_config.yaml` file is missing from the cache.

```
$ /opt/homebrew/bin/tanzu version
version: v1.3.0-alpha.2
buildDate: 2024-03-29
sha: fdbe2498
arch: arm64

$ /opt/homebrew/bin/tanzu config set env.TEST 1
# Here there should have been a notification printed about upgrading to v1.3.3 but there is not.  That is the problem.

# Let's try the same thing with this PR
# which will first force a re-download of the OCI image
$ tz config set env.TEST 1
Some initialization of the CLI is required.
Let's set things up for you.  This will just take a few seconds.

[i] Refreshing plugin inventory cache for "localhost:9876/tanzu-cli/plugins/central:small", this will take a few seconds.


Initialization done!
==

==
Note: A new version of the Tanzu CLI is available. You are at version: v1.3.0-dev.
To benefit from the latest security and features, please update to a recommended version:
  - v1.5.0-beta.0
  - v1.3.3

Please refer to these instructions for upgrading: https://github.com/vmware-tanzu/tanzu-cli/blob/main/docs/quickstart/install.md.

This message will print at most once per 24 hours until you update the CLI.
Set TANZU_CLI_RECOMMEND_VERSION_DELAY_DAYS to adjust this period (0 to disable).

# And finally, just to see it with our own eyes, check the presence of `central_config.yaml`
$ ls ~/.cache/tanzu/plugin_inventory/default
central_config.yaml                                                     metadata.digest.none
digest.ca212530af87bcf8c41b4b89c48cf453c86621e5e8f9fb501d219a582878f336 plugin_inventory.db
```

### Release note

```release-note
Implement "global initializers" and use them to fix the plugin cache if the `central_config.yaml` file is missing.
```

### DIFF
--- a/pkg/centralconfig/central_config_cache_init.go
+++ b/pkg/centralconfig/central_config_cache_init.go
@@ -1,0 +1,79 @@
+// Copyright 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package centralconfig
+
+import (
+	"io"
+	"os"
+	"path/filepath"
+
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/constants"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/discovery"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/globalinit"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+)
+
+// The Central Configuration feature uses a central_config.yaml file that gets downloaded
+// along with the plugin inventory cache and is stored in the cache.  Older versions of the
+// CLI (< 1.3.0) do not include this file in the cache.  It is therefore possible that
+// the plugin inventory cache was setup by an older version of the CLI and is missing
+// the central_config.yaml file.  In such a case, the digest of the cache will still indicate
+// that the latest plugin inventory is present and the content of the cache will not be refreshed
+// until the plugin inventory data changes in the central repo itself.  In such a case, to be able
+// to benefit from the central configuration feature once the CLI is upgraded to >= 1.3.0
+// we need to fix the cache.  This initializer checks if the central_config.yaml file is
+// present in the cache and if not, it invalidates the cache.
+
+func init() {
+	globalinit.RegisterInitializer(triggerForInventoryCacheInvalidation, invalidateInventoryCache)
+}
+
+// triggerForInventoryCacheInvalidation returns true if the central_config.yaml file is missing
+// in the plugin inventory cache.
+func triggerForInventoryCacheInvalidation() bool {
+	sources, err := config.GetCLIDiscoverySources()
+	if err != nil {
+		// No discovery source
+		return false
+	}
+
+	for _, source := range sources {
+		centralConfigFile := filepath.Join(common.DefaultCacheDir, common.PluginInventoryDirName, source.OCI.Name, constants.CentralConfigFileName)
+
+		if _, err := os.Stat(centralConfigFile); os.IsNotExist(err) {
+			// As soon as we find a source that doesn't have a central_config.yaml file,
+			// we need to perform some initialization.
+			return true
+		}
+	}
+	// If we get here, then all sources have a central_config.yaml file
+	return false
+}
+
+// invalidateInventoryCache performs the required actions
+func invalidateInventoryCache(_ io.Writer) error {
+	sources, err := config.GetCLIDiscoverySources()
+	if err != nil {
+		// No discovery source
+		return nil
+	}
+
+	var errorList []error
+	for _, source := range sources {
+		centralConfigFile := filepath.Join(common.DefaultCacheDir, common.PluginInventoryDirName, source.OCI.Name, constants.CentralConfigFileName)
+
+		if _, err := os.Stat(centralConfigFile); os.IsNotExist(err) {
+			// This source doesn't have a central_config.yaml file,
+			// we need to invalidate its plugin inventory cache.
+			err = discovery.RefreshDiscoveryDatabaseForSource(source, discovery.WithForceInvalidation())
+			if err != nil {
+				errorList = append(errorList, err)
+			}
+		}
+	}
+	return kerrors.NewAggregate(errorList)
+}
--- a/pkg/centralconfig/central_config_cache_init.go
+++ b/pkg/centralconfig/central_config_cache_init.go
@@ -29,7 +29,7 @@ import (
 // present in the cache and if not, it invalidates the cache.
 
 func init() {
-	globalinit.RegisterInitializer(triggerForInventoryCacheInvalidation, invalidateInventoryCache)
+	globalinit.RegisterInitializer("Central Config Initializer", triggerForInventoryCacheInvalidation, invalidateInventoryCache)
 }
 
 // triggerForInventoryCacheInvalidation returns true if the central_config.yaml file is missing
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -345,8 +345,7 @@ func checkGlobalInit(cmd *cobra.Command) {
 
 		err := globalinit.PerformInitializations(outStream)
 		if err != nil {
-			fmt.Fprintln(outStream, "The initialization encountered an error but the CLI remains functional.")
-			log.V(6).Warningf("Error initializing CLI: %v", err)
+			log.Warningf("The initialization encountered the following error: %v", err)
 		}
 
 		fmt.Fprintln(outStream)
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -24,6 +24,7 @@ import (
 	cliconfig "github.com/vmware-tanzu/tanzu-cli/pkg/config"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/constants"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/discovery"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/globalinit"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/pluginmanager"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/pluginsupplier"
 	"github.com/vmware-tanzu/tanzu-cli/pkg/recommendedversion"
@@ -265,6 +266,13 @@ func newRootCmd() *cobra.Command {
 			// Sets the verbosity of the logger if TANZU_CLI_LOG_LEVEL is set
 			setLoggerVerbosity()
 
+			// Perform some global initialization of the CLI if necessary
+			// We do this as early as possible to make sure the CLI is ready for use
+			// for any other logic below.
+			if !shouldSkipGlobalInit(cmd) {
+				checkGlobalInit(cmd)
+			}
+
 			// Ensure mutual exclusion in current contexts just in case if any plugins with old
 			// plugin-runtime sets k8s context as current when tanzu context is already set as current
 			if err := utils.EnsureMutualExclusiveCurrentContexts(); err != nil {
@@ -325,6 +333,25 @@ func setLoggerVerbosity() {
 		if err == nil {
 			log.SetVerbosity(int32(logValue))
 		}
+	}
+}
+
+func checkGlobalInit(cmd *cobra.Command) {
+	if globalinit.InitializationRequired() {
+		outStream := cmd.OutOrStderr()
+
+		fmt.Fprintf(outStream, "Some initialization of the CLI is required.\n")
+		fmt.Fprintf(outStream, "Let's set things up for you.  This will just take a few seconds.\n\n")
+
+		err := globalinit.PerformInitializations(outStream)
+		if err != nil {
+			fmt.Fprintln(outStream, "The initialization encountered an error but the CLI remains functional.")
+			log.V(6).Warningf("Error initializing CLI: %v", err)
+		}
+
+		fmt.Fprintln(outStream)
+		fmt.Fprintln(outStream, "Initialization done!")
+		fmt.Fprintln(outStream, "==")
 	}
 }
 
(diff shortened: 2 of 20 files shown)


pkg/centralconfig/central_config_cache_init.go
```diff
@@ -0,0 +1,79 @@
...
+
+	for _, source := range sources {
+		centralConfigFile := filepath.Join(common.DefaultCacheDir, common.PluginInventoryDirName, source.OCI.Name, constants.CentralConfigFileName)
+
+		if _, err := os.Stat(centralConfigFile); os.IsNotExist(err) {
+			// As soon as we find a source that doesn't have a central_config.yaml file,
+			// we need to perform some initialization.
+			return true
+		}
+	}
```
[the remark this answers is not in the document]
marckhouzam: You are right.  There will only be one source: `default`.
However, if we ever start supporting more than one source, then we'd have to change this logic, so I figured it wasn't too hard to make the code future-proof.

Note that this logic does not include sources set using `TANZU_CLI_ADDITIONAL_PLUGIN_DISCOVERY_IMAGES_TEST_ONLY` since they are not in the config file.  Those sources are not important anyway for the central config since only the central config of the `default` source is used.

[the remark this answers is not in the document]
marckhouzam: But now that I write this I realize that the recommended version feature is not future-proof if we have multiple (official) plugin sources (in the future).
Do you think we should have an issue to track that?

pkg/globalinit/global_initializers.go
```diff
@@ -0,0 +1,66 @@
...
+		if i.triggerFunc() {
+			return true
+		}
+	}
+	return false
+}
+
+// PerformInitializations run each initializer which which the trigger function returns true.
+func PerformInitializations(outStream io.Writer) error {
+	if !InitializationRequired() {
```
vuil, line 52: nit: seems the cost of running all the triggers will have to be paid regardless, so why is there a need for this function, (and now end up paying the cost twice)?

pkg/globalinit/global_initializers.go
```diff
@@ -0,0 +1,66 @@
...
+
+type initializer struct {
+	// The trigger functions are kept separate from the initialization functions
+	// for a couple of reasons:
+	// 1. The trigger functions can be used first to determine if the initialization
+	//    should be performed.  If so, a global message can be printed to the user
+	//    before the initialization is performed.
+	// 2. Some trigger functions or initialization functions could be re-used by
+	//    different features for slightly different purposes.
+	triggerFunc        func() bool
```
vuil, line 24: like the separation of trigger and initialization functions, and good explanation for it.

pkg/globalinit/global_initializers.go
```diff
@@ -0,0 +1,66 @@
...
+package globalinit
+
+import (
+	"fmt"
+	"io"
+
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+type initializer struct {
```
vuil, line 16: a suggestion: how about supporting providing a name for each initializer registered? It can be useful for a few things:
1. Used in aggregated error list to provide some consistency in the output
2. In the future, should it happen that a particular initializer consistently fails for a CLI user (intentionally caused by the customer or otherwise), it is much easier to provide some means to turn off certain initializers if we can refer to them by name

pkg/globalinit/global_initializers.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+var (
+	initializers []initializer
+)
+
+// RegisterInitializer registers a new initializer with the global list of initializers.
+// The trigger function is used to determine if the initialization function should be run.
+// The initialization function is the function that will be run if the trigger function returns true.
+// The set of initializer triggers is checked whenever the CLI is run.
```
vuil, line 35: weak suggestion:
while no initializer should be written in a way that depends on other initializers, consider mentioning that the initializers will be run in the order that they are registered.

pkg/globalinit/global_initializers_test.go
```diff
@@ -0,0 +1,211 @@
...
+				func(w io.Writer) error { fmt.Fprintln(w, "1"); return nil },
+				func(w io.Writer) error { fmt.Fprintln(w, "2"); return nil },
+				func(w io.Writer) error { fmt.Fprintln(w, "3"); return nil },
+			},
+			expectError:   false,
+			expectedOut:   []string{"2"},
+			unexpectedOut: []string{"1", "3"},
+		},
+		{
+			test: "init throws error",
```
vuil, line 175: nit: I think one interesting case to include in multiple initFuncs, returning a mix of success and error (e.g. erro just on the second of 3). Demonstrates that error does not interrupt other initializers.

vuil (commented):
Good stuff!
Some minor comments on the framework part

test/e2e/cli_lifecycle/initialize_central_config_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/common"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/config"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/constants"
+	"github.com/vmware-tanzu/tanzu-cli/test/e2e/framework"
+)
+
+// These tests verify that when the central config file is missing from the cache,
+// a global initializer is triggered to invalidate the cache.
+var _ = framework.CLICoreDescribe("[Tests:E2E][Feature:Initialize-central-config-cache]", func() {
```
vuil, line 21: nice!

test/e2e/cli_lifecycle/initialize_central_config_test.go
```diff
@@ -0,0 +1,87 @@
...
+			Expect(errStream).ToNot(ContainSubstring(initErrorStr))
+
+			// Run the command again to see that the global initializer is not triggered
+			// since the cache was fixed
+			_, _, errStream, err = tf.PluginCmd.ListPlugins()
+			Expect(err).To(BeNil())
+			Expect(errStream).ToNot(ContainSubstring(initializationStr))
+			Expect(errStream).ToNot(ContainSubstring(initErrorStr))
+		})
+		It("initialization when config file is missing a second time", func() {
```
vuil, line 60: nit: test title and comments aside, this test looks identical to the previous one!
To prove the point that an initialized CLI seeing a deleted central config afterwards can still re-initialize correctly (which I gather is the purpose of this test), you might be better off just running the previous test in a loop of 2 iterations and combining the comments.

pkg/command/root.go
```diff
@@ -307,6 +315,25 @@ func setLoggerVerbosity() {
...
+func checkGlobalInit(cmd *cobra.Command) {
+	if globalinit.InitializationRequired() {
+		outStream := cmd.OutOrStderr()
+
+		fmt.Fprintf(outStream, "Some initialization of the CLI is required.\n")
+		fmt.Fprintf(outStream, "Let's set things up for you.  This will just take a few seconds.\n\n")
+
+		err := globalinit.PerformInitializations(outStream)
+		if err != nil {
+			fmt.Fprintln(outStream, "The initialization encountered an error but the CLI remains functional.")
```
vuil, line 327: I feel a little torn about this message. 
If we can be sure that is the case, it is definitely a very useful message to show to reassure the user, given the additional error messages that are also shown. 

It is just that I am not sure with each new initializer added that we wont include some initializations that could lead to rather bad consequences if they do not run successfully.

suggestion for the future:
If we want to do this right, we might need some framework during registration, or establish some convention in the returned initializer error message to help hint at the severity of a failure and tune our message accordingly.

If we elect to pass on above for now,  another option is to be a little more conservative in the tone, or do nothing.

vuil (approved):
A couple more nits and suggestions.

pkg/globalinit/global_initializers.go
marckhouzam: You mean, why first call `if !InitializationRequired()` when we will be checking each trigger below?
Good point.  I was to protect against a coding error of calling `PerformInitializations` without first having checked `InitializationRequired()` but you make me realize there isn't much value but here is a performance impact.

I'll remove it.

test/e2e/cli_lifecycle/initialize_central_config_test.go
marckhouzam: Good eye 👍 , will do

pkg/command/root.go
marckhouzam: I see your concern.
I did this because I was concerned about the different approaches possible:
1. silently ignore the error (with a log that only prints with increased logging) -> as you point this could leave the CLI in a bad state with no hint of why
2. completely fail execution -> this could render the CLI unusable because of a problem with an initializer
3. print the error but continue -> this may worry the user enough that they won't feel confident continuing to use the CLI

I felt 3 was the least problematic of the three choices, and to avoid the user feeling like the CLI could no longer be used, I printed "the CLI remains functional" even though that may not be the case.  I figure it is better to have the user continue to use a broken CLI and see other issues confirming the problem, than to stop using a properly functioning CLI.

> If we want to do this right, we might need some framework during registration, or establish some convention in the returned initializer error message to help hint at the severity of a failure and tune our message accordingly.

Maybe the initializer should simply not return an error if the error can be safely ignored?  Instead it could print a log itself.  Then, in `root.go` we would treat errors more seriously; still with just a printout though, not to debilitate the CLI.

marckhouzam: > Maybe the initializer should simply not return an error if the error can be safely ignored? Instead it could print a log itself. Then, in root.go we would treat errors more seriously; still with just a printout though, not to debilitate the CLI.

I will use this approach and print a warning if an error is returned.  For example:\
```
Some initialization of the CLI is required.
Let's set things up for you.  This will just take a few seconds.

[i] Refreshing plugin inventory cache for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.
[!] The initialization encountered the following error: failed to invalidate the plugin inventory cache for source

Initialization done!
==
```

vuil: The updated approach looks reasonable as well.

vuil (approved):
Thanks for making prompt updates. The remaining issues (e.g. regarding ordering of the initializers) are not blockers at the moment, and can be addressed in a followup after some discussion

pkg/globalinit/global_initializers.go
marckhouzam: We will keep this point for later

PR #725, opened 2024-04-05
Add support for individual plugin download with `tanzu plugin download-bundle` command

anujc25:

- Add support for individual plugin download with `tanzu plugin download-bundle` command and Add E2E tests

- This PR adds a new flag `--plugin` as part of `tanzu plugin download-bundle` command which takes pluginID as value.
- Based on the specified pluginID, matching plugins will be downloaded as a bundle that can be transferred and uploaded to the airgapped repository. Please see documentation update on how to use `--plugin` flag.

Fixes #

```sh
~ $ tanzu plugin download-bundle --help
Download a plugin bundle to the local file system to be used when migrating plugins
to an internet-restricted environment. Please also see the "upload-bundle" command.

Usage:
  tanzu plugin download-bundle [flags]

Examples:

    # Download a plugin bundle for a specific group version from the default discovery source
    tanzu plugin download-bundle --group vmware-tkg/default:v1.0.0 --to-tar /tmp/plugin_bundle_vmware_tkg_default_v1.0.0.tar.gz

    # To download plugin bundle with a specific plugin from the default discovery source
    #     --plugin name                 : Downloads the latest available version of the plugin. (Returns an error if the specified plugin name is available across multiple targets)
    #     --plugin name:version         : Downloads the specified version of the plugin. (Returns an error if the specified plugin name is available across multiple targets)
    #     --plugin name@target:version  : Downloads the specified version of the plugin for the specified target.
    #     --plugin name@target          : Downloads the latest available version of the plugin for the specified target.
    tanzu plugin download-bundle --plugin cluster:v1.0.0 --to-tar /tmp/plugin_bundle_cluster.tar.gz

    # Download a plugin bundle with the entire plugin repository from a custom discovery source
    tanzu plugin download-bundle --image custom.registry.vmware.com/tkg/tanzu-plugins/plugin-inventory:latest --to-tar /tmp/plugin_bundle_complete.tar.gz

Flags:
      --group strings    only download the plugins specified in the plugin-group version (can specify multiple)
  -h, --help             help for download-bundle
      --image string     URI of the plugin discovery image providing the plugins (default "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest")
      --plugin strings   only download plugins matching specified pluginID. Format: name/name:version/name@target:version (can specify multiple)
      --to-tar string    local tar file path to store the plugin images
```

- Added E2E tests

### Release note

```release-note
Add support for individual plugin download with `tanzu plugin download-bundle` command
```

### DIFF
--- a/docs/quickstart/install.md
+++ b/docs/quickstart/install.md
@@ -447,13 +447,16 @@ If you do not specify a group's version, the latest version available for the gr
 tanzu plugin download-bundle --group vmware-tkg/default --to-tar /tmp/plugin_bundle_tkg_latest.tar.gz
 ```
 
-If you want to download a specific plugin, the `--plugin` flag can be used. Below are the support format for `--plugin` flag:
+If you want to download a specific plugin, the `--plugin` flag can be used. Using
+the `--plugin` flag is for cases where a plugin is not part of a plugin group,
+but that normally using `--group` is the recommended approach.
+Below are the supported formats for the `--plugin` flag:
 
 ```sh
---plugin name                 : Downloads all available versions of the plugin for all matching targets.
---plugin name:version         : Downloads specified version of the plugin for all matching targets. Use 'latest' as version for latest available version
---plugin name@target:version  : Downloads specified version of the plugin for the specified target. Use 'latest' as version for latest available version
---plugin name@target          : Downloads all available versions of the plugin for the specified target.
+--plugin name                 : Downloads the latest available version of the plugin for all matching targets.
+--plugin name:version         : Downloads the specified version of the plugin for all matching targets. Use 'latest' as version for latest available version
+--plugin name@target:version  : Downloads the specified version of the plugin for the specified target. Use 'latest' as version for latest available version
+--plugin name@target          : Downloads the latest available version of the plugin for the specified target.
 ```
 
 ```sh
--- a/docs/quickstart/install.md
+++ b/docs/quickstart/install.md
@@ -414,10 +414,10 @@ resulting images without authentication.
 To download plugins you will use the `tanzu plugin download-bundle` command and
 specify the different plugin groups or plugins relevant to your environment.
 
-This command will download the plugin bundle containing specified plugins and
-the plugin versions specified in the plugin group definition. The bundle will
-also include the plugin group definition itself if specified, so that it can be
-used for plugin installation by users.
+This command will download the plugin bundle containing the specified plugin
+versions as well as the plugin versions specified in the plugin group definition.
+The bundle will also include the plugin group definition itself if specified,
+so that it can be used for plugin installation by users.
 
 Note that the latest version of the `vmware-tanzucli/essentials` plugin group
 and the plugin versions it contains will automatically be included in any plugin
@@ -449,21 +449,21 @@ tanzu plugin download-bundle --group vmware-tkg/default --to-tar /tmp/plugin_bun
 
 If you want to download a specific plugin, the `--plugin` flag can be used. Using
 the `--plugin` flag is for cases where a plugin is not part of a plugin group,
-but that normally using `--group` is the recommended approach.
+however, when possible, using `--group` is the recommended approach.
 Below are the supported formats for the `--plugin` flag:
 
 ```sh
---plugin name                 : Downloads the latest available version of the plugin for all matching targets.
---plugin name:version         : Downloads the specified version of the plugin for all matching targets. Use 'latest' as version for latest available version
---plugin name@target:version  : Downloads the specified version of the plugin for the specified target. Use 'latest' as version for latest available version
+--plugin name                 : Downloads the latest available version of the plugin. (Returns an error if the specified plugin name is available across multiple targets)
+--plugin name:version         : Downloads the specified version of the plugin. (Returns an error if the specified plugin name is available across multiple targets)
+--plugin name@target:version  : Downloads the specified version of the plugin for the specified target.
 --plugin name@target          : Downloads the latest available version of the plugin for the specified target.
 ```
 
 ```sh
-# To download plugin bundle with all available versions 'cluster' plugin across all targets
+# To download plugin bundle with the latest available version of the 'cluster' plugin. (Returns an error if the specified plugin name is available across multiple targets)
 tanzu plugin download-bundle --plugin cluster --to-tar /tmp/plugin_bundle_cluster.tar.gz
 
-# To download plugin bundle with all available versions 'cluster' plugin for `kubernetes` target
+# To download plugin bundle with the latest available version of the 'cluster' plugin for `kubernetes` target
 tanzu plugin download-bundle --plugin cluster@kubernetes --to-tar /tmp/plugin_bundle_cluster.tar.gz
 
 # To download plugin bundle with v1.0.0 version of 'cluster' plugin for `kubernetes` target
@@ -473,7 +473,7 @@ tanzu plugin download-bundle --plugin cluster@kubernetes:v1.0.0 --to-tar /tmp/pl
 tanzu plugin download-bundle --plugin cluster@kubernetes:latest --to-tar /tmp/plugin_bundle_cluster.tar.gz
 ```
 
-Using `--group` and `--plugin` flag together are also supported and if specified the union of all the
+Using the `--group` and `--plugin` flags together is also supported.  In such a case the union of all the
 plugins and plugin-groups will be downloaded.
 
 To migrate plugins from a specific plugin repository and not use the default
--- a/docs/quickstart/install.md
+++ b/docs/quickstart/install.md
@@ -412,11 +412,12 @@ resulting images without authentication.
 #### Downloading plugin bundle
 
 To download plugins you will use the `tanzu plugin download-bundle` command and
-specify the different plugin groups relevant to your environment.
+specify the different plugin groups or plugins relevant to your environment.
 
-This command will download the plugin bundle containing the plugin versions specified
-in the plugin group definition.  The bundle will also include the plugin group
-definition itself, so that it can be used for plugin installation by users.
+This command will download the plugin bundle containing specified plugins and
+the plugin versions specified in the plugin group definition. The bundle will
+also include the plugin group definition itself if specified, so that it can be
+used for plugin installation by users.
 
 Note that the latest version of the `vmware-tanzucli/essentials` plugin group
 and the plugin versions it contains will automatically be included in any plugin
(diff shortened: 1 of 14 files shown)


pkg/airgapped/plugin_bundle_download.go
```diff
@@ -173,6 +176,21 @@ func (o *DownloadPluginBundleOptions) getSelectedPluginInfo() ([]*plugininventor
...
+		pluginEntries, err := pi.GetPlugins(&plugininventory.PluginInventoryFilter{
+			Name:          pluginName,
+			Target:        configtypes.Target(pluginTarget),
+			Version:       pluginVersion,
+			IncludeHidden: true,
+		}) // Include the hidden plugins during plugin migration
+
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "unable to read plugins from database")
+		}
```
marckhouzam, line 190: I think we should check if 0 plugins were returned which would mean the value of `--plugin` is wrong.
For example, if the user typed `--plugin cluster:kubernetes` it won't download the plugin but will not print an error either.

marckhouzam: Could we add a unit test to check that the command fails if the plugin specified is not found?

marckhouzam (commented):
Nice work!  This is something we really need!
Some minor comments to discuss.

docs/quickstart/install.md
```diff
@@ -446,6 +447,35 @@ If you do not specify a group's version, the latest version available for the gr
 tanzu plugin download-bundle --group vmware-tkg/default --to-tar /tmp/plugin_bundle_tkg_latest.tar.gz
 ```
 
+If you want to download a specific plugin, the `--plugin` flag can be used. Using
+the `--plugin` flag is for cases where a plugin is not part of a plugin group,
+but that normally using `--group` is the recommended approach.
```
marckhouzam, line 452: but that normally using -> however, when possible, using

pkg/airgapped/plugin_bundle_download.go
```diff
@@ -180,6 +191,36 @@ func (o *DownloadPluginBundleOptions) getSelectedPluginInfo() ([]*plugininventor
 	return selectedPluginEntries, selectedPluginGroups, nil
 }
 
+func (o *DownloadPluginBundleOptions) getPluginFromPluginID(pluginID string, pi plugininventory.PluginInventory) ([]*plugininventory.PluginInventoryEntry, error) {
+	pluginName, pluginTarget, pluginVersion := utils.ParsePluginID(pluginID)
+	if pluginVersion == "" {
+		pluginVersion = "latest"
```
marckhouzam, line 197: Could use `cli.VersionLatest` instead of `"latest"`?

pkg/command/plugin_bundle.go
```diff
@@ -42,7 +43,14 @@ func newDownloadBundlePluginCmd() *cobra.Command {
 to an internet-restricted environment. Please also see the "upload-bundle" command.`,
 		Example: `
     # Download a plugin bundle for a specific group version from the default discovery source
-    tanzu plugin download-bundle --to-tar /tmp/plugin_bundle_vmware_tkg_default_v1.0.0.tar.gz --group vmware-tkg/default:v1.0.0
+    tanzu plugin download-bundle --group vmware-tkg/default:v1.0.0 --to-tar /tmp/plugin_bundle_vmware_tkg_default_v1.0.0.tar.gz
+
+    # To download plugin bundle with specific plugin from the default discovery source
+    #     --plugin name                 : Downloads the latest available version of the plugin for all matching targets.
```
marckhouzam, line 49: We error out now if there are multiple targets, right?

pkg/command/plugin_bundle.go
```diff
@@ -42,7 +43,14 @@ func newDownloadBundlePluginCmd() *cobra.Command {
...
+    #     --plugin name:version         : Downloads the specified version of the plugin for all matching targets. Use 'latest' as version for latest available version
```
marckhouzam, line 50: Same here about targets

docs/quickstart/install.md
```diff
@@ -446,6 +447,35 @@ If you do not specify a group's version, the latest version available for the gr
...
+Below are the supported formats for the `--plugin` flag:
+
+```sh
+--plugin name                 : Downloads the latest available version of the plugin for all matching targets.
+--plugin name:version         : Downloads the specified version of the plugin for all matching targets. Use 'latest' as version for latest available version
```
marckhouzam, line 457: For the two above, we no longer download all targets, right?

docs/quickstart/install.md
```diff
@@ -446,6 +447,35 @@ If you do not specify a group's version, the latest version available for the gr
...
+--plugin name@target:version  : Downloads the specified version of the plugin for the specified target. Use 'latest' as version for latest available version
+--plugin name@target          : Downloads the latest available version of the plugin for the specified target.
+```
+
+```sh
+# To download plugin bundle with all available versions 'cluster' plugin across all targets
```
marckhouzam, line 463: no longer "all available versions"

docs/quickstart/install.md
```diff
@@ -446,6 +447,35 @@ If you do not specify a group's version, the latest version available for the gr
...
+tanzu plugin download-bundle --plugin cluster --to-tar /tmp/plugin_bundle_cluster.tar.gz
+
+# To download plugin bundle with all available versions 'cluster' plugin for `kubernetes` target
```
marckhouzam, line 466: same

docs/quickstart/install.md
```diff
@@ -446,6 +447,35 @@ If you do not specify a group's version, the latest version available for the gr
...
+tanzu plugin download-bundle --plugin cluster@kubernetes --to-tar /tmp/plugin_bundle_cluster.tar.gz
+
+# To download plugin bundle with v1.0.0 version of 'cluster' plugin for `kubernetes` target
+tanzu plugin download-bundle --plugin cluster@kubernetes:v1.0.0 --to-tar /tmp/plugin_bundle_cluster.tar.gz
+
+# To download plugin bundle with latest available version of 'cluster' plugin for `kubernetes` target
+tanzu plugin download-bundle --plugin cluster@kubernetes:latest --to-tar /tmp/plugin_bundle_cluster.tar.gz
+```
+
+Using `--group` and `--plugin` flag together are also supported and if specified the union of all the
```
marckhouzam, line 476:
```suggestion
Using the `--group` and `--plugin` flags together is also supported.  In such a case the union of all the
```

marckhouzam (commented):
Using the short form of the target does not seems to work:
```
$ tz plugin download-bundle --image localhost:9876/tanzu-cli/plugins/central:small --dry-run --plugin cluster@k8s

[i] Getting selected plugin information...
[i] will be downloading the one plugin from group: vmware-tanzucli/essentials:v9.9.9
[x] : error while getting selected plugin and plugin group information: no plugins found for pluginID "cluster@k8s"
```

pkg/command/plugin_bundle.go
```diff
@@ -249,3 +261,110 @@ func completeGroupsAndVersionForBundleDownload(_ *cobra.Command, _ []string, toC
...
+		// Since more recent versions are more likely to be
+		// useful, we tell the shell to preserve the order
+		// using cobra.ShellCompDirectiveKeepOrder
+		pluginID := toComplete[:idx]
+		versionToComplete := toComplete[idx+1:]
+
+		return completeVersionInPluginIDForDownloadBundle(PluginEntries, pluginID, versionToComplete), cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveKeepOrder
+	}
+
+	// Complete plugin group names
```
marckhouzam, line 351: Please fix comment

pkg/command/plugin_bundle.go
```diff
@@ -249,3 +261,110 @@ func completeGroupsAndVersionForBundleDownload(_ *cobra.Command, _ []string, toC
...
+	var comps []string
+	for _, p := range PluginEntries {
+		id := plugininventory.PluginToID(p)
+		if strings.HasPrefix(id, toComplete) {
```
marckhouzam, line 355: You might have wondered why we don't check for `HasPrefix()` when dealing with plugin groups.
I should have put a comment to explain.  This is some advanced shell completion stuff 😄 

If we filter on prefix, we are making the assumption that the only completions we should show _must_ start with what the user has typed.  This turns out to be wrong because `zsh` and `fish` are smarter than that: they can match what the user typed on something else than the prefix of the completion, mostly it is matching on substring (fuzzy matching).  Fish can even match on substrings within the _descriptions_ of the completions.

For example you can try doing this with fish:
```
tanzu completion fish | source
# Set the variable so it exists in the config file
tanzu config set env.TANZU_CLI_ADDITIONAL_PLUGIN_DISCOVERY_IMAGES_TEST_ONLY ''

# The below will match the variable name on the _substring_ ADD
tanzu config set ADD<TAB>
```
But for that to work, we need to return all the completions that _could_ match.  So, it is better not to use `HasPrefix()` but instead to return all valid completions.  The shell will then filter appropriately, doing prefix first, and if nothing matches, then trying substring.

P.S. I'm not able to get this concept to work for this PR (modified), I'll have to investigate later

pkg/command/plugin_bundle.go
```diff
@@ -249,3 +261,110 @@ func completeGroupsAndVersionForBundleDownload(_ *cobra.Command, _ []string, toC
...
+			comps = append(comps, id)
```
marckhouzam, line 356: Could you also add the description of the plugin to make the completion more informative?
`comps = append(comps, fmt.Sprintf("%s\t%s", plugininventory.PluginToID(p), p.Description))`

```
$ tz plugin download-bundle --dry-run --plugin cluster<TAB>
cluster@kubernetes                                     -- Kubernetes cluster operations
cluster@mission-control                                -- A TMC managed Kubernetes cluster
cluster@operations                                     -- Plugin for operating clusters in the organization
clustergroup@mission-control  clustergroup@operations  -- A group of Kubernetes clusters
```

pkg/command/plugin_bundle.go
```diff
@@ -249,3 +261,110 @@ func completeGroupsAndVersionForBundleDownload(_ *cobra.Command, _ []string, toC
...
+	}
+
+	if len(comps) == 0 {
+		return cobra.AppendActiveHelp(nil, fmt.Sprintf("There are no plugins matching: '%s'", toComplete)), cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveKeepOrder
+	}
+
+	// Sort to allow for testing
+	sort.Strings(comps)
+
+	// Don't add a space after the group name so the uer can add a : if
```
marckhouzam, line 367: Please fix the comment

pkg/command/plugin_bundle.go
```diff
@@ -249,3 +261,110 @@ func completeGroupsAndVersionForBundleDownload(_ *cobra.Command, _ []string, toC
...
+	var comps []string
+	for _, p := range PluginEntries {
+		id := plugininventory.PluginToID(p)
+		if strings.HasPrefix(id, toComplete) {
+			comps = append(comps, id)
+		}
+	}
+
+	if len(comps) == 0 {
+		return cobra.AppendActiveHelp(nil, fmt.Sprintf("There are no plugins matching: '%s'", toComplete)), cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveKeepOrder
```
marckhouzam, line 361: Although this is a good idea, it will no longer work if we stop using `HasPrefix()` because our go code doesn't know in advance if the shell will match on anything or not.

This is a limitation of supporting fuzzy matching.

marckhouzam (commented):
Thank you for spending the time to do the shell completion.
It can be pretty complicated but I believe it makes the use of the new flag much easier for users!

I'll be honest that I'm not loving the fact that we always complete the target automatically.  I would have liked that we only complete the target if needed, but with fuzzy matching it may not be easy.  I'm just mentioning it but I feel it is ok to do it like this for now.

pkg/command/plugin_bundle.go
anujc25: Thanks for the explanation. I have updated the PR to not use `HasPrefix()`.

pkg/command/plugin_bundle.go
```diff
@@ -249,3 +269,87 @@ func completeGroupsAndVersionForBundleDownload(_ *cobra.Command, _ []string, toC
...
+	// Complete pluginIDs
+	var comps []string
+	for _, p := range pluginEntries {
+		comps = append(comps, fmt.Sprintf("%s\t%s", plugininventory.PluginToID(p), p.Description))
+	}
+
+	// Sort to allow for testing
+	sort.Strings(comps)
+
+	// Don't add a space after the pluginID so the uer can add a : if
```
marckhouzam, line 352: uer -> user

marckhouzam (approved):
I love it!

PR #726, opened 2024-04-05
Add support to address the multiple callback requests during redirect by browser during login

prkalle:
This PR adds support to address the multiple callback requests during redirect by browser during login

Fixes #

Was able to create a context (in normal scenario) successfully.
```
❯ ./bin/tanzu context create test1 --type tanzu --staging --endpoint https://test-project-syncer.stacks.bluesky.tmc-dev.cloud.vmware.com/
Log in by visiting this link:

    https://console-stg.cloud.vmware.com/csp/gateway/discovery?client_id=tanzu-cli-client-id&code_challenge=BbKJTKr1V_AcWyuuM2UaGpqEmzvwxaFDk77g0Hfa6WU&code_challenge_method=S256&orgId=fca1d60c-1f4c-4fd2-a5de-da4b9cc39bec&redirect_uri=http%3A%2F%2F127.0.0.1%3A50977%2Fcallback&response_type=code&state=bfdad08242cb5e3589d7cf0be3e5ca49

[ok] Successfully logged into 'TAP-SaaS-Staging' organization and created a tanzu context
[i] Checking for required plugins for context 'test1'...

```

### Release note

```release-note

```

### DIFF
--- a/pkg/auth/csp/tanzu.go
+++ b/pkg/auth/csp/tanzu.go
@@ -59,8 +59,10 @@ type cspLoginHandler struct {
 	token                 *oauth2.Token
 	refreshToken          string
 	orgID                 string
+	orgName               string
 	promptForValue        func(ctx context.Context, promptLabel string, out io.Writer) (string, error)
 	isTTY                 func(int) bool
+	callbackHandlerMutex  sync.Mutex
 }
 
 // LoginOption is an optional configuration for Login().
@@ -240,6 +242,20 @@ func (h *cspLoginHandler) runLocalListener(listener net.Listener) func() {
 }
 
 func (h *cspLoginHandler) callbackHandler(w http.ResponseWriter, r *http.Request) {
+	// Lock the mutex to ensure only one request can access/update shared state at a time
+	// Note: This should be remote corner case, but we have seen cases where Chrome browser redirects
+	// and make 2 back to back requests with the same URL before the local server is exited.
+	// In such scenario, the second request would be blocked till the prior request is finished
+	// and then gets unblocked to check if the prior request already acquired token, if so just
+	// return with empty message else let the request go through for token exchange and return
+	// the response which would fail anyway(context would be canceled for second request)
+	h.callbackHandlerMutex.Lock()
+	defer h.callbackHandlerMutex.Unlock()
+
+	// Check if token is already set by prior request
+	if h.token != nil {
+		return
+	}
 	// token exchange should be complete once this callback handler completes execution.
 	defer h.tokenExchangeComplete()
 	code := r.URL.Query().Get("code")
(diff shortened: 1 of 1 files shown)


pkg/auth/csp/tanzu.go
```diff
@@ -240,6 +242,21 @@ func (h *cspLoginHandler) runLocalListener(listener net.Listener) func() {
 }
 
 func (h *cspLoginHandler) callbackHandler(w http.ResponseWriter, r *http.Request) {
+	// Lock the mutex to ensure only one request can access/update shared state at a time
+	// Note: This should be remote corner case, but we have seen cases where Chrome browser redirects
+	// and make 2 back to back requests with the same URL before the local server is exited.
+	// In such scenario, the second request would be blocked till the prior request is finished
+	// and then gets unblocked to check if the prior request already acquired token, if so return
+	// the success message else let the request go through for token exchange and return
+	// the response which would fail anyway(context would be canceled for second request)
```
prkalle, line 251: This is the best effort to provide the success reponse if the prior request is success(for most scenarios).

marckhouzam: Does this approach mean the user will see two success messages?
If so, maybe the second time (`if h.token != nil`) we should not print anything?

pkg/auth/csp/tanzu.go
```diff
@@ -240,6 +242,21 @@ func (h *cspLoginHandler) runLocalListener(listener net.Listener) func() {
...
+	h.callbackHandlerMutex.Lock()
+	defer h.callbackHandlerMutex.Unlock()
+
+	// Check if token is already set by prior request
+	if h.token != nil {
+		printSuccessMessage(w, h.orgName)
+		return
+	}
 	// token exchange should be complete once this callback handler completes execution.
 	defer h.tokenExchangeComplete()
```
marckhouzam, line 261: In the case of two requests, could the value of `h.tokenExchangeComplete()` here be the one created by the second request? Since the locking happens after this value is set.
If so, is it a problem that we don't call the one from the first request?

marckhouzam (commented):
I'm not familiar with this logic, so I have a couple of minor questions.

pkg/auth/csp/tanzu.go
prkalle: Since I cannot reproduce this issue, I am hoping it will only print only one success message ( I am not sure though) because by the time the second message is done, the first request handler would have called `tokenExchangeComplete()` which would close the local server too (but there is possiblity that second response is written after this method call and server close). 

(While debugging on the system where we saw this issue(not reproducible now anymore), we tried second request checking if the first request is in progress and returning without printing anything(but not with mutex in place) , we saw the browser not showing any text message (the first reponse may not have reached the browser as the second request returned immediately with out printing anything and closed the server).  )

However with mutex inplace I am thinking not printing anything should also make sense (but unfortunately I can't reproduce it and test this theory).  I will update the PR with not printing anything for now, we can revisit this again if we can reproduce it. what do you think?

pkg/auth/csp/tanzu.go
prkalle: Not sure I understand your question completely.
The first request acquires the lock, would get the token and call the `h.tokenExchangeComplete()` and then release the lock, so now the second request would get unblocked and then return immediately if the token is set, else it would try to get the token and should fail because the first request already cancelled the context by calling `h.tokenExchangeComplete()`.

pkg/auth/csp/tanzu.go
prkalle: updated the PR with the second request just returning without printing anything.

pkg/auth/csp/tanzu.go
marckhouzam: After discussing, @prkalle explained that the only function that is called more than once is `callbackHandler()`.  My comment here was because I thought code run earlier could be called twice.  Since this is not the case and that a second call to `callbackHandler()` immediately blocks waiting for the lock, then my concern is invalid.

marckhouzam (approved):
Seems correct to me.

PR #732, opened 2024-04-13
Enable tests that were not being run

marckhouzam:

I noticed that the ginko tests in `pkg/config` did not have a test suite and therefore were not being run.

Before this PR, notice the `[no tests to run]`.
The test coverage happens because the `init()` function is automatically run, but nothing is verified, so it is not real test coverage.
```
$ cd tanzu-cli
~tc
$ g co main
Switched to branch 'main'
Your branch is up to date with 'upstream/main'.
$ go test ./pkg/config -coverprofile coverage.txt
ok  	github.com/vmware-tanzu/tanzu-cli/pkg/config	0.634s	coverage: 12.2% of statements [no tests to run]
```

With this PR, notice the increased coverage and the fact that it no longer says `[no tests to run]`:
```
$ go test ./pkg/config -coverprofile coverage.txt
ok  	github.com/vmware-tanzu/tanzu-cli/pkg/config	1.010s	coverage: 36.7% of statements
```

Fixes # N/A

Let CI run

### Release note

```release-note
More unit tests
```

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -5,6 +5,7 @@ package config
 
 import (
 	"os"
+	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


anujc25 (approved):
Thanks for catching this. 👍

PR #735, opened 2024-04-16
Update signal handling logic

chandrareddyp:
This PR refactors the signal handling logic to improve efficiency and maintainability. Previously, the signal handling was part of the plugin installation logic, where a signal handler was registered with a channel to receive specific signals (e.g., Ctrl+C) and a separate goroutine was started to handle the signals for each plugin installation.

The changes in this PR streamline the signal handling process by moving the logic to a centralized location in the commands package (specifically, in root.go). Instead of registering the signal handler and starting a goroutine for each plugin installation, the signal handling is now registered and handled only once at a higher level.

By centralizing the signal handling, we eliminate the need to register and handle the signal handler for each sub-command.

Fixes #

Manual testing has been done for the following use cases:
1. Specific plugin installation: Terminate the process with Ctrl+C during installation and ensure it is properly terminated.
2. Plugin sync command: Terminate the process with Ctrl+C during installation and ensure it is properly terminated.
3. Plugin group installation: Terminate the process with Ctrl+C during installation and ensure it is properly terminated.
Here are screenshots of the testing:
image
image

### Release note

```release-note
The signal handling logic has been refactored and moved to a centralized location for reuse across all use cases.
```

### DIFF
--- a/pkg/pluginmanager/manager.go
+++ b/pkg/pluginmanager/manager.go
@@ -10,10 +10,8 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"os/signal"
 	"path/filepath"
 	"strings"
-	"syscall"
 
 	"github.com/Masterminds/semver"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/pluginmanager/manager.go
```diff
@@ -788,6 +764,7 @@ func installOrUpgradePlugin(p *discovery.Discovered, version string, installTest
 			spinner.SetFinalText(errMsg, log.LogTypeERROR)
 		}
 	}
+	spinner.SetFinalText(installedMsg, log.LogTypeINFO)
```
anujc25, line 767: May be we can optimize even further here:
```
	pluginErr := verifyInstallAndInitializePlugin(plugin, p, version, installTestPlugin)
	if pluginErr == nil && spinner != nil {
        	spinner.SetFinalText(installedMsg, log.LogTypeINFO)
	}
```

chandrareddyp: Updated.

pkg/pluginmanager/manager.go
```diff
@@ -750,42 +748,22 @@ func installOrUpgradePlugin(p *discovery.Discovered, version string, installTest
...
 	} else {
 		log.Info(installingMsg)
 	}
 
 	pluginErr := verifyInstallAndInitializePlugin(plugin, p, version, installTestPlugin)
-	if pluginErr != nil {
-		if spinner != nil {
+	if spinner != nil {
+		if pluginErr != nil {
 			spinner.SetFinalText(errMsg, log.LogTypeERROR)
```
anujc25, line 764: Considering L755 already sets the error message as the final text, do we need to set it again here on error?
Will it be possible to simplify this with just as mentioned in the previous comment?

```
        pluginErr := verifyInstallAndInitializePlugin(plugin, p, version, installTestPlugin)
	if pluginErr == nil && spinner != nil {
        	spinner.SetFinalText(installedMsg, log.LogTypeINFO)
	}
```

chandrareddyp: good catch, thanks. Updated.

anujc25 (approved):
LGTM. Thanks for simplifying this.

PR #756, opened 2024-04-30
Bump `github.com/docker/docker` and `golang.org/x/net` dependencies

anujc25:

- Bump github.com/docker/docker from 24.0.7+incompatible to 24.0.9+incompatible
- Bump golang.org/x/net from 0.17.0 to 0.23.0

Combines #739, #740, #741, #710

### Release note

```release-note
None
```


marckhouzam (commented):
The `test/e2e/framework/go.sum` and `test/sample-plugin/go.sum` need to be tidied some more.

I suggest we update the main `Makefile` to do this automatically to avoid manually having to do this (and easily forgetting).

Basing ourselves on the old code from `tanzu-framework` we should add this towards the top:
```
GO_MODULES=$(shell find . -path "*/go.mod" | xargs -I _ dirname _)
```
then modifying the `gomod` target to become:
```
.PHONY: gomod
gomod: ## Update go module dependencies
	for i in $(GO_MODULES); do \
		echo "-- Tidying $$i --"; \
		pushd $${i}; \
		$(GO) mod tidy || exit 1; \
		popd; \
	done
```

marckhouzam (approved):
Nice!
LGTM

PR #762, opened 2024-05-06
Add placeholder tpe command group

vuil:

Shows adding of a top level command group. Group stays hidden until actual commands are added under it.

NOTE: in the future, central config could be used to provide out-of-band updates to available command groups.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -94,6 +94,7 @@ func NewRootCmd() (*cobra.Command, error) { //nolint: gocyclo,funlen
 		k8sCmd,
 		tmcCmd,
 		opsCmd,
+		tpeCmd,
 		// Note(TODO:prkalle): The below ceip-participation command(experimental) added may be removed in the next release,
 		//       If we decide to fold this functionality into existing 'tanzu telemetry' plugin
 		newCEIPParticipationCmd(),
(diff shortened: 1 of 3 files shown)


pkg/command/root.go
```diff
@@ -471,6 +472,18 @@ var opsCmd = &cobra.Command{
...
 
+var tpeCmd = &cobra.Command{
+	Use:     "platform-engineering",
+	Short:   "Commands that provide functionality for Tanzu Platform Engineering",
+	Aliases: []string{"tpe"},
+	Annotations: map[string]string{
+		"group": string(plugin.ExtraCmdGroup),
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		handleTargetHelp(cmd, args)
```
marckhouzam, line 483: Since we now have this great way to add a group of commands without having to add a new target, could you update the printout that users can see in `handleTargetHelp()` to avoid mentioning "target"?
```
$ tz platform-engineering
Note: No plugins are currently installed for the "platform-engineering" target.

Commands that provide functionality for Tanzu Platform Engineering
[...]
```
Maybe we can say:
`Note: No installed plugin provides a sub-command for the  "platform-engineering" command`

pkg/command/root.go
```diff
@@ -471,6 +472,18 @@ var opsCmd = &cobra.Command{
 	},
 }
 
+var tpeCmd = &cobra.Command{
```
marckhouzam, line 475: How about putting a comment explaining that this command is not a target but that plugins can provide sub-commands by using 
```
CommandMap: []plugin.CommandMapEntry{
   {DestinationCommandPath: "platform-engineering myplugin"},
},
```

marckhouzam (approved):
This is really cool!
Two nits but LGTM

pkg/command/root.go
vuil: Thanks, because it is a something that only gets invoked when the command is hidden, I wasn't too concerned.
That said, I had similar thoughts and was thinking of renaming the function and doing away with the last word (target) altogether since over time there should be little distinction between a target and a plain command group since that's how we want the former to be viewed. wdyt?

marckhouzam: Sounds good to me.

pkg/command/root.go
vuil: I pointed to the doc section that mentions this use case instead, hope that's is okay as well.

marckhouzam (approved):
LGTM
Thanks!

PR #764, opened 2024-05-08
Experimental: Get the Tanzu Hub Endpoint based on the Central Config Metadata

anujc25:
- Get the Tanzu Hub Endpoint based on the Central Config Metadata
- Adds following new configuration to the central config
```
cli.core.tanzu_hub_metadata:
  cspProductIdentifier: "TANZU-SAAS"
  cspDisplayName: "Tanzu Platform"
  endpointProduction: https://www.production.fake.vmware.com/hub
  endpointStaging: https://www.staging.fake.vmware.com/hub
  useCentralConfig: false
```

- If `useCentralConfig` is `false`, Tanzu CLI will use CSP to get the hub endpoint
  - While using CSP it will get the `cspProductIdentifier` and `cspDisplayName` from central config keys `cli.core.tanzu_platform_csp_product_identifier` and `cli.core.tanzu_platform_csp_display_name` respectively.

- If `useCentralConfig` is `true`, Tanzu CLI uses the central config key `endpointProduction` to get the endpoint for production and `endpointStaging` for staging. If the values are empty it will use default values specified within the CLI.

Fixes #

- To be added

### Release note

```release-note
Experimental: Get the Tanzu Hub Endpoint based on the Central Config Metadata
```


marckhouzam (approved):
Looks great.  This gives us much more flexibility.

I on the fence about the config format using 5 separate keys instead of a single key. I see that the code reads one value at a time to make decisions so I see why you used different keys, but I still wonder if it would be more readable to have a single key with a struct; then you'd read the whole struct at the start and use only the content that you need. 

I'm only suggesting this because it would make the `central_config.yaml` file more readable for a human, which is not something that is critical, so I'm not set on it.

I'll leave it up to you and other reviewers.

LGTM

pkg/auth/csp/tanzu.go
```diff
@@ -429,8 +439,40 @@ func GetOrgNameFromOrgID(orgID, accessToken, issuer string) (string, error) {
 	return org.Name, nil
 }
 
-// GetTanzuHubEndpointForTAP retrieves Tanzu Hub Endpoint For TAP SaaS through the CSP API
+// GetTanzuHubEndpointForTAP retrieves Tanzu Hub Endpoint For TAP SaaS through the CSP API or through Central Config as fallback
 func GetTanzuHubEndpointForTAP(orgID, accessToken string, useStagingIssuer bool) (string, error) {
```
vuil, line 443: preexisting, nit:
Is it reasonable to remove "For TAP SaaS" in command and rename function to just GetTanzuHubEndpoint, especially now that we have another means to provide the hub endpoints via CC?

pkg/auth/csp/tanzu.go
```diff
@@ -476,13 +518,37 @@ func GetTanzuHubEndpointForTAP(orgID, accessToken string, useStagingIssuer bool)
 		return "", errors.Wrap(err, "could not unmarshal the services response")
 	}
 
+	var productIdentifier, displayName string
+	_ = centralConfigReader.GetCentralConfigEntry(centralConfigTanzuPlatformCSPProductIdentifierKey, &productIdentifier)
```
vuil, line 522: So unlike prior to the change, central config is a requirement for getting hub endpoint now? 
Would it have been better if we can retain some default values for these, as the ultimate fallback?

anujc25: Yeah makes sense. I will update the PR accordingly.

anujc25: Updated

pkg/auth/csp/tanzu.go
```diff
@@ -476,13 +518,37 @@ func GetTanzuHubEndpointForTAP(orgID, accessToken string, useStagingIssuer bool)
...
+	_ = centralConfigReader.GetCentralConfigEntry(centralConfigTanzuPlatformCSPDisplayName, &displayName)
```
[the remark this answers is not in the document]
anujc25: Updated the logic so this might not be relevant anymore.

vuil (approved):
lgtm, thanks for the updates!

marckhouzam (approved):
Very nice
LGTM!

PR #773, opened 2024-05-28
Hide the init command

marckhouzam:

There was a time when the `tanzu init` command did a fair bit, but it is a no-op now and kept around for backward compatibility.  There is however, no need to keep it visible, as it also leads to some confusion with other commands like "tanzu app init".

This commit hides the command.

We don't deprecate it because it may become useful again in the future.

Fixes # N/A

```
# notice there is no `init` command shown in the `system` group
$ tz
The Tanzu CLI

Usage:
  tanzu [command]

Available command groups:

  Admin
[...]
  System
    completion              Output shell completion code
    config                  Configuration for the CLI
    context                 Configure and manage contexts for the Tanzu CLI
    login                   Login to Tanzu Application Platform
    plugin                  Manage CLI plugins
    version                 Version information
  Target
    kubernetes              Commands that interact with a Kubernetes endpoint
    operations              Commands that support Kubernetes operations for Tanzu Application Platform

Flags:
  -h, --help   help for tanzu

Use "tanzu [command] --help" for more information about a command.

# Shell completion no longer suggests `init`
$ tz ini<TAB>

# The init command is still available to avoid breaking scripts
$ tz init
[ok] successfully initialized CLI

$ tz init -h
Initialize the CLI

Usage:
  tanzu init [flags]

Flags:
  -h, --help   help for init
```

### Release note

```release-note
Hide the `tanzu init` command which is no longer performing any actual initialization.
```

### DIFF
--- a/pkg/command/init.go
+++ b/pkg/command/init.go
@@ -17,14 +17,15 @@ func init() {
 }
 
 var initCmd = &cobra.Command{
-	Use:   "init",
-	Short: "Initialize the CLI",
+	Use:    "init",
+	Hidden: true,
+	Short:  "Initialize the CLI",
 	Annotations: map[string]string{
 		"group": string(plugin.SystemCmdGroup),
 	},
 	SilenceErrors:     true,
 	ValidArgsFunction: noMoreCompletions,
-	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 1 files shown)


prkalle (approved):
LGTM. Thanks!

PR #777, opened 2024-06-01
Respect command mapping for help from plugins

marckhouzam:

When remapping plugin commands, the `tanzu help` command on those remapped commands is not correct.

For example
```
# The trait commands is actually remapped from "space trait"
# Notice that we are getting the help from "space" instead of "trait"
$ tz help trait
Tanzu space lifecycle management

Usage:
  tanzu space [command]

Aliases:
  space, spaces, sp

Available Commands:
  create              Create a Space with specified configuration
  delete              Delete a space
[...]

# The deploy command is remapped from "resource deploy"
# Notice that we are getting the help from "resource" instead of "deploy"
$ tz help deploy
manage resources in a Kubernetes cluster

Usage:

Flags:
  -h, --help   help for resource
```

Fixes # N/A

After the PR:

```
$ tz help trait
Trait lifecycle management

Usage:
  tanzu trait [command]

Aliases:
  trait, traits, tr

Available Commands:
  create      Create a trait with specified configuration
  delete      Delete a trait
  generate    Generate a trait manifest with specified configuration.
  get         Get details from a trait.
  list        List the traits in the current workspace.

Flags:
  -h, --help   help for trait

Global Flags:
      --no-color       deactivate color, bold, animations, and emoji output
      --project name   name of the project to use when creating resources, or switching spaces (default is project name from current-context defined by Application Engine)

Use "tanzu trait [command] --help" for more information about a command.

$ tz help deploy
Deploy resources on the cluster

Usage:
  tanzu deploy [flags]

Flags:
      --diff                Show diff
      --from-build string   Set build output directory (format: /tmp/foo)
  -h, --help                help for deploy
      --only strings        Set files or directories to patch, ignoring tanzu.yml (format: file1.yml, dirname/) (can repeat)
      --patch               Patch resources
  -y, --yes                 Confirm all prompts with yes
```

Notice that the invocation form is respected:
```
# Calling "space trait" instead of "trait" is reflected in the help
$ tz help space trait
Trait lifecycle management

Usage:
  tanzu space trait [command]
[...]

# Aliases also work
$ tz help tr
Trait lifecycle management

Usage:
  tanzu trait [command]

Aliases:
  trait, traits, tr

[...]

$ tz help space tr
Trait lifecycle management

Usage:
  tanzu space trait [command]

Aliases:
  trait, traits, tr
[...]
```

### Release note

```release-note
Respect command mapping for the help command for plugins
```

### DIFF
--- a/pkg/cli/plugin_cmd.go
+++ b/pkg/cli/plugin_cmd.go
@@ -164,6 +164,7 @@ func getCmdForPluginEx(p *PluginInfo, cmdName string, mapEntry *plugin.CommandMa
 
 		runner := NewRunner(p.Name, p.InstallationPath, completion)
 		ctx := context.Background()
+		setupPluginEnv(srcHierarchy, dstHierarchy)
 		output, _, err := runner.RunOutput(ctx)
 		if err != nil {
 			return nil, cobra.ShellCompDirectiveError
@@ -198,10 +199,14 @@ func getCmdForPluginEx(p *PluginInfo, cmdName string, mapEntry *plugin.CommandMa
 		// calls such as "tanzu help cluster list", we need to do some argument
 		// parsing ourselves and modify what gets passed along to the plugin.
 		helpArgs := getHelpArguments()
+		if len(srcHierarchy) > 0 {
+			helpArgs = append(srcHierarchy, helpArgs...)
+		}
 
 		// Pass this new command in to our plugin to have it handle help output
 		runner := NewRunner(p.Name, p.InstallationPath, helpArgs)
 		ctx := context.Background()
+		setupPluginEnv(srcHierarchy, dstHierarchy)
 		err := runner.Run(ctx)
 		if err != nil {
 			log.Errorf("Help output for '%s' is not available.", c.Name())
--- a/pkg/command/root_test.go
+++ b/pkg/command/root_test.go
@@ -1381,7 +1381,7 @@ func TestCommandRemapping(t *testing.T) {
 					name:   "dummy2",
 					target: configtypes.TargetK8s,
 					commandMap: []plugin.CommandMapEntry{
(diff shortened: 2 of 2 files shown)


pkg/cli/plugin_cmd.go
```diff
@@ -164,6 +164,7 @@ func getCmdForPluginEx(p *PluginInfo, cmdName string, mapEntry *plugin.CommandMa
 
 		runner := NewRunner(p.Name, p.InstallationPath, completion)
 		ctx := context.Background()
+		setupPluginEnv(srcHierarchy, dstHierarchy)
```
marckhouzam, line 167: This one does not have any impact at the moment since shell completion does not use any of those environment variables.  However, to be future-proof, I felt adding it was the right thing to do.

vuil (commented):
Nice catch, and fix!
I suggest we add a test case or two to exercise the `tanzu help` functionality if possible.

marckhouzam:
> Nice catch, and fix! I suggest we add a test case or two to exercise the `tanzu help` functionality if possible.

I've added a couple of tests that fail before this change and pass after.

pkg/command/root_test.go
```diff
@@ -1965,6 +1965,39 @@ func TestCommandRemapping(t *testing.T) {
...
+					commandMap: []plugin.CommandMapEntry{
+						{
+							SourceCommandPath:      "show-invoke-context",
+							DestinationCommandPath: "dummy",
+						},
+					},
+				},
+			},
+			args: []string{"help", "dummy"},
+			// Output for the show-invoke-context command which prints the passed arguments
```
vuil, line 1996: Neat :-)

vuil (approved):
lgtm, thanks!

PR #780, opened 2024-06-06
Update URI for packaging key

marckhouzam:

The `packages.vmware.com` is now available through `packages-prod.broadcom.com`.
This PR changes such references in case the `vmware.com` domain is taken down.

Fixes # N/A

I check on VPN and without VPN that the new URI was accessible and provided the same key.
I also installed the CLI with RPM and APT using the new URI to the key.

### Release note

```release-note
Update `packages.vmware.com` to `packages-prod.broadcom.com`
```


vuil (approved):
lgtm, thanks!

PR #781, opened 2024-06-07
Deprecate the kubernetes target command

marckhouzam:

Today plugins created with TargetK8s target type have their commands visible/invocable via both `tanzu <pluginname> ...` and `tanzu kubernetes|k8s <plugingname> ...`

The form without a target prefix is available only to kubernetes target for historical and compatibility reasons, but is the predominant invocation style. Since Tanzu Platform's plugin functionality are delivered through a mix of kubernetes-targeted and nonkubernetes-targeted plugins, encouraging this more compact form is important in promoting consistent and intuitive invocation styles. 

This PR deprecates the invocation of plugins using the `kubernetes` prefix.
Due to backward compatibility invoking `tanzu kubernetes ...` remains possible with this PR, but the top-level help no longer shows this target/command group. Furthermore the help for  `tanzu kubernetes` will print a deprecation message to stderr guiding the user to use the short form.

Take notice that the deprecation message will only be shown for `tanzu kubernetes` and not for `tanzu kubernetes <subcmd>`.  This is to avoid hiding all the sub-commands under the kubernetes target (a deprecated command is automatically hidden).

To remove the invocation using the `kubernetes` prefix from the help, I have opened: https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/192.

Fixes # N/A

```
# No kubernetes target shown in the help
$ tz
The Tanzu CLI

Usage:
  tanzu [command]

Available command groups:

  Admin
  Build
    app                     Applications on Tanzu Platform
    build                   Generate app image from source code
    services                Commands for working with services, classes and claims
  Manage
    availability-target     Availability-target lifecycle management
    deploy                  Deploy resources
    kubectl                 Provides the kubectl functionality kubectl functionality
    profile                 Profile lifecycle management
    project                 View, list and use Tanzu Projects.
    rbac                    Impose role based access restrictions on spaces
    resource                manage resources in a Kubernetes cluster
    space                   Tanzu space lifecycle management
    trait                   Trait lifecycle management
  Run
    kubetool                Invokes any binary after setting KUBECONFIG to match the tanzu CLI
    package                 tanzu package management
    secret                  Tanzu secret management
    telemetry               configure cluster-wide settings for vmware tanzu telemetry
  System
    completion              Output shell completion code
    config                  Configuration for the CLI
    context                 Configure and manage contexts for the Tanzu CLI
    login                   Login to Tanzu Platform for Kubernetes
    plugin                  Manage CLI plugins
    version                 Version information
  Target
    operations              Commands that support Kubernetes operations for Tanzu Platform for Kubernetes

Flags:
  -h, --help   help for tanzu

Use "tanzu [command] --help" for more information about a command.

# Kuberetes target still accessible but with a deprecation message
# and its sub commands are visiable
$ tz k8s -h
Command "kubernetes" is deprecated, you should invoke its sub-commands directly without the "kubernetes" prefix.
Commands that interact with a Kubernetes endpoint

Usage:
  tanzu kubernetes [command]

Aliases:
  kubernetes, k8s

Available command groups:

  Build
    services                Commands for working with services, classes and claims
  Run
    package                 tanzu package management
    secret                  Tanzu secret management

Flags:
  -h, --help   help for kubernetes

Use "tanzu kubernetes [command] --help" for more information about a command.

# K8s plugins can still be invoked with the target
$ tz k8s secret -h
Tanzu secret management

Usage:
  tanzu secret [command]
  tanzu kubernetes secret [command]

Available Commands:
  registry      Registry secret operations

Flags:
  -h, --help                help for secret
      --kubeconfig string   The path to the kubeconfig file, optional
      --verbose int32       Number for the log level verbosity(0-9)

Use "tanzu secret [command] --help" for more information about a command.
Use "tanzu kubernetes secret [command] --help" for more information about a command.
```

### Release note

```release-note
Deprecate the plugin invocation form `tanzu kubernetes <plugin>` in favour of the shorter and more consistent `tanzu <plugin>`
```


pkg/command/root_test.go
```diff
@@ -1300,21 +1306,22 @@ func TestCommandRemapping(t *testing.T) {
...
-					invokedAs: []string{"dummy"},
-					aliases:   []string{"dum"},
-				},
-			},
-			args:     []string{""},
-			expected: []string{"kubernetes"},
-		},
-		{
-			test: "mapping plugins when conditional on active contexts",
+		// Only works for the Kubernetes target which is now always hidden
```
vuil, line 1309: nit: not sure if you meant the phrasing in the comment. Perhaps you meant something like "only works for the operations target since kubernetes target is always hidden?"

pkg/command/root.go
```diff
@@ -300,6 +300,11 @@ func setupTargetPlugins() error {
 		if targetCmd, exists := mapTargetToCmd[plugins[i].Target]; exists {
 			cmd := cli.GetUnmappedCmdForPlugin(&plugins[i])
 			if cmd != nil {
+				if plugins[i].Target == configtypes.TargetK8s {
+					// Add a deprecation message which will be printed when the user invokes
+					// any of the plugins commands using the form `tanzu kubernetes <plugin> ...`
+					cmd.Deprecated = fmt.Sprintf(`when called with the "kubernetes" prefix. Please invoke %q directly without the "kubernetes" prefix.`, cmd.Name())
```
vuil, line 306: Because cobra is opinionated about how it wants to format the deprecation message, this ends up reading a little awkward. I would have preferred that the messages says that invoking a command with k8s prefix is deprecated, not that the command itself is. 

I don't think we can do much about it. But one option I am considering is not even showing this message at all...

pkg/command/root.go
```diff
@@ -440,6 +445,10 @@ var k8sCmd = &cobra.Command{
 	Use:     "kubernetes",
 	Short:   "Commands that interact with a Kubernetes endpoint",
 	Aliases: []string{"k8s"},
+	// We are moving away from the 'kubernetes' target.
+	// All commands under this target are accessible as sub-commands of the root command.
+	// For backwards compatibility, we are keeping the target but are hiding it.
+	Deprecated: `you should invoke its sub-commands directly without the "kubernetes" prefix.`,
```
vuil, line 451: > Note that to show the deprecation message for any command invoked using tanzu kubernetes ... I had to put the deprecation message on the plugin command itself. This has the (possibly good) side-effect of hiding all kubernetes plugin commands in the help of tanzu kubernetes -h and also turning off shell-completion in tanzu kubernetes <TAB>.

While I understand why the side-effect happens, it was still somewhat surprising as a user. I think we are considering at least one situation where a command group (apps, v1) _should_ be visible , so perhaps we should still explore just making the output the same as before.

Also:
The side-effect does not appear to apply to remapped commands

```
> tz kubernetes
Command "kubernetes" is deprecated, you should invoke its sub-commands directly without the "kubernetes" prefix.
Commands that interact with a Kubernetes endpoint

Usage:
  tanzu kubernetes [command]

Aliases:
  kubernetes, k8s

Available command groups:

  Manage
    cluster                 remapped commands

Flags:
  -h, --help   help for kubernetes

Use "tanzu kubernetes [command] --help" for more information about a command.

> tanzu kubernetes
Commands that interact with a Kubernetes endpoint

Usage:
  tanzu kubernetes [command]

Aliases:
  kubernetes, k8s

Available command groups:

  Manage
    cluster                 deeper commands
    sample                  sample plugin to test command mapping
  Run
    feature                 Operate on features and featuregates

Flags:
  -h, --help   help for kubernetes

Use "tanzu kubernetes [command] --help" for more information about a command.
```

vuil (commented):
Nice, and thanks for updating the markdowns as well. I have a couple of questions, mostly about whether we should bent the output of the deprecated invocations to suit our needs more.

pkg/command/root.go
marckhouzam: Yeah, this is not ideal and it annoys me that Cobra enforces a prefix instead of letting us override completely.  Something we should improve in Cobra.

Until then we could could remove this deprecation and simply hide the command.
To be honest, I had started with this deprecation before figuring out how to get a deprecation message on all sub-commands (which Cobra does not support, but that we can pull off since we only have a single command for a plugin).

What do you prefer?

pkg/command/root.go
marckhouzam: Let me look into that

pkg/command/root.go
vuil: > (which Cobra does not support, but that we can pull off since we only have a single command for a plugin).
I forgot to also mention, the prefixing + custom string, when, coupled with the fact that we only reference the command name for the plugin (and not the actual subcommand invoked), is what makes the message feel awkward. 

What are the downsides for just hiding instead of showing the deprecation message? I figure for folks navigating from top level "tanzu", they wouldn't have seen the `kubernetes` command anyway, so it is no biggie. It appears what we lose is the ability to flag existing/established use of this form (e.g in scripts) as deprecated.

marckhouzam: > It appears what we lose is the ability to flag existing/established use of this form (e.g in scripts) as deprecated.

I agree that is the disadvantage.  Originally I was just going to hide the `tanzu kubernetes` command and that was it, but the requirement asked for a deprecation message.  We could revisit the requirement 😄

vuil: So there was a couple of things in the comment and I may not have been clear. I was thinking of keeping the deprecation notice for `tanzu kubernetes` if there is a way to make it still work as before (show child commands under it)
I am debating on the merits of just letting `tanzu kubernetes xxx...` run without a deprecation message because the message looks awkward atm. We lose a bit not having the deprecation message, but we can always come back and add it later as well.

marckhouzam: Let's start with that then.  Only hide the `kubernetes` target and keep its sub commands like before.

pkg/command/root_test.go
vuil: sgtm

vuil (approved):
lgtm, thanks for the updates

PR #783, opened 2024-06-18
Entry-specific active context type check in command mapping

vuil:
Introduce CommandMapProcessor to facilitate processing a plugin's command map. Uses an instance of said processor that is aware of the active contexts so that each command map entry can be evaluated against them for applicability.

Fixes #

Updated unit tests.

Manual:
Created sample plugin containing CommandMapEntry with explicity RequireContextType, verify
that the mapping only takes effect when a context of matching type is active. 

### Release note

```release-note
Support entry-specific active context type check in command mapping
```


anujc25 (approved):
LGTM. Nice unit tests.

PR #786, opened 2024-06-20
Add flags validation check to error if `--kubeconfig` flag is specified while creating tanzu context type using `--type` flag

prkalle:
This PR adds validation check to error if `--kubeconfig` flag is specified while creating tanzu context type using `--type` flag

Fixes #

Tested creating the tanzu context by providing both `--type tanzu` and `--kubeconfig` flags, CLI throws error as expected.
```
❯ ./bin/tanzu context create testTAPSaaSBeta3Prod --type tanzu  --kubeconfig ~/junk/config.yaml
[x] : the '–-kubeconfig' flag is not applicable when creating a context of type 'tanzu'
```
Tested creating the tanzu context by providing both `--type tanzu` and `--kubecontext` flags, CLI throws error as expected.
```
❯ ./bin/tanzu context create  --type tanzu  --kubecontext test
[x] : the '–-kubecontext' flag is not applicable when creating a context of type 'tanzu
```

### Release note

```release-note

```

### DIFF
--- a/pkg/command/context.go
+++ b/pkg/command/context.go
@@ -253,7 +253,9 @@ func createCtx(cmd *cobra.Command, args []string) (err error) {
 		}
 		ctxName = args[0]
 	}
-
+	if err := validateContextCreateFlagValues(); err != nil {
+		return err
+	}
 	if !configtypes.IsValidContextType(contextTypeStr) {
 		return errors.New(invalidContextType)
 	}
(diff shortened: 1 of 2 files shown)


pkg/command/context.go
```diff
@@ -286,6 +288,13 @@ func createCtx(cmd *cobra.Command, args []string) (err error) {
 	return nil
 }
 
+func validateContextCreateFlagValues() error {
+	if contextTypeStr == "tanzu" && kubeConfig != "" {
```
marckhouzam, line 292: Nit: can we use `ContextTypeTanzu` instead of `"tanzu"`

prkalle: Thanks, updated and also added condition check for kubecontext and throw error if --kubecontext flag provided along with `-type tanzu` context

marckhouzam (commented):
LGTM
I don't think the corner case I mentioned is very important.  We can leave it as is.

marckhouzam (approved):

PR #789, opened 2024-06-28
Remove dead links from the documents

anujc25:

- Remove dead cla.vmware.com links

Fixes #

### Release note

```release-note
None
```


vuil (commented):
We have sort out after cleaning out the links if what is described still makes sense.

PR #798, opened 2024-08-09
Update TMC endpoint mapping rules for Tanzu context

anujc25:
- Update TMC endpoint mapping rules for Tanzu context

Fixes #

- Unit tests

### Release note

```release-note
Update TMC endpoint mapping rules for the Tanzu context
```

### DIFF
--- a/pkg/command/context.go
+++ b/pkg/command/context.go
@@ -12,7 +12,6 @@ import (
 	"net/url"
 	"os"
 	"reflect"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


pkg/command/context.go
```diff
@@ -2143,14 +2142,30 @@ func renderDynamicTable(slices interface{}, tableWriter component.OutputWriter,
...
 func mapTanzuEndpointToTMCEndpoint(tanzuEndpoint string) string {
-	// Define the regular expression pattern
-	apiPattern := regexp.MustCompile(`https://api\.tanzu(-\w*)?\.cloud\.vmware\.com`)
-	// Replace "api" with "tmc" in the input URL
-	tmcEndpoint := apiPattern.ReplaceAllString(tanzuEndpoint, "https://tmc.tanzu$1.cloud.vmware.com")
+	tmcEndpoint := ""
+	// Define the mapping rules
+	mappingRules := []struct {
+		prefix string
+		suffix string
```
vuil, line 2149: nit: it is technically not a suffix but an ops specific prefix.

pkg/command/context.go
```diff
@@ -2143,14 +2142,30 @@ func renderDynamicTable(slices interface{}, tableWriter component.OutputWriter,
...
+	}{
+		{"https://api.tanzu", "https://tmc.tanzu"},
```
vuil, line 2151: wondering: should we be concerned that we are replacing for not just .cloud.vmware.com/.tanzu.broad.com but _every_ other domain?

anujc25: Yeah. I wanted to keep things flexible as much as possible for now. So this is intentional. Do you think we need specific one?

pkg/command/context.go
```diff
@@ -2143,14 +2147,35 @@ func renderDynamicTable(slices interface{}, tableWriter component.OutputWriter,
...
+		tmcEndpointPrefix   string
+	}{
+		{"https://api.tanzu", "https://tmc.tanzu"},
+		{"https://ucp.platform", "https://ops.platform"},
+	}
+
+	// Iterate through the mapping rules
+	for _, rule := range mappingRules {
+		if strings.HasPrefix(tanzuEndpoint, rule.tanzuEndpointPrefix) {
+			// Replace the prefix with the suffix
```
vuil, line 2168: nit: comment needs update

vuil (approved):
nit about a comment update, but lgtm, thanks.

pkg/command/context.go
anujc25: We had an offline discussion and decided to keep this flexible as we have in this change.

vuil (approved):

PR #805, opened 2024-09-05
Fix double-encoding of ca cert data in kube context

vuil:

When a tanzu CLI context is created with a custom ca cert. The ca cert data was erroneously base64 encoded before being given to the json Marshaler, which base64 encodes all []byte fields anyway.
This leads to invalide ca cert data being stored in the kube context, and the follow errors when used:

```
Error: Unable to create http client unable to load root certificates: unable to parse bytes as PEM block
Error: Unable to create rest mapper. signk8s.io/dynamicrestmapper states httpClient must not be nil, consider using rest.HTTPClientFor(c) to create a client
```

Fixes #

Updated and ran unit tests.
Verified that after the fix that kube clients is able to use the custom ca cert data to validate calls to an endpoint with custom certs

### Release note

```release-note
Properly support the use of custom ca cert data for tanzu type CLI contexts.
```

### DIFF
--- a/pkg/auth/tanzu/kubeconfig.go
+++ b/pkg/auth/tanzu/kubeconfig.go
@@ -5,7 +5,6 @@
 package tanzu
 
 import (
-	"encoding/base64"
 	"encoding/json"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


marckhouzam (approved):
LGTM

PR #807, opened 2024-09-06
Fix context type detection logic and e2e tests for TMC context creation

vuil:

    Updated e2e tests to specify mission-control context type in tests
    involving such contexts.

    Also fixed context type detection logic so that forceCSP does not take
    precedence over explicit context type provided.

- Updated and ran unit tests.
- Ran e2e tests.
- Manually verify `tanzu context create --type mission-control --force-csp` creates a mission-control type context

### Release note

```release-note
```

### DIFF
--- a/pkg/command/context.go
+++ b/pkg/command/context.go
@@ -359,9 +359,11 @@ func createNewContext() (context *configtypes.Context, err error) {
 	var ctxCreationType ContextCreationType
 	contextType := getContextType()
 
-	if (contextType == configtypes.ContextTypeTanzu) || (endpoint != "" && isTanzuPlatformSaaSEndpoint(endpoint)) {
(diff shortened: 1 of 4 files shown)


anujc25 (approved):
LGTM. Thanks

marckhouzam:
The change looks good however it is not just an E2E fix right?  Could you update the title to reflect that?

IIUC there are now cases where the CLI can no longer automatically create a TMC context based on the endpoint alone and the user must explicitly specify the context type as TMC. Correct?

vuil:
> The change looks good however it is not just an E2E fix right? Could you update the title to reflect that?
> 
> IIUC there are now cases where the CLI can no longer automatically create a TMC context based on the endpoint alone and the user must explicitly specify the context type as TMC. Correct?

Thanks you are right, updated the title to match the description.

PR #808, opened 2024-09-12
Fix incorrect context name being created when using new tanzu platform endpoints

anujc25:

Fix incorrect context name being created when using new endpoints

Fixes #

**Before this change:**
```
$ tz login --staging --endpoint https://platform-dev.tanzu.broadcom.com

$ tz context list
  NAME                                      ISACTIVE  TYPE   PROJECT  SPACE
  tpsm-5a2f0150                             false     tanzu
```

**After this change:**
```
$ tz login --staging --endpoint https://platform-dev.tanzu.broadcom.com

$ tz context list
  NAME                                      ISACTIVE  TYPE   PROJECT  SPACE
  Falcons-staging-5a2f0150                  true      tanzu
```

### Release note

```release-note
Fix incorrect context name being created when using new tanzu platform endpoints
```

### DIFF
--- a/pkg/command/login_test.go
+++ b/pkg/command/login_test.go
@@ -11,61 +11,91 @@ import (
 
 func TestPrepareTanzuContextName(t *testing.T) {
 	testCases := []struct {
-		orgName   string
-		endpoint  string
-		isStaging bool
-		forceCSP  bool
-		expected  string
+		orgName       string
+		ucpEndpoint   string
+		endpoint      string
+		saasEndpoints []string
+		isStaging     bool
+		forceCSP      bool
+		expected      string
 	}{
 		// Test case for normal input with no staging environment and default endpoint.
 		{
-			orgName:   "MyOrg",
-			endpoint:  centralconfig.DefaultTanzuPlatformEndpoint,
-			isStaging: false,
-			expected:  "MyOrg",
+			orgName:       "MyOrg",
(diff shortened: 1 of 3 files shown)


marckhouzam (approved):
I actually hit this problem yesterday while debugging and it was pretty annoying (and wrong).
Thanks for fixing!

In the PR description did you mean to refer to #801 instead of #808 (which is the current PR 😄 )

pkg/command/login_test.go
```diff
@@ -11,61 +11,91 @@ import (
...
-			forceCSP:  true,
+			orgName:       "MyOrg",
+			ucpEndpoint:   "https://ucp.custom-endpoint.com",
+			endpoint:      "https://custom-endpoint.com",
+			saasEndpoints: []string{centralconfig.DefaultTanzuPlatformEndpoint},
+			isStaging:     true,
+			expected:      "MyOrg-staging-70217fc3",
+			forceCSP:      true,
+		},
+		// Test case for normal input with new ucpEndpoint and platform endpoints which is actually a SaaS endpoints
```
vuil, line 60: nit: which is actually a SaaS endpoints -> which are actually SaaS endpoints. Same for L68

vuil (approved):
Thanks for the fix and updating the tests to clarify the behavior.
Just a small nit on the comments.

pkg/command/login_test.go
anujc25: Thanks. Updated.

PR #812, opened 2024-09-13
Fix missing org name in browser login

marckhouzam:

When doing an interactive login through a browser, the Tanzu Platform org name was no longer showing.
This PR fixes it.

Fixes # N/A

Before the PR, run `tanzu login` built from `main` and notice that the login succeeds but that the browser does not show the org name in the success message.

Run `tanzu login` using tanzu CLI v1.4.1 and notice that the org name is shown in the browser.

Use this PR and run `tanzu login` and notice the org name is now shown again.

### Release note

```release-note

```


prkalle (approved):
LGTM. Thanks for fixing it!

vuil (approved):
thanks for the fix, lgtm

PR #819, opened 2024-10-03
Support login to UAA using API token

vuil:
### What this PR does / why we need it

Enables login via a previously obtained API token to a UAA based endpoint.
Also: fixed a bug where the expiration time on a token refresh is incorrect set to too close to current time.

Note : adding more tests and waiting for the alternative client id to be determined.

### Which issue(s) this PR fixes

Fixes #

### Describe testing done for PR

Obtain refresh token from CLI context
TANZU_API_TOKEN=_obtainedtoken_  tanzu login --endpoint uaa-based-endpoint

Updated and ran unit tests.

### Release note

```release-note
Token obtained from the "api-token create" command can be used as value to environment variable TANZU_API_TOKEN to connect to uaa-based endpoint.
```

PR #820, opened 2024-10-03
Add support to generate API token with `tanzu api-token create`

anujc25:
- Add support to generate API token with the `tanzu api-token create` command
- Add unit tests

Sample output:

```
> tz api-token create 
Log in by visiting this link: https://wdc-10-206-98-55.eng.vmware.com/auth/oauth/authorize?client_id=tp_cli_app&code_challenge=XfCGpFSLGz5x_qYhXfsjYkXpfF6865RUh32wVd20-Ys&code_challe pe=code&state=1f5e2644c6e07bf7ff00cb5064bfabbe 

===

API Token Generation Successful! Your generated API token is: cle82767a13748debd53e33b2d46d92b-r

For non-interactive login use the API token as follows: TANZU_API_TOKEN=cle82767a13748debd53e33b2d46d92b-r tanzu login --endpoint https://wdc-10-206-98-55.eng.vmware.com

Please copy and save your token securely. Note that you will need to regenerate a new token before expiration time and login again to continue using the CLI.
```

Fixes #

- Manually tests the UAA login workflow. See the above screenshot

### Release note

```release-note
Add support to generate API token with `tanzu api-token create` (supported for UAA as IDP)
```

### DIFF
--- a/docs/cli/commands/tanzu_api-token_create.md
+++ b/docs/cli/commands/tanzu_api-token_create.md
@@ -1,0 +1,29 @@
+## tanzu api-token create
+
+Create a new API Token for Tanzu Platform
+
+```
+tanzu api-token create [flags]
+```
+
+### Examples
+
+```
+
+    # Create an API Token for the Tanzu Platform
+    tanzu api-token create
+
+    # Note: The retrieved token can be used as the value of TANZU_API_TOKEN
+    # environment variable when running 'tanzu login' for non-interactive workflow.
+```
+
+### Options
+
+```
+  -h, --help   help for create
+```
+
+### SEE ALSO
+
+* [tanzu api-token](tanzu_api-token.md)	 - Manage API Tokens for Tanzu Platform
+
--- a/docs/cli/commands/tanzu_api-token_create.md
+++ b/docs/cli/commands/tanzu_api-token_create.md
@@ -1,6 +1,6 @@
 ## tanzu api-token create
 
-Create a new API Token for Tanzu Platform
+Create a new API Token for Tanzu Platform Self-managed
 
 ```
 tanzu api-token create [flags]
@@ -10,7 +10,7 @@ tanzu api-token create [flags]
 
 ```
 
-    # Create an API Token for the Tanzu Platform
+    # Create an API Token for the Tanzu Platform Self-managed
     tanzu api-token create
 
     # Note: The retrieved token can be used as the value of TANZU_API_TOKEN
@@ -25,5 +25,5 @@ tanzu api-token create [flags]
 
 ### SEE ALSO
 
-* [tanzu api-token](tanzu_api-token.md)	 - Manage API Tokens for Tanzu Platform
+* [tanzu api-token](tanzu_api-token.md)	 - Manage API Tokens for Tanzu Platform Self-managed
 
--- a/pkg/auth/common/token.go
+++ b/pkg/auth/common/token.go
@@ -14,13 +14,13 @@ import (
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
+
+	timeutils "github.com/vmware-tanzu/tanzu-cli/pkg/utils/time"
 )
 
 const (
 	extraIDToken = "id_token"
 )
-
-var currentTime = time.Now
 
 const (
 	APITokenType   = "api-token"
@@ -99,7 +99,7 @@ func GetToken(g *types.GlobalServerAuth, tokenGetter func(refreshOrAPIToken, acc
 	g.RefreshToken = token.RefreshToken
 	g.AccessToken = token.AccessToken
 	g.IDToken = token.IDToken
-	expiration := currentTime().Local().Add(time.Duration(token.ExpiresIn) * time.Second)
+	expiration := timeutils.Now().Local().Add(time.Duration(token.ExpiresIn) * time.Second)
 	g.Expiration = expiration
 	g.Permissions = claims.Permissions
 
@@ -173,7 +173,7 @@ func ParseToken(tkn *oauth2.Token, idpType config.IdpType) (*Claims, error) {
 func IsExpired(tokenExpiry time.Time) bool {
 	// refresh at half token life
 	two := 2
-	now := currentTime().Unix()
+	now := timeutils.Now().Unix()
 	halfDur := -time.Duration((tokenExpiry.Unix()-now)/int64(two)) * time.Second
 	return tokenExpiry.Add(halfDur).Unix() < now
 }
--- a/pkg/auth/uaa/tanzu.go
+++ b/pkg/auth/uaa/tanzu.go
@@ -45,7 +45,7 @@ func GetAlternateClientID() string {
 	return clientID
 }
 
-func TanzuLogin(issuerURL string, opts ...common.LoginOption) (*common.Token, error) {
+var TanzuLogin = func(issuerURL string, opts ...common.LoginOption) (*common.Token, error) {
 	issuerEndpoints := getIssuerEndpoints(issuerURL)
 
 	h := common.NewTanzuLoginHandler(issuerURL, issuerEndpoints.AuthURL, issuerEndpoints.TokenURL, tanzuCLIClientID, tanzuCLIClientSecret, defaultListenAddress, defaultCallbackPath, config.UAAIdpType, nil, nil, term.IsTerminal)
--- a/pkg/command/apitoken.go
+++ b/pkg/command/apitoken.go
@@ -24,7 +24,7 @@ import (
 func newAPITokenCmd() *cobra.Command {
 	apiTokenCmd := &cobra.Command{
 		Use:     "api-token",
-		Short:   "Manage API Tokens for Tanzu Platform",
+		Short:   "Manage API Tokens for Tanzu Platform Self-managed",
 		Aliases: []string{"apitoken"},
 		Annotations: map[string]string{
 			"group": string(plugin.SystemCmdGroup),
@@ -42,10 +42,10 @@ func newAPITokenCmd() *cobra.Command {
 func newAPITokenCreateCmd() *cobra.Command {
 	createCmd := &cobra.Command{
 		Use:     "create",
-		Short:   "Create a new API Token for Tanzu Platform",
+		Short:   "Create a new API Token for Tanzu Platform Self-managed",
 		Aliases: []string{},
 		Example: `
-    # Create an API Token for the Tanzu Platform
+    # Create an API Token for the Tanzu Platform Self-managed
     tanzu api-token create
 
     # Note: The retrieved token can be used as the value of TANZU_API_TOKEN
@@ -60,10 +60,10 @@ func newAPITokenCreateCmd() *cobra.Command {
 func createAPIToken(cmd *cobra.Command, _ []string) (err error) {
 	c, err := config.GetActiveContext(types.ContextTypeTanzu)
 	if err != nil {
-		return errors.New("no active context of type `tanzu`. Please login to Tanzu Platform first to generate an API token")
+		return errors.New("no active context found for Tanzu Platform. Please login to Tanzu Platform first to generate an API token")
 	}
 	if c == nil || c.GlobalOpts == nil || c.GlobalOpts.Auth.Issuer == "" {
-		return errors.New("invalid active context of type `tanzu`. Please login to Tanzu Platform first to generate an API token")
+		return errors.New("invalid active context found for Tanzu Platform. Please login to Tanzu Platform first to generate an API token")
 	}
 	// Make sure it is of type tanzu with tanzuIdpType as `uaa` else return error
 	if idpType, exist := c.AdditionalMetadata[config.TanzuIdpTypeKey]; !exist || idpType != string(config.UAAIdpType) {
--- a/pkg/command/apitoken.go
+++ b/pkg/command/apitoken.go
@@ -1,0 +1,104 @@
+// Copyright 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package command
+
+import (
+	"fmt"
+	"net/url"
+
+	"github.com/fatih/color"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/plugin"
+
+	commonauth "github.com/vmware-tanzu/tanzu-cli/pkg/auth/common"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/auth/uaa"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/cli"
+	"github.com/vmware-tanzu/tanzu-cli/pkg/constants"
+)
+
+func newAPITokenCmd() *cobra.Command {
+	apiTokenCmd := &cobra.Command{
+		Use:     "api-token",
+		Short:   "Manage API Tokens for Tanzu Platform",
+		Aliases: []string{"apitoken"},
+		Annotations: map[string]string{
+			"group": string(plugin.SystemCmdGroup),
+		},
+	}
+
+	apiTokenCmd.SetUsageFunc(cli.SubCmdUsageFunc)
+	apiTokenCmd.AddCommand(
+		newAPITokenCreateCmd(),
+	)
+
+	return apiTokenCmd
+}
+
+func newAPITokenCreateCmd() *cobra.Command {
+	createCmd := &cobra.Command{
+		Use:     "create",
+		Short:   "Create a new API Token for Tanzu Platform",
+		Aliases: []string{},
+		Example: `
+    # Create an API Token for the Tanzu Platform
+    tanzu api-token create
+
+    # Note: The retrieved token can be used as the value of TANZU_API_TOKEN
+    # environment variable when running 'tanzu login' for non-interactive workflow.`,
+		RunE:              createAPIToken,
+		ValidArgsFunction: noMoreCompletions,
+	}
+
+	return createCmd
+}
+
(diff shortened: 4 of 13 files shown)


pkg/command/apitoken.go
```diff
@@ -0,0 +1,89 @@
...
+	c, err := config.GetActiveContext(types.ContextTypeTanzu)
+	if err != nil {
+		return errors.New("No active context of type `tanzu`. Please login to Tanzu Platform first to generate the API token")
+	}
+	if c == nil || c.GlobalOpts == nil || c.GlobalOpts.Auth.Issuer == "" {
+		return errors.New("Invalid active context of type `tanzu`. Please login to Tanzu Platform first to generate the API token")
+	}
+	// Make sure it is of type tanzu with tanzuIdpType as `uaa` else return error
+	if idpType, exist := c.AdditionalMetadata[config.TanzuIdpTypeKey]; !exist || idpType != "uaa" {
+		return errors.New("invalid IDP type for the active context. Only UAA IDP type is supported for generating API token")
```
vuil, line 64: I think we need a message more comprehensible to user not familiar with how auth is implemented in TP. Maybe something to the effect that "The generation of the API TOKEN for <public endpoint> should be done via <link to CSP UI>")

pkg/command/apitoken.go
```diff
@@ -0,0 +1,89 @@
...
+	token, err = uaa.TanzuLogin(c.GlobalOpts.Auth.Issuer, loginOptions...)
+	if err != nil {
+		return errors.Wrap(err, "unable to login")
+	}
+	cyanBold := color.New(color.FgCyan).Add(color.Bold)
+	bold := color.New(color.Bold)
+
+	fmt.Fprint(cmd.OutOrStdout(), bold.Sprint("==\n\n"))
+	fmt.Fprintf(cmd.OutOrStdout(), "%s Your generated API token is: %s\n\n", bold.Sprint("API Token Generation Successful!"), cyanBold.Sprint(token.RefreshToken))
+	fmt.Fprintf(cmd.OutOrStdout(), "%s The token will expire on %s.\n", bold.Sprint("Token Expiration:"), bold.Sprint(timeutils.Now().Local().Add(time.Second*time.Duration(token.ExpiresIn)).Format(("2006-01-02 15:04:05"))))
```
vuil, line 84: This is not quite right, unfortunately ExpiresIn only refers to validity of the access token. I suggest we remove this line,
then L86 should be next, followed by L85, which needs to be edited to remove reference to the expiration time.

vuil (commented):
Some comments about the output as I continue the review

pkg/utils/time/time.go
```diff
@@ -0,0 +1,9 @@
+// Copyright 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package time contains time specific functions and variable to make it to overwrite for unit tests
+package time
+
+import "time"
+
+var Now = time.Now
```
vuil, line 9: had to do something similar in this [PR](https://github.com/vmware-tanzu/tanzu-cli/pull/819/files#diff-e7351021bcc81adaa7ed3b1f735852841a7bad195f45b957c71aca058f04d9a4R23), but putting this in its own package is cleaner.

pkg/command/apitoken.go
```diff
@@ -0,0 +1,89 @@
...
+	if idpType, exist := c.AdditionalMetadata[config.TanzuIdpTypeKey]; !exist || idpType != "uaa" {
+		return errors.New("invalid IDP type for the active context. Only UAA IDP type is supported for generating API token")
+	}
+
+	var token *commonauth.Token
+	// If user chooses to use a specific local listener port, use it
+	// Also specify the client ID to use for token generation
+	loginOptions := []commonauth.LoginOption{
+		commonauth.WithListenerPortFromEnv(constants.TanzuCLIOAuthLocalListenerPort),
+		commonauth.WithClientID(uaa.TanzuCLIClientIDExtended),
```
vuil, line 72: reminder for one of the two PRs to leverage https://github.com/vmware-tanzu/tanzu-cli/pull/819/files#diff-21acb112c12bd989b0b1cc0d898ba6479f1f7c5c914fc9baf5b9500d65c7833eR28

pkg/command/apitoken_test.go
```diff
@@ -0,0 +1,204 @@
...
+			cmd := &cobra.Command{}
+			outputBuf := &bytes.Buffer{}
+			cmd.SetOutput(outputBuf)
+
+			mockTimeService := &mockTimeService{}
+			timeutils.Now = mockTimeService.Now
+			defer func() { timeutils.Now = time.Now }()
+
+			mockTanzuLogin := &mockTanzuLogin{}
+			uaa.TanzuLogin = mockTanzuLogin.TanzuLogin
```
vuil, line 179: Should we remember the uaa.TanzuLogin and restore it on a defer so we minimize the impact of the function patching?

vuil (commented):
looking good in general, thanks the for changes.
I have a few questions and suggestions.
And some UX output needs to be updated

pkg/command/apitoken.go
```diff
@@ -0,0 +1,86 @@
...
+		RunE:    createAPIToken,
+	}
+
+	return createCmd
+}
+
+func createAPIToken(cmd *cobra.Command, _ []string) (err error) {
+	c, err := config.GetActiveContext(types.ContextTypeTanzu)
+	if err != nil {
+		return errors.New("No active context of type `tanzu`. Please login to Tanzu Platform first to generate an API token")
```
marckhouzam, line 55: Nit: Error message normally don't start with an uppercase letter.  We used to get a warning about these.  I'm not sure why we don't anymore.  Unless I remember the warning from another repo...

pkg/command/apitoken.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	if c == nil || c.GlobalOpts == nil || c.GlobalOpts.Auth.Issuer == "" {
+		return errors.New("Invalid active context of type `tanzu`. Please login to Tanzu Platform first to generate an API token")
```
marckhouzam, line 58: Same nit about uppercase first letter

pkg/command/apitoken.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	// Make sure it is of type tanzu with tanzuIdpType as `uaa` else return error
+	if idpType, exist := c.AdditionalMetadata[config.TanzuIdpTypeKey]; !exist || idpType != "uaa" {
```
marckhouzam, line 61: Nit: can we use `config.UAAIdpType` instead of `"uaa"`?

pkg/command/apitoken.go
```diff
@@ -0,0 +1,86 @@
...
+	token, err = uaa.TanzuLogin(c.GlobalOpts.Auth.Issuer, loginOptions...)
+	if err != nil {
+		return errors.Wrap(err, "unable to login")
+	}
+	cyanBold := color.New(color.FgCyan).Add(color.Bold)
+	bold := color.New(color.Bold)
+
+	fmt.Fprint(cmd.OutOrStdout(), bold.Sprint("==\n\n"))
+	fmt.Fprintf(cmd.OutOrStdout(), "%s Your generated API token is: %s\n\n", bold.Sprint("API Token Generation Successful!"), cyanBold.Sprint(token.RefreshToken))
+	fmt.Fprintf(cmd.OutOrStdout(), "For non-interactive login use the API token as follow: %s\n\n", cyanBold.Sprint("TANZU_API_TOKEN=<token> tanzu login --endpoint <tanzu-platform-endpoint>"))
```
marckhouzam, line 82: "follow" -> "follows"

pkg/command/apitoken.go
```diff
@@ -0,0 +1,86 @@
...
```
marckhouzam, line 82: Just a thought; would it be nicer for the user if we printed the real token instead `<token>` in the above instructions?

marckhouzam, line 82: This token can only be used for UAA, correct?
If so, instead of `<tanzu-platform-endpoint>`, should we put something with `uaa` in it?

pkg/command/apitoken.go
marckhouzam: Do we know the expiration time easily?
Since we tell the user `regenerate a new token before expiration`, I find it nice to tell them when that will be.  If not, is there another way for them to avoid their usage failing when expiration happens?

marckhouzam (commented):
Nice.
Minor suggestions to take or leave.

pkg/command/apitoken.go
vuil: That UAA is used is implementation details that may be lost on many users. In a comment earlier, I suggested using terminology that is more comprehensible. Still thinking of options...

anujc25:
> Do we know the expiration time easily?
Since we tell the user regenerate a new token before expiration, I find it nice to tell them when that will be. If not, is there another way for them to avoid their usage failing when expiration happens?

@marckhouzam regarding this, I discussed with @vuil and the conclusion is we do not have a way to know the expiration time for the Refresh Token at the moment. So, if needed we need to document the expiration time separately in the docs.

marckhouzam:
> @marckhouzam regarding this, I discussed with @vuil and the conclusion is we do not have a way to know the expiration time for the Refresh Token at the moment. So, if needed we need to document the expiration time separately in the docs.

So we will document that the token has a lifetime of X time?
Do we prefer only putting it in the docs so we can more easily change it, instead of printing the value we currently know?

I assume there is no advanced notification for the user to know the token is about to expire?

anujc25:
> > @marckhouzam regarding this, I discussed with @vuil and the conclusion is we do not have a way to know the expiration time for the Refresh Token at the moment. So, if needed we need to document the expiration time separately in the docs.
> 
> So we will document that the token has a lifetime of X time? Do we prefer only putting it in the docs so we can more easily change it, instead of printing the value we currently know?
> 
> I assume there is no advanced notification for the user to know the token is about to expire?

Yes. We do not know the lifetime of this refresh token at the moment. And that can be updated without CLI's knowledge from the backend so it is safe to put that information in the docs.
Correct, there isn't any advanced notification for the user to know the token is about to expire.
Please note that this API token generation from the CLI is a temporary workaround for supporting the non- interactive login with UAA because UAA does not have any UI to generate this token. There will be updates and enhancement in this area in future would likely make CLI more user friendly.

pkg/command/apitoken.go
marckhouzam: Nice !

marckhouzam: You are right.  Maybe "TPSM" is more appropriate, or the long form of it?

pkg/command/apitoken.go
anujc25: Apologies for the confusion and not using this thread to discuss this topic. I responded separately https://github.com/vmware-tanzu/tanzu-cli/pull/820#issuecomment-2397636551

Also, I have updated the PR description to remove the outdated screenshot and added the sample output (without color).

pkg/auth/uaa/tanzu.go
```diff
@@ -14,10 +14,14 @@ import (
...
 	// only allowed Redirect URI and does not have an associated client secret.
-	tanzuCLIClientID     = "tp_cli_app"
-	tanzuCLIClientSecret = ""
-	defaultListenAddress = "127.0.0.1:0"
-	defaultCallbackPath  = "/callback"
+	TanzuCLIClientID = "tp_cli_app"
+	// Alternate client ID for UAA associated with a longer refresh token
+	// validity. Use this for CLI use cases where it is impractical to
+	// interactively reauthenticated once the refresh token expires.
+	TanzuCLIClientIDExtended = "tp_cli_app" // TODO(vui) update once available.
```
marckhouzam, line 21: This is the only thing that prevents me from approving.  Unless we want to merge and update this when the info is available?

pkg/utils/time/time.go
marckhouzam, line 9: I don't see this used anywhere.
Also, I don't understand how it would be used...  What is the different in calling `time.Now()` from go and `time.Now()` from this new package?

pkg/auth/common/login_handler.go
```diff
@@ -113,6 +113,14 @@ func WithOrgID(orgID string) LoginOption {
 	}
 }
 
+// WithClientID causes the login with given ClientID.
+func WithClientID(clientID string) LoginOption {
+	return func(h *TanzuLoginHandler) error {
+		h.clientID = clientID
```
marckhouzam, line 119: I see in @vuil PR https://github.com/vmware-tanzu/tanzu-cli/pull/819/files#diff-c2506254f4d5d93b4c84bfa32c18782a79fbca5c788f38444419c9af5060cca1R130 that he also sets the `h.oauthConfig.ClientID` field here.

pkg/utils/time/time.go
anujc25: I will remove this now. I used in the earlier version of this PR for unit tests but I removed that because we removed the expiration time from the log message.

pkg/auth/common/login_handler.go
anujc25: This needs to be updated as part of rebase after #819 gets merged because the same function is implemented there as well.

pkg/utils/time/time.go
anujc25: I will keep this for now and update the https://github.com/vmware-tanzu/tanzu-cli/pull/819#discussion_r1790783947 as part of the rebase.

pkg/command/apitoken.go
anujc25: Updated to use `uaa.GetAlternateClientID()` after rebase.

pkg/auth/uaa/tanzu.go
anujc25: Rebased. This does not apply anymore.

pkg/command/apitoken.go
vuil: Maybe the way it is implemented is reasonable enough. Users who have to use the form "tanzu login --endpoint ..." and who has successfully use "api-token create" to get to this point should know what tanzu-platform-endpoint refers to.

anujc25: I have updated the log message to mention actual Tanzu Platform endpoint that needs to be used instead of `<tanzu-platform-endpoint>`. Let me know if any other changes are needed here.

pkg/auth/common/token.go
```diff
@@ -99,7 +99,7 @@ func GetToken(g *types.GlobalServerAuth, tokenGetter func(refreshOrAPIToken, acc
 	g.RefreshToken = token.RefreshToken
 	g.AccessToken = token.AccessToken
 	g.IDToken = token.IDToken
-	expiration := currentTime().Local().Add(time.Duration(token.ExpiresIn) * time.Second)
+	expiration := timeutils.Now().Local().Add(time.Duration(token.ExpiresIn) * time.Second)
```
vuil, line 102: thanks :-)

pkg/command/apitoken.go
```diff
@@ -0,0 +1,105 @@
...
+	c, err := config.GetActiveContext(types.ContextTypeTanzu)
+	if err != nil {
+		return errors.New("no active context of type `tanzu`. Please login to Tanzu Platform first to generate an API token")
+	}
+	if c == nil || c.GlobalOpts == nil || c.GlobalOpts.Auth.Issuer == "" {
+		return errors.New("invalid active context of type `tanzu`. Please login to Tanzu Platform first to generate an API token")
+	}
+	// Make sure it is of type tanzu with tanzuIdpType as `uaa` else return error
+	if idpType, exist := c.AdditionalMetadata[config.TanzuIdpTypeKey]; !exist || idpType != string(config.UAAIdpType) {
+		return errors.New("only UAA IDP type is supported for generating an API token. The generation of the API token for public endpoint should be done via https://console.tanzu.broadcom.com")
```
vuil, line 71: Suggest: remove reference to UAA IDP type. 
maybe

Command no supported. Please refer to documentation on how to generate an API token for a public Tanzu Platform endpoint via https://console.tanzu.broadcom.com

pkg/command/apitoken.go
```diff
@@ -0,0 +1,105 @@
...
+		if err == nil {
+			tpEndpoint = fmt.Sprintf("%s://%s", u.Scheme, u.Host)
+		}
+	}
+
+	cyanBold := color.New(color.FgCyan).Add(color.Bold)
+	bold := color.New(color.Bold)
+
+	fmt.Fprint(cmd.OutOrStdout(), bold.Sprint("==\n\n"))
+	fmt.Fprintf(cmd.OutOrStdout(), "%s Your generated API token is: %s\n\n", bold.Sprint("API Token Generation Successful!"), cyanBold.Sprint(token.RefreshToken))
```
vuil, line 100: We may have to workshop the err/output messages in a followup. Some thoughts


- Is it better to output the token in stdout, and show the rest in stderr?
- do we want to advocate TANZU_API_TOKEN=%s tanzu login, or mention the config file way to persist the env var?

The messages in these function

- may confuse users who are interacting with the public endpoint, who ends up creating a valid `tanzu` endpoint and still get a surprise when this fails for him
- the instructions for regenerate and re-login might be confusing to the user of the current CLI instance (who has obviously logged in)

docs/cli/commands/tanzu_api-token_create.md
```diff
@@ -0,0 +1,30 @@
...
+### Examples
+
+```
+
+		# Create an API Token for the Tanzu Platform
+		tanzu api-token create
+	
+		# Note: The retrieved token can be used as the value of TANZU_API_TOKEN
+		# when running 'tanzu login'. For example:
+		TANZU_API_TOKEN=<token> tanzu login
```
anujc25, line 18: extra spaces at the front.

vuil (approved):
I think we may still want to tweak the messages but the rest lgtm, thanks!

pkg/command/apitoken.go
anujc25: > Is it better to output the token in stdout, and show the rest in stderr?

I would prefer using `stdout` for both the token and the rest of the output. Since this is an output message intended for the user, it's more conventional to keep all output in `stdout`. 

> do we want to advocate TANZU_API_TOKEN=%s tanzu login, or mention the config file way to persist the env var?

I feel persisting the env var in the config file is a generic way to configure any env variable. So I don't feel to strongly that we also mention the persisting the TANZU_API_TOKEN in the configuration file. Also if user is not aware of what it means to persist a setting in the config file as env.TANZU_API_TOKEN it might create even more confusion because all the follow up `tanzu login` attempt will try to use the token automatically.

PR #821, opened 2024-10-04
Improve startup speed

### LINKED ISSUE #823: Inefficient startup
**Bug description**

The Tanzu CLI has an inefficient startup which becomes slower and slower as the number of configured contexts increases.

`pkg/config/init.go` makes a call to `SyncContextsAndServers()` in its `init()` function; this means `SyncContextsAndServers()` is called every time the CLI is executed.  Within the `SyncContextsAndServers()` function, there is a loop that iterates over all the configured contexts, and for each one a call to `config.SetContext()` is made which will in turn persist the configuration.  For each `tanzu` context, the configuration is persisted once, meaning writing to the `config.yaml` file and the `config-ng.yaml` file. For non-`tanzu` contexts, the configuration is actually persisted twice, which means writing to files 4 times.

I somehow accumulated 29 contexts in my configuration, which, because of the logic described above, make every command of the CLI noticeably slower.

Furthermore, the point of calling `SyncContextsAndServers()` is too handle certain scenarios where old plugins can modify the `config.yaml` file incorrectly, and the `SyncContextsAndServers()` will fix it.  However, it the vast, vast majority of cases, this synchronization does nothing, although it still writes to the 2 configuration files multiple time and drastically slows down the CLI.

Below shows that with 27 contexts, a simple help printout takes a full second, while if I remove all contexts, it drops to close to 0.2 seconds.
```
$ tz version
version: v1.5.0
buildDate: 2024-09-26
sha: 9294e28
arch: arm64

$ tz context list | wc -l
[i] Use '--wide' to view additional columns.
      29

$ time tz -h> /dev/null
tanzu -h > /dev/null  0.91s user 0.07s system 105% cpu 0.924 total
$ mv ~/.config/tanzu/config* ~
overwrite /Users/kmarc/config-ng.yaml? (y/n [n]) y
overwrite /Users/kmarc/config.yaml? (y/n [n]) y
$ time tz -h> /dev/null
tanzu -h > /dev/null  0.06s user 0.02s system 37% cpu 0.222 total
```
**Expected behavior**

The execution time of CLI commands should not be affected so dramatically by the number of contexts configured.

Run `time tanzu -h >/dev/null` and keep creating more contexts and see the time increase steadily.
Any speed slower than 0.5 seconds starts being very noticeable when doing shell completion.

**Output of `tanzu version`**

This started with CLI v1.2.0 and still affecting `main` today.

Locally on my ARM64 Mac.

marckhouzam:

This PR avoids writing to the config files and the `.data-store.yaml` on startup if not necessary.

Writing to the config file to update the server list has a cost for each context.  I ended up with 27 contexts configured, which triggered writing to file over 50 times which added 0.65 seconds on my M1 Mac, making the startup jump from around 0.35 seconds to a full second. Having the CLI take a full second to execute any command implies that each call to shell completion takes at least 1 second, which is slow for the user.

Updating the server list does not actually need to be done all the time. 

Furthermore, the CLI currently writes to the `.data-store.yaml` and to the config files on every command when it stores the LastCLIVersionUsed.  This file writes are also unnecessary most of the time.

This PR has two commits:
1. compute a sha of the contexts and another for the servers and store it in the data-store.  Then for each command, instead of blindly synching contexts and servers, only do the sync if either sha has changed.
2. on write the last executed cli version to the data-store if the version is different than what is stored already, and only disable the `features.global.context-target-v2` feature if it is currently enabled (to avoid writing to the config files)

This reduces the execution time for commands that don't access the backend; this means that shell completion for commands and flags becomes noticeably more responsive when a user has multiple contexts configured.

Before this PR, running `tanzu version` with 9 contexts configures takes around 0.5 seconds.
With this PR is becomes closer to 0.3 seconds.

Fixes #823

I use this little utility called `entr` which will execute a command whenever a specific file is changed.
```
# Monitor changes to the two config file and the datastore file
$ echo ~/.config/tanzu/config.yaml| entr echo config changed >>out.txt
$ echo ~/.config/tanzu/config-ng.yaml| entr echo config-ng changed >>out.txt
$ echo ~/.config/tanzu/.data-store.yaml | entr echo datastore changed >>out.txt
$ tail -f out.txt
```

Before the PR, if I run any command such as `tanzu -h >/dev/null` I see the config files change for every command multiple times (one for each context).

And for commands that trigger the lastTanzuVersion check (most commands), such as `tanzu plugin list >/dev/null`, I see the datastore file updated and the config files also. 

With this PR, running those same commands, I don't see any changes done to the config or datastore files.

I also ran some normal shell completion down the command tree, and the responsiveness is noticeably better.

### Release note

```release-note
Improve CLI execution speed, especially for shell completion responsiveness
```

### DIFF
--- a/pkg/config/context.go
+++ b/pkg/config/context.go
@@ -4,16 +4,35 @@
 package config
 
 import (
-	"github.com/pkg/errors"
+	"crypto/sha256"
+	"encoding/hex"
 
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+
+	"github.com/vmware-tanzu/tanzu-cli/pkg/datastore"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 )
+
+type contextAndServerSha struct {
+	ContextsSha        string
+	CurrentContextsSha string
+	ServersSha         string
+	CurrentServerSha   string
+}
+
+const lastContextServerShasKey = "lastContextServerShas"
 
 // SyncContextsAndServers populate or sync contexts and servers
 func SyncContextsAndServers() error {
 	cfg, err := config.GetClientConfig()
 	if err != nil {
 		return errors.Wrap(err, "failed to get client config")
+	}
+
+	if !shouldSyncContextsAndServers(cfg) {
+		return nil
 	}
 
 	config.PopulateContexts(cfg)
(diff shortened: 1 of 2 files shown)


pkg/config/context.go
```diff
@@ -16,7 +17,11 @@ func SyncContextsAndServers() error {
...
 
-	config.PopulateContexts(cfg)
+	hasChanged := config.PopulateContexts(cfg)
+	needsSync := doServersNeedUpdate(cfg)
+	if !hasChanged && !needsSync {
+		return nil
+	}
 
 	// Now write the context to the configuration file.  This will also create any missing server for its corresponding context
 	for _, c := range cfg.KnownContexts {
```
vuil, line 27: This optimization is quite clean and much appreciated.

I am not sure if we had in the past rely on this to update existing (not create new) entry to correct for any inconsistencies. cc @prkalle 
If we want to be really safe, we can consider some undocumented env var to restore back the old behavior for any user who needs it.

vuil (commented):
Overall optimization is quite clean and much appreciated, thanks for looking into this.

I have a very tiny concern if the skipping of the SetContext() has any unintended consequences, and if we want to cover our bases.

pkg/config/context.go
marckhouzam: Looking at the #616 and what was being done before, I see that we were calling https://github.com/vmware-tanzu/tanzu-plugin-runtime/blob/7278b2554915a10ad3abc9e8befe0a3eb196e1a3/config/legacy_clientconfig_factory.go#L21.  I based my check on that code to know if there was any servers needing to be updated.  So I feel we are covered on that front.

However, there is another aspect to consider.  When saving the configuration to file I notice that there is some cleanup of "legacyDiscardedContexts" done here: https://github.com/vmware-tanzu/tanzu-plugin-runtime/blob/7278b2554915a10ad3abc9e8befe0a3eb196e1a3/config/config_factory.go#L144-L152

To reproduce the situation I installed `cluster:k8s:v0.25.4` and simply ran `tanzu cluster version`.  What I see is that the `config.yaml` file now has a `contexts` node with all the servers included there.  Before this PR, that `contexts` node would be removed on the next CLI run.  With this PR it is only cleaned up whenever the configuration is changed later on.

I don't think keeping this `contexts` node is problematic, but let's discuss it

I

marckhouzam: We have revisited the solution and are now much more thorough.

pkg/config/context.go
```diff
@@ -34,5 +53,76 @@ func SyncContextsAndServers() error {
...
+	}
+	serverBytes, err := yaml.Marshal(nonTanzuServers)
+	if err != nil {
+		return nil
+	}
+
+	return &contextAndServerSha{
+		ContextsSha: hashString(string(ctxBytes)),
+		// Only the k8s current context is synched
+		CurrentContextsSha: hashString(cfg.CurrentContext[types.ContextTypeK8s]),
```
anujc25, line 113: What about the `ContextTypeTMC`? Do we sync the current TMC context as well?

anujc25 (approved):
LGTM. Very nice way to test the change as well. I just have one question.

pkg/config/context.go
marckhouzam: In TPR, when we save a TMC context, we also save it on the config.yaml file as a server.
However, when the change the current TMC context, we don't sync this.
Therefore I only computed the SHA on the k8s context.

See https://github.com/vmware-tanzu/tanzu-plugin-runtime/blob/b5f34992ce053bc8979eec0a05260eaab828a193/config/contexts.go#L265-L266

anujc25: Ack. Thanks for confirming.

anujc25 (approved):

vuil (approved):
Very nice approach of comparing the server and context contents!
lgtm!

PR #827, opened 2024-11-01
Print stderr output of plugin when doing completion

marckhouzam:

A plugin, when doing shell completion, may print debug printouts to stderr.  When the plugin is invoked through the tanzu cli, those printouts were being lost.  This PR prints them out to stderr to allow for better troubleshooting of plugin shell completion.

Fixes # N/A

Before the PR:
```
# Notice that if I call the __complete command directly on the plugin, I get a debug printout "EDS query"
# This printout is on stderr
$ TANZU_BIN=tanzu TANZU_CLI_LOG_LEVEL=7 $rbac/artifacts/plugins/darwin/arm64/global/rbac/v0.0.0-dev-31f037f/tanzu-rbac-darwin_arm64 __complete role get '' > /dev/null
[i] Fetched 4 roles from EDS Query

# Notice here however, that if I call the __complete command on the tanzu CLI
# which will in turn call __complete on the plugin, the debug printout is not seen
$ TANZU_CLI_LOG_LEVEL=7 tanzu __complete rbac role get '' > /dev/null
```
With this PR:
```
# Now the debug printout is visible.
# Notice also that as a side-effect, there is a second "directive" printout;
# the first one is coming from the plugin, which was previously lost, while
# the second is the one from the CLI itself.
$ TANZU_CLI_LOG_LEVEL=7 tanzu __complete rbac role get '' > /dev/null
[i] Fetched 4 roles from EDS Query
```

### Release note

```release-note
Improve shell completion debugging by showing debug printouts.
```

### DIFF
--- a/pkg/cli/plugin_cmd.go
+++ b/pkg/cli/plugin_cmd.go
@@ -211,10 +211,14 @@ func getCmdForPluginEx(p *PluginInfo, cmdName string, mapEntry *plugin.CommandMa
 		runner := NewRunner(p.Name, p.InstallationPath, completion)
 		ctx := context.Background()
 		setupPluginEnv(srcHierarchy, dstHierarchy)
-		output, _, err := runner.RunOutput(ctx)
+		output, errOutput, err := runner.RunOutput(ctx)
 		if err != nil {
 			return nil, cobra.ShellCompDirectiveError
 		}
+
+		// Print the stderr output from the plugin back to stderr.
+		// This allows for debugging printouts to be seen.
+		fmt.Fprint(cmd.ErrOrStderr(), errOutput)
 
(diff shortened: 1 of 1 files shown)


vuil (approved):
Just want to make sure:
It felf weird initially, but the double-printing of the "Completion ended with..." looks purely cosmetic and for a command whose output is hidden no less, so change looks good. Thanks

anujc25 (approved):
LGTM. Thanks for fixing this.

marckhouzam:
> Just want to make sure: It felf weird initially, but the double-printing of the "Completion ended with..." looks purely cosmetic and for a command whose output is hidden no less, so change looks good. Thanks

Yes, it's a printout on stderr just meant to help a cobra-program developer debug their completion code, it is completely ignored by the completion shell script

PR #831, opened 2024-11-18
Do not hide remapped subcommands

### LINKED ISSUE #830: A hidden root command for a plugin hides all remapped sub-commands
**Bug description**

Say I create a plugin with a descriptor that has the field `Hidden == true`.
My goal in doing that is to hide the plugin's root command.
For example:
```
plugin.PluginDescriptor{
  Name:           "rbac",
  CommandMap: []plugin.CommandMapEntry{
  {
    // Remap "tanzu rbac role" to "tanzu role" leaving no other commands under "tanzu rbac" 
    SourceCommandPath:      "role",
    DestinationCommandPath: "role",
  },
  // Hide the "tanzu rbac" command since it has no subcommands left
  Hidden: true,
}
```

However, with the above and the current tanzu CLI, any remapped subcommands also get hidden.
So in the example above not only is `tanzu rbac` hidden, but `tanzu role` is also hidden, which is not correct.

**Expected behavior**

Remapped subcommands don't get hidden, while only the root plugin command gets hidden.

See above.

**Output of `tanzu version`**

Today's `main`: ef7601df4

marckhouzam:

If the plugin's descriptor's `Hidden` field is set to true, only the root command of the plugin should be hidden while any mapped subcommands should stay visible.

Fixes #830 

I tested with a plugin that set the descriptor's `Hidden` field to true, and remaps a subcommand.
```
$ tz rbac info | jq
{
  "name": "rbac",
 [...]
 "hidden": true,
  "completionType": 0,
  "commandMap": [
    {
      "srcPath": "role",
      "dstPath": "role",
      "overrides": "",
      "description": "Tanzu Platform Role-Based Access Control",
      "aliases": [
        "roles"
      ]
    }
  ],
  "pluginRuntimeVersion": "v1.4.3",
  "binaryArch": "arm64"
}

# Notice that the role command is visible and that the rbac command is hidden
# Before this fix, the role command was also hidden
$ tz -h | egrep 'rbac|role'
    role                    Tanzu Platform Role-Based Access Control
```

### Release note

```release-note
Do not hide remapped subcommands when the plugin's root command is hidden.
```

### DIFF
--- a/pkg/cli/plugin_cmd.go
+++ b/pkg/cli/plugin_cmd.go
@@ -155,6 +155,7 @@ func getCmdForPluginEx(p *PluginInfo, cmdName string, mapEntry *plugin.CommandMa
 	var srcHierarchy, dstHierarchy []string
 	aliases := p.Aliases
 	description := p.Description
+	hidden := p.Hidden
 
 	if mapEntry != nil {
 		srcHierarchy = hierarchyFromPath(mapEntry.SourceCommandPath)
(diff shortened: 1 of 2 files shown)


vuil (approved):
LGTM.
Nice clean fix and tests.
Thanks for reporting and fixing the issue !

PR #837, opened 2025-01-09
Allow building multiple RPM packages

marckhouzam:

This PR splits the RPM package-building script into two scripts:
- one to build a package and sign it
- one to build the repo with all the packages that may have been built

The first script can be called more than once by specifying the new `RPM_PACKAGE_NAME` with a different name.  This allows to build the same RPM package but with a different name,  The value in doing this is that a different signer can be used for the different package; for example, a different signing key could be used for a second package.

This will allow to build a package specific to Centos8/9 which will be signed with a special key that works on those OS:s.

Fixes # N/A

I have built the package and repo locally using the same make target as before and confirms it builds like before and is installable.  This is not a signed package.  Notice I fake a version to build a `tanzu-cli` package instead of a `tanzu-cli-unstable` package
```
$ make build-cli-linux-amd64 BUILD_VERSION=v1.12.0
build linux-amd64 CLI with version: v1.12.0
$ make rpm-package BUILD_VERSION=v1.12.0
[...]
$ ls hack/rpm/_output/rpm/tanzu-cli/
repodata                       tanzu-cli-1.12.0-1.aarch64.rpm tanzu-cli-1.12.0-1.x86_64.rpm
```
I have build the unstable package ONLY and then built a second unstable package ONLY by overriding the name, which is what we will do for Centos8/9, then built a repo containing both:
```
$ make build-cli-linux-amd64
build linux-amd64 CLI with version: v1.6.0-dev
$ make rpm-package-only
[...]
$ ls hack/rpm/_output/rpm/tanzu-cli/
tanzu-cli-unstable-1.6.0-0.1_dev.aarch64.rpm tanzu-cli-unstable-1.6.0-0.1_dev.x86_64.rpm

# Build a second package with a different name (would normally use a different signer)
# Notice that the packages accumulate in the directory
$ make rpm-package-only RPM_PACKAGE_NAME=tanzu-cli-centos9
$ ls hack/rpm/_output/rpm/tanzu-cli/
tanzu-cli-centos9-unstable-1.6.0-0.1_dev.aarch64.rpm tanzu-cli-unstable-1.6.0-0.1_dev.aarch64.rpm
tanzu-cli-centos9-unstable-1.6.0-0.1_dev.x86_64.rpm  tanzu-cli-unstable-1.6.0-0.1_dev.x86_64.rpm

# Now build the repo
$ make rpm-package-repo
[...]
$ ls hack/rpm/_output/rpm/tanzu-cli/
repodata                                             tanzu-cli-unstable-1.6.0-0.1_dev.aarch64.rpm
tanzu-cli-centos9-unstable-1.6.0-0.1_dev.aarch64.rpm tanzu-cli-unstable-1.6.0-0.1_dev.x86_64.rpm
tanzu-cli-centos9-unstable-1.6.0-0.1_dev.x86_64.rpm

# Using vi I can look at the repo metadata and confirm it contains both the standard unstable 1.6.0-dev package
# and the centos9 unstable 1.6.0-dev package
vi hack/rpm/_output/rpm/tanzu-cli/repodata/d4de9bf00fa8dd470abc166a7feaada659037cb9926d9ede4e8efed25189e1c2-filelists.xml.zst
```
I have testing that both packages can be installed using docker.

I've also ran the above tests with the `rpm-*-in-docker` make targets to confirm they work as expected.

### Release note

```release-note
Support building multiple RPM packages to be used with different signing keys.
```

### DIFF
--- a/hack/rpm/tanzu-cli.spec
+++ b/hack/rpm/tanzu-cli.spec
@@ -1,13 +1,4 @@
-%if "%{unstable}" == "false"
-Name:       tanzu-cli
-Provides:   tanzu-cli
-Obsoletes:  tanzu-cli  < %{rpm_package_version}
-%else
-Name:       tanzu-cli-unstable
-Provides:   tanzu-cli-unstable
-Obsoletes:  tanzu-cli-unstable  < %{rpm_package_version}
-%endif
-
+Name:       %{rpm_package_name}
 Version:    %{rpm_package_version}
 Release:    %{rpm_release_version}
 License:    Apache 2.0
--- a/Makefile
+++ b/Makefile
@@ -189,17 +189,32 @@ apt-package-in-docker: ## Build a debian package from within a container already
 .PHONY: apt-package
 apt-package: apt-package-only apt-package-repo  ## Build a debian package to use with APT
 
-.PHONY: rpm-package
-rpm-package: ## Build an RPM package
+.PHONY: rpm-package-only
+rpm-package-only: ## Build an RPM package
 	@if [ "$$(command -v docker)" == "" ]; then \
 		echo "Docker required to build rpm package" ;\
 		exit 1 ;\
 	fi
-	docker run --rm -e VERSION=$(BUILD_VERSION) -e RPM_SIGNER=$(RPM_SIGNER) -e RPM_METADATA_BASE_URI=$(RPM_METADATA_BASE_URI) -v $(ROOT_DIR):$(ROOT_DIR) $(RPM_IMAGE) $(ROOT_DIR)/hack/rpm/build_package.sh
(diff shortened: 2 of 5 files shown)


hack/rpm/tanzu-cli.spec
```diff
@@ -1,13 +1,4 @@
-%if "%{unstable}" == "false"
-Name:       tanzu-cli
-Provides:   tanzu-cli
-Obsoletes:  tanzu-cli  < %{rpm_package_version}
```
marckhouzam, line 4: It turns out that we were using `Provides` and `Obsoletes` wrongly.  Those fields are meant to be used when we rename a package (e.g., `tanzu-cli` -> `newname-cli`) so that the new package can indicate it replaces the old package.  We were not using it correctly.

Those fields do not support using a variable, so removing them allowed me to inject the name of the package.

As a separate change, we could use those fields to allow the `tanzu-cli` package to replace an already installed `tanzu-cli-unstable` package; I tried it and couldn't get it to work quickly, so I left it for later, since it is not relevant to the Centos9 issue.

anujc25 (approved):
LGTM. Thanks for all the investigation and reaching to a solution to support centos usecase. 👏

PR #838, opened 2025-01-09
Document how to install on RHEL/CentOS 8/9

marckhouzam:

This PR documents how to install the Tanzu CLI on RHEL/CentOS 8 and 9.

The idea is that users should install the CLI as follows when running on RHEL/CentOS 8 and 9:
```
cat << EOF | sudo tee /etc/yum.repos.d/tanzu-cli.repo
[tanzu-cli]
name=Tanzu CLI
baseurl=https://storage.googleapis.com/tanzu-cli-installer-packages/rpm/tanzu-cli
enabled=1
gpgcheck=1
repo_gpgcheck=1
gpgkey=https://storage.googleapis.com/tanzu-cli-installer-packages/keys/TANZU-PACKAGING-GPG-RSA-KEY-CENTOS9.gpg
       https://storage.googleapis.com/tanzu-cli-installer-packages/keys/TANZU-PACKAGING-GPG-RSA-KEY.gpg
EOF

sudo rpm --import https://storage.googleapis.com/tanzu-cli-installer-packages/keys/TANZU-PACKAGING-GPG-RSA-KEY-CENTOS9.gpg

sudo yum install -y tanzu-cli-centos9 # dnf install can also be used

# or for pre-releases
sudo yum install -y tanzu-cli-centos9-unstable # dnf install can also be used
```

**Why we need to run `rpm --import` manually?**

When testing on a real VM running Centos 9, I learned that because both key are configured in the `/etc/yum.repos.d/tanzu-cli.repo` file, yum will run an `rpm --import` on both and will fail on the standard key `TANZU-PACKAGING-GPG-RSA-KEY.gpg` and the installation will still fail.

So, I did two things to fix this:
1- put the `TANZU-PACKAGING-GPG-RSA-KEY-CENTOS9.gpg` key first in the list of keys in the `/etc/yum.repos.d/tanzu-cli.repo` file.  This means that on a first `yum install` the key will be imported successfully, although the `yum install` will still fail; this allows to run `yum install` a second time, and have it succeed because it will find the required key.
2- have the user run `rpm --import` manually first, to avoid the first failure.

With the instructions in this PR, there should be no `yum install` failure at all.

An alternative since we ask users to import the new `TANZU-PACKAGING-GPG-RSA-KEY-CENTOS9.gpg` key manually, is that we could remove that key from the `/etc/yum.repos.d/tanzu-cli.repo` file.  The value in doing that is that it would keep the instructions the same as for the normal `tanzu-cli` package except for the `rpm –import` and the `yum install tanzu-cli-centos9`.  The value in leaving that second key in the file is that if the user forgets to run the `rpm –import`, then running `yum install` twice will work.

Fixes # N/A

Let CI run

### Release note

```release-note
Document special installation instructions for RHEL/CentOS 8/9
```

The new `tanzu-cli-centos` packages will only be made available starting with CLI `v1.5.2`.
To make them available for already released versions would require building each RPM package with the new name and signing it with the new key.  This would take more effort than we felt was worthwhile.
To install older versions on Centos 8/9, users will need to deactivate the gpg check using `--nogpgcheck`; I have specified this in the documentation update in this PR.

### DIFF
--- a/docs/quickstart/install.md
+++ b/docs/quickstart/install.md
@@ -188,6 +188,12 @@ To uninstall: `sudo apt remove tanzu-cli-unstable`
 
 ### From yum/dnf (RHEL)
 
+> **_NOTE:_** When installing on versions 8 and 9 of RHEL and CentOS, a special `tanzu-cli-centos9`
+> package should be used along with a second GPG key.  Please refer to section:
+> [Using yum/dnf on RHEL/CentOS versions 8 and 9](#using-yumdnf-on-rhelcentos-versions-8-and-9)
+
+For normal installation (not RHEL/CentOS 8/9):
+
 ```console
 cat << EOF | sudo tee /etc/yum.repos.d/tanzu-cli.repo
 [tanzu-cli]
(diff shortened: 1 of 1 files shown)


docs/quickstart/install.md
```diff
@@ -243,6 +249,38 @@ sudo yum install tanzu-cli-unstable
...
+name=Tanzu CLI
+baseurl=https://storage.googleapis.com/tanzu-cli-installer-packages/rpm/tanzu-cli
+enabled=1
+gpgcheck=1
+repo_gpgcheck=1
+gpgkey=https://storage.googleapis.com/tanzu-cli-installer-packages/keys/TANZU-PACKAGING-GPG-RSA-KEY.gpg
+       https://storage.googleapis.com/tanzu-cli-installer-packages/keys/TANZU-PACKAGING-GPG-RSA-KEY-CENTOS9.gpg
+EOF
+
+sudo yum install -y tanzu-cli-centos9 # dnf install can also be used
```
vuil, line 278: nit (feel free to ignore if I am too paranoid): 

maybe also add a `# install latest official release` before line 278.
maybe the other can be 
```
or
# install latest pre-release
```

Not that the first yum install line is not obvious, but more to make it clear that the OR is for deciding official vs prelease, but in both cases all the pub key setup before the steps still applies.

vuil (approved):
Nice! Just have a tiny suggestion to maybe improve clarity marginally.

vuil (approved):
lgtm

PR #840, opened 2025-01-10
Implement singleton pattern for NewRootCmd to prevent multiple creations

### LINKED ISSUE #825: tanzu-cli 1.4.1 on Linux context create hangs on kubernetes-release
**Bug description**
I am essentially here: https://docs.vmware.com/en/VMware-vSphere/8.0/vsphere-with-tanzu-tkg/GUID-E5A804FA-BB03-436F-BF01-14CBF13DBB9D.html . Attemping to create a context against my Tanzu environment.

This works on Windows with tanzu-cli 1.4.1 down to generating me my kubeconfig. The same exact process, same endpoints, hangs on linux.

There is a load of CPU activity from the kswap0 process during this hang. My only way to exit is to eventually ctrl c out of it as the system becomes quite unresponsive.

**Expected behavior**
Plugins are installed and drops back out to shell, ready to move on to listing/connecting to a tanzu namespace.

```
dnf install tanzu-cli
tanzu plugin install --group vmware-tkg/default (installs pinniped auth)
tanzu context create MyContext --endpoint https://10.80.1.2


[hang on Linux Alma 9] @ - Installing plugin 'kubernetes-release:v0.29.0' with target 'kubernetes'

…
[or succeeds in Win 11]
```

**Output of `tanzu version`**
```
version: v1.4.1
buildDate: 2024-08-14
sha: bbde4b21
arch: amd64
```

on both OS'es

Win11 or Alma 9.1

anujc25:

- Fix pinniped plugin invocation by not creating NewRootCmd and using existing one

Fixes #825

### Release note

```release-note
Implement singleton pattern for `NewRootCmd` to prevent multiple creations
```

### DIFF
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -12,6 +12,7 @@ import (
 	"os/signal"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -80,8 +81,8 @@ func convertInvokedAs(plugins []cli.PluginInfo) {
 	}
 }
 
-// NewRootCmd creates a root command.
-func NewRootCmd() (*cobra.Command, error) { //nolint: gocyclo
+// createRootCmd creates a root command.
+func createRootCmd() (*cobra.Command, error) { //nolint: gocyclo
 	go interruptHandle()
 	var rootCmd = newRootCmd()
 	uFunc := cli.NewMainUsage().UsageFunc()
@@ -816,13 +817,29 @@ func shouldSkipGlobalInit(cmd *cobra.Command) bool {
 	return isSkipCommand(skipGlobalInitCommands, cmd.CommandPath())
 }
 
+var globalRootCmd *cobra.Command
+var globalRootCmdLock = &sync.Mutex{}
+
+// NewRootCmd create a new root command instance if it was not created before
+func NewRootCmd() (*cobra.Command, error) {
+	globalRootCmdLock.Lock()
+	defer globalRootCmdLock.Unlock()
+
+	if globalRootCmd == nil {
+		var err error
+		globalRootCmd, err = createRootCmd()
+		return globalRootCmd, err
+	}
+	return globalRootCmd, nil
+}
+
 // Execute executes the CLI.
 func Execute() error {
-	root, err := NewRootCmd()
+	rootCmd, err := NewRootCmd()
 	if err != nil {
 		return err
 	}
-	executionErr := root.Execute()
+	executionErr := rootCmd.Execute()
 	exitCode := 0
 	if executionErr != nil {
 		exitCode = 1
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -926,3 +926,10 @@ func printShortDescOfCmdInActiveHelp(cmd *cobra.Command, args []string) {
 		}
 	}
 }
+
+// NewRootCmdForTest creates a new instance of the root command for unit test purpose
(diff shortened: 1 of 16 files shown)


pkg/command/root.go
```diff
@@ -816,13 +817,29 @@ func shouldSkipGlobalInit(cmd *cobra.Command) bool {
...
+func NewRootCmd() (*cobra.Command, error) {
+	globalRootCmdLock.Lock()
+	defer globalRootCmdLock.Unlock()
+
+	if globalRootCmd == nil {
+		var err error
+		globalRootCmd, err = createRootCmd()
+		return globalRootCmd, err
+	}
+	return globalRootCmd, nil
```
marckhouzam, line 833: Nit: this piece of logic may be cleaner like this:

```
	globalRootCmdLock.Lock()
	defer globalRootCmdLock.Unlock()

	var err error
	if globalRootCmd == nil {
		globalRootCmd, err = createRootCmd()
	}
	return globalRootCmd, err
```
I find it makes it very clear that all the if condition does is set the singleton variable

pkg/command/root.go
```diff
@@ -816,13 +817,29 @@ func shouldSkipGlobalInit(cmd *cobra.Command) bool {
 	return isSkipCommand(skipGlobalInitCommands, cmd.CommandPath())
 }
 
+var globalRootCmd *cobra.Command
+var globalRootCmdLock = &sync.Mutex{}
+
+// NewRootCmd create a new root command instance if it was not created before
+func NewRootCmd() (*cobra.Command, error) {
+	globalRootCmdLock.Lock()
```
marckhouzam, line 825: Since the unit tests will aways first set `globalRootCmd = nil`, I don't think the locking is necessary here anymore.  But, to be really safe, we may want to add it in `NewRootCmdForTests()`

pkg/command/root.go
```diff
@@ -909,3 +926,10 @@ func printShortDescOfCmdInActiveHelp(cmd *cobra.Command, args []string) {
 		}
 	}
 }
+
+// NewRootCmdForTest creates a new instance of the root command for unit test purpose
+// Note: This must not be used as part of the production code and only used for unit tests
+func NewRootCmdForTest() (*cobra.Command, error) {
```
marckhouzam, line 932: I think this is where the locking should be done to allow running tests in parallel.
Without it, it is possible that one test does the `globalRootCmd = nil` and then gets scheduled out, letting another test call `NewRootCmd()`, which will change `globalRootCmd` to not nil.

If we add the lock here, please add a comment saying this is to allow to run unit tests in parallel.

marckhouzam (commented):
Nice solution!

pkg/command/root.go
anujc25: Decided to use `createRootCmd` directly that will by pass the usage of `globalRootCmd` entirely and always create a new instance for testing.

pkg/command/root.go
anujc25: Had a discussion and we decided to keep the looking to protect us against the goroutine sceneario mentioned by Vui.

pkg/command/root.go
anujc25: Thanks @marckhouzam for the suggestion. Updated.

marckhouzam (approved):
LGTM
Thanks for doing this!

prkalle (approved):
LGTM. Thanks for the changes!

PR #844, opened 2025-01-14
Update Code of Conduct

marckhouzam:

The repository https://github.com/vmware/.github became private today which broke our CI because we have a link to that repo.  To unblock our CI, the repo was made public again, but only temporarily.
To prepare for the repo to be private again, this PR:
- update the local CODE_OF_CONDUCT.md file to match the latest one from that repo
- updates the link in this repo to use the local CODE_OF_CONDUCT.md

Fixes # N/A

Let CI run

### Release note

```release-note
Update CODE_OF_CONDUCT.md
```

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -2,12 +2,12 @@
 
 ## Our Pledge
 
-We as members, contributors, and leaders pledge to make participation in the
-tanzu-cli project and our community a harassment-free experience for everyone,
-regardless of age, body size, visible or invisible disability, ethnicity, sex
-characteristics, gender identity and expression, level of experience,
-education, socio-economic status, nationality, personal appearance, race,
-religion, or sexual identity and orientation.
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
 
 We pledge to act and interact in ways that contribute to an open, welcoming,
 diverse, inclusive, and healthy community.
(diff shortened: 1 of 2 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -2,12 +2,12 @@
...
 ## Our Pledge
 
-We as members, contributors, and leaders pledge to make participation in the
-tanzu-cli project and our community a harassment-free experience for everyone,
-regardless of age, body size, visible or invisible disability, ethnicity, sex
-characteristics, gender identity and expression, level of experience,
-education, socio-economic status, nationality, personal appearance, race,
-religion, or sexual identity and orientation.
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
```
prkalle, line 6: question: I see we removed `tanzu-cli` project reference in this sentence. Is it intentional?

marckhouzam: Good eye.  I hadn't noticed the difference.

So, the link in CONTRIBUTING.md was referring to the code of conduct from this repo: https://github.com/vmware/.github/blob/main/CODE_OF_CONDUCT.md

And we also have a local CODE_OF_CONDUCT.md file, which I guess was adapted a bit for the CLI.

What I did was replace our local one with the exact one from the global vmware repo https://github.com/vmware/.github/blob/main/CODE_OF_CONDUCT.md

We could fixup our version for Broadcom (we never did) instead of replacing it completely, but I wonder if it is worth it?  Taking the global one will allow use to blindly update it if there is another change in the future.

But I'm fine with keeping ours if you prefer.
What do you think?

prkalle: >Taking the global one will allow use to blindly update it if there is another change in the future.

Sounds reasonable to me. Thanks for the clarification.

prkalle (approved):
LGTM. Thanks for the changes!

PR #848, opened 2025-01-28
Update telemetry to support command-remapping

marckhouzam:

This PR adapts the telemetry logic to respect command-remapping of plugin commands.

The important changes are as follows.
 
1- The telemetry plugin command-tree (`~/.cache/tanzu/plugins_command_tree/command_tree.yaml`) is no longer built when a plugin is installed, but when the plugin is first invoked.  This is because the new logic needs to find the cobra command that has been created for a remapped command, and this cobra command does not exist yet when the plugin is being installed.

2- The telemetry plugin command-tree was previously only showing the sub-commands of a plugin.  This is no longer sufficient because a plugin can remap a command to be outside its own command-tree path.  For example the `resource` plugin provides a `tanzu deploy` command which is clearly not under the `tanzu resource` command-path.  To fix this, this PR build the telemetry command-tree in such a way that it includes the entire, remapped, hierarchy of the commands provided by a plugin.  For example, for the `resource` plugin, we would see both the `deploy` and the `resource` commands at the first level of that plugin's command-tree.

3- The telemetry plugin command-tree did not previously include targets (e.g., `operations`).  This was fine because the command-tree represented sub-commands of the plugin, and when telemetry was gathered when a command was typed, it could ignore the target.  This is no longer working because when collecting telemetry, we need to start from the exact command-path that the user typed because we need to accept remapped commands; and this user input sometimes has the target present.

4- When determining aliases for a command, we now need to account for remapped command and for such cases, map the user input to the how the command can be invoked directly in the plugin.  For example, when looking for aliases for `tanzu deploy`, we need to call `<resource_plugin_binary> deploy` and this mapping is not always this straight forward.  This PR therefore looks for the cobra command that is invoked by what the user typed, and uses a new annotation to know how to map it back to a plugin invocation.  Note that it would have been possible to invoke the tanzu CLI directly to print the help of a plugin command and let it do the remapping automatically; the problem however, is that the CLI is much slower than calling a plugin directly (due to its overhead), so building the command-tree becomes very slow.  This PR therefore continues to call the plugin binary directly to invoke the help of the plugin's different command to obtain any aliases.

5- Backwards-compatibility: because the telemetry plugin command-tree is different than the one built by older CLIs, it would affect telemetry of those older CLIs, when running them after having run the new CLI with the new command-tree format.  To avoid this, this PR changes the location of the command-tree from `command_tree.yaml` to `command_tree_v2.yaml`.  With this change, both the new CLI and the older ones can be used alternatively and they will each have their own command-tree with its own format.

Fixes # N/A

### Release note

```release-note
Update the telemetry logic to support command-remapping
```


pkg/plugincmdtree/cache.go
```diff
@@ -13,13 +17,12 @@ import "github.com/vmware-tanzu/tanzu-cli/pkg/cli"
...
-	GetTree(plugin *cli.PluginInfo) (*CommandNode, error)
-	// ConstructAndAddTree constructs and adds the plugin command tree to the cache
-	// If the plugin command tree already exists, it returns success immediately
-	ConstructAndAddTree(plugin *cli.PluginInfo) error
-	// DeleteTree deletes the plugin command tree from the cache
-	DeleteTree(plugin *cli.PluginInfo) error
+	// and then returns the plugin command tree, otherwise it returns an error
+	GetTree(rootCmd *cobra.Command, plugin *cli.PluginInfo) (*CommandNode, error)
+	// DeletePluginTree deletes the plugin command tree from the cache
+	DeletePluginTree(plugin *cli.PluginInfo) error
```
prkalle, line 23: Nit: Function name `DeletePluginTree` is not opposite to `GetTree`. `GetTree` get the plugin command tree( not all plugin command trees in cache) where as `DeleteTree` deletes all the plugin command trees from cache.

pkg/plugincmdtree/plugin_cache_reset.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/vmware-tanzu/tanzu-cli/pkg/lastversion"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
+)
+
+// This global initializer checks if the last executed CLI version is < 1.5.3.
+// If so it removes the plugin command-tree cache file used by telemetry.
+// This allows a fresh cache to be gradually built using the latest command tree format.
+// Note that this initializer does not build the cache, it only removes the existing cache file.
+
+func init() {
```
vuil, line 19: Also have you checked after new format is in place what happens when we
1. use old cli to install a new plugin
2. use old cli to install the same plugin
?

vuil (commented):
This is a really nice set of changes, thanks!
Minor comment nit.
Also a few typos in the PR description

near "how the command can be invoked"
because very slow -> becomes very slow

finally some questions about behavior of an old CLI using the new command tree format.

prkalle (commented):
Thanks for changes. Nicely done! 
Looks good at the first pass. I will do some testing and update if I find any issues/comments

pkg/plugincmdtree/cache.go
marckhouzam: Good point.  Ok for me to rename `GetTree` to `GetPluginTree`?

pkg/plugincmdtree/plugin_cache_reset.go
marckhouzam: Excellent points.  I haven't tested scenarios of the old CLI.
I wonder if the co-existance tests cover some of this?
But I will try some manual scenarios.

marckhouzam: We have decided to use a different command-tree cache by renaming the file to `command_tree_v2.yaml`.
So all these concerns no longer apply as older CLIs will continue to use the original command-tree format and their telemetry won't be affected.

Thanks @vuil for bringing up this point!

vuil (approved):
thanks for the prompt updates, lgtm!

PR #851, opened 2025-01-30
Allow replacing an existing RPM package

marckhouzam:

It has happened that an RPM package was not signed correctly but we still built and pushed the repo with this package.

This commit allows to replace a published RPM package within the repo.

Fixes # N/A

### Release note

```release-note
Allow replacing a wrongly signed RPM package
```

### DIFF
--- a/hack/rpm/build_package_repo.sh
+++ b/hack/rpm/build_package_repo.sh
@@ -22,6 +22,15 @@ if [ -z "$DNF" ]; then
    exit 1
 fi
 
+# VERSION should be set when calling this script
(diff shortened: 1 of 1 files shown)


marckhouzam:
This PR was meant to rebuild an RPM package, sign it and rebuild the repo replacing the rebuilt package.
The idea was to fix a situation where a package had not been signed.

A better approach is not to rebuild the RPM package, but simply to download the existing package, and to sign it and then rebuild the repo.  Therefore we won't need this PR.

PR #856, opened 2025-02-25
Improve testing for help command for plugins

marckhouzam:

There was a [change in Cobra 1.9.0](https://github.com/spf13/cobra/pull/2173) which caused our unit tests to fail and showed that calling `rootCmd.SetArgs(spec.args)` is the proper approach when setting arguments to a cobra command in tests, even if `os.Args` is also used.

Also, because of the way we handle the help command for plugins in `getHelpArguments()`, we must also continue to set `os.Args`; however, the way the tests were written before this PR would not cause a failure if the call to `os.Args` was removed.

This commit updates the tests to better handle both cases.

Note that because of impacts to other programs, the [particular change in Cobra was reverted in release v1.9.1](https://github.com/spf13/cobra/pull/2235).

Fixes # N/A

Only impacts tests, so CI.

### Release note

```release-note
Improve testing for plugins' help command
```

### DIFF
--- a/pkg/command/root_test.go
+++ b/pkg/command/root_test.go
@@ -37,15 +37,15 @@ const (
 info() {
    cat << EOF
 {
-  "name": "%s",
-  "description": "%s functionality",
+  "name": "%[1]s",
+  "description": "%[1]s functionality",
   "version": "v0.1.0",
   "buildSHA": "01234567",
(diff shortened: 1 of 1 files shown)


prkalle (approved):
LGTM. Thanks for the changes!

PR #860, opened 2025-02-28
Substitute wrong dash character

marckhouzam:

My IDE showed me that there are a couple of places where we used the wrong dash character instead of the ascii one.
This PR simply replaces them.

For example, when copying the docs where this wrong character was used:
```
$ tanzu plugin install all -–group vmware-tanzucli/essentials
[x] : unknown shorthand flag: 'â' in -–group
```

Fixes # N/A

Let CI run

### Release note

```release-note
Convert some characters to their corresponding ascii version.
```

### DIFF
--- a/docs/quickstart/install.md
+++ b/docs/quickstart/install.md
@@ -474,13 +474,13 @@ To manually install the essentials plugin group.
 To install latest version
 
 ```shell
-tanzu plugin install all -–group vmware-tanzucli/essentials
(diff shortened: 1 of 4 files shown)


prkalle (approved):
LGTM, Thanks for fixing this.

(This is very hard to differentiate visually and I encountered simialr issue in the past while testing)

anujc25 (approved):
LGTM.

